PR #7, opened 2023-03-10
add function to enable gptel in a saved chat

thanhvg:
Hi, 

Thank you for your comprehensive review of GPT clients for Emacs on Reddit. I like yours the most :).

This merge request allows us to save the current session as an MD file. Later, when we open the file again, `gptel-enable-current-buffer` will resume the session.

This solution may not be the best. Perhaps a derived mode based on markdown-mode and GPTel for a dedicated chat file (for example .gpt extension) is a better solution. However, this solution scratches the itch for me.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -262,7 +262,10 @@ Ask for API-KEY if `gptel-api-key' is unset."
                                (read-passwd "OpenAI API key: ")))))
   (unless api-key
     (user-error "No API key available"))
-  (with-current-buffer (get-buffer-create name)
+  (gptel--ensure-buffer (get-buffer-create name)))
+
+(defun gptel--ensure-buffer (buffer)
+  (with-current-buffer buffer
     (cond ;Set major mode
      ((eq major-mode gptel-default-mode))
      ((eq gptel-default-mode 'text-mode)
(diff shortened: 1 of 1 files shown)


minad:
Yes, I would also like to have this functionality.

karthink:
@thanhvg Thanks for the PR!

The reason I haven't implemented persistence yet is because I don't know a good way to do it. Right now gptel uses text properties to determine who said what, _i.e._ to distinguish between what you typed and what ChatGPT generated.  These are not saved to disk along with the buffer contents, so upon loading the buffer this information is lost, making the conversation unrecoverable.

Other ChatGPT packages are dealing with this by adding structure to the conversation, where what you type and the generated response are formatted differently: e.g.
- queries are Org headlines and responses are contents, or
- responses are delimited by org source blocks (`#+begin_src` etc), or
- by adding text markers like horizontal rules (`----`) before and after responses, etc.

My goal is to have freeform interaction in _any_ buffer without markup constraints, this allows the user to format things any way they want.  I'm almost there now, you can select a region of text in _any_ buffer and call `gptel-send`.  If the region includes past responses from ChatGPT, those are identified and sent as well.

So I'll need to think about persistence a little more.  The only other idea I have is to store the metadata to disk separately and apply it when opening the file (or when calling `gptel-enable-current-buffer`), but this is a fragile solution that breaks easily depending on how the file is modified.

If you have any ideas, please let me know.

minad:
It doesn't sound bad to me if you add metadata below the headlines. Org `:PROPERTIES:` drawers seem to be a good fit for this.

karthink:
Yes, that is a good solution for org-mode buffers. For markdown, text or prog-mode buffers it's looking like the only possible solutiosn are

- metadata outside the document, or
- explicit delimiters: code comments in prog-mode, headings or rules in markdown etc.

minad:
I would just introduce similar metadata blocks within the document (source blocks, comments, ...). You could also put the metadata at the end of the document, but this will make the diffs worse. The same applies to metadata outside the document - I would avoid that.

TxGVNN:
I found a more complicated package. https://github.com/CarlQLange/chatgpt-arcana.el
But it saves conversations in `~/.emacs.d/.local/cache/chatgpt-arcana/sessions`, it is so useful. Just FYI

Issue #16, opened 2023-03-12
No response/error indicator when executing `gptel`


randomwangran:
Execute `M-x gptel`, it saids Waiting, but no response can received. 
`(setq gptel-api-key "sk-xxxxx")` has been executed. Here's a screenshot:

https://user-images.githubusercontent.com/8197831/224576530-a1888203-46a3-4886-970b-f5e7a1511c59.mp4

karthink:
Hi @randomwangran,

I just added a debug setting to diagnose issues like this. Could you

(i) Pull the latest from this repo and install/load it,
(ii) Run `(setq gptel--debug t)` before running gptel,
(iii) Run gptel and wait for a buffer named `*gptel-error*` response to pop up,
(iv) and paste the contents of the buffer here?

If no such buffer pops up, that's helpful too and you can let me know.

jmercouris:
I ran the code, waited several minutes, and no error buffer popped up.

karthink:
@jmercouris Thank you. Is the value of `gptel-use-curl` non-nil for you? (It might look like `/usr/bin/curl`)

jmercouris:
```
gptel-use-curl is a variable defined in ‘gptel.el’.

Its value is t

Whether gptel should prefer Curl when available.

  You can customize this variable.
```

jmercouris:
Please note:

```
[jmercouris@system:~/.emacs.d/packages]$ which curl
/run/current-system/sw/bin/curl
```

I am running NixOS.

karthink:
Could you set `gptel-use-curl` to `nil` and check if the problem persists?

ydche3:
The same problem. And set `gptel-use-curl` in `gptel.el` to `nil` couldn't solve the problem.

karthink:
@ydche3 Could you try again? I removed the aio dependency and made the code simpler to debug in 040baad9.

ydche3:
> @ydche3 Could you try again (with `gptel--debug` set to `t`)? I removed the aio dependency and made the code simpler to debug in [040baad](https://github.com/karthink/gptel/commit/040baad9101d8bb2bedbedc15b9b8f00cc7d9c05).

Without any difference. But I noticed that whenever run `C-c RET` to send my query, the request wasn't sent behind my proxy.
Here is my emacs proxy setting:
```
(setq url-proxy-services '(("http" . "127.0.0.1:7890")
                           ("https" . "127.0.0.1:7890")))
```
Should I change my proxy setting?

[...]

Issue #2, opened 2023-03-07
Request failure when prompt has special characters, like accented characters or cedilla


jjnilton:
Here's what I get in the `**Messages**` buffer:
```
Querying ChatGPT...
error in process sentinel: url-http-create-request: Multibyte text in HTTP request: POST /v1/chat/completions HTTP/1.1
MIME-Version: 1.0
Connection: keep-alive
Extension: Security/Digest Security/SSL
Host: api.openai.com
Accept-encoding: gzip
Accept: */*
User-Agent: URL/Emacs Emacs/27.1 (X11; x86_64-pc-linux-gnu)
Content-Type: application/json
Authorization: Bearer [token]
Content-length: 104

{"model":"gpt-3.5-turbo","messages":[{"role":"user","content":"qual é o melhor editor, vim ou emacs?"}]}
error in process sentinel: Multibyte text in HTTP request: POST /v1/chat/completions HTTP/1.1
```

Thanks for sharing the package.

karthink:
Thanks for the report.  It looks like this is related to Emacs Bug 23750.  Unfortunately using the [solution emacs-devel suggests](https://lists.gnu.org/archive/html/emacs-devel/2020-06/msg00509.html) isn't working for me.

So here's what I'm thinking instead.  Do you have `curl` available on your system?  I'll make it so requests are handled by curl if it's available, falling back to Emacs' `url-retrieve`.

karthink:
If you have access to curl, this problem should be fixed in 3c10147a, please test?

If you don't have curl, it might take a while before I have time to dig into the multibyte encoding issues.

There are also a few new user options.

jjnilton:
Thanks, it worked! I had to load the `gptel-curl` manually, though. `package-install-file` just let me install the `gptel` but threw an error for the `gptel-curl`:

> package-buffer-info: Package lacks a "Version" or "Package-Version" header

karthink:
That's odd, it installed fine using `package-install-file` for me.

Anyway, I've added the Version header to gptel-curl. When you have time could you try uninstalling and reinstalling the package? No hurry though.

jjnilton:
I did the following:

1. Uninstalled `gptel` with `package-delete`
2. Installed `gptel` with `package-install-file`
3. Entered the API key and tried a prompt with special characters, it did not fall back to `gptel-curl`.
4. Tried to install `gptel-curl` with `package-install-file` and got the error I mentioned in my last comment.

[...]

PR #18, opened 2023-03-18
Add gptel-api-key-from-auth-source (Fix #13)

### LINKED ISSUE #13: Add auth-source api function
Would it make sense to provide this function or even set `gptel-apikey` to this function by default?

~~~ elisp
(defun gptel-apikey-from-auth-source ()
  (funcall (plist-get (car (auth-source-search :host "openai.com" :user "apikey")) :secret)))))
~~~

minad:
In order to use this your authinfo.gpg file should contain an entry like this:

~~~
machine openai.com login apikey password <key>
~~~

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -46,15 +46,9 @@ PROMPTS is the data to send, TOKEN is a unique identifier."
          (data (encode-coding-string
                 (json-encode (gptel--request-data prompts))
                 'utf-8))
-         (api-key
-          (cond
-           ((stringp gptel-api-key) gptel-api-key)
-           ((functionp gptel-api-key) (funcall gptel-api-key))
-           (t (setq gptel-api-key (read-passwd "OpenAI API key: ")))))
          (headers
           `(("Content-Type" . "application/json")
-            ("Authorization" . ,(concat "Bearer " api-key)))))
-
+            ("Authorization" . ,(concat "Bearer " (gptel--api-key))))))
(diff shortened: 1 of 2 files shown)


gptel-curl.el
```diff
@@ -46,15 +46,9 @@ PROMPTS is the data to send, TOKEN is a unique identifier."
...
-          (cond
-           ((stringp gptel-api-key) gptel-api-key)
-           ((functionp gptel-api-key) (funcall gptel-api-key))
-           (t (setq gptel-api-key (read-passwd "OpenAI API key: ")))))
          (headers
           `(("Content-Type" . "application/json")
-            ("Authorization" . ,(concat "Bearer " api-key)))))
-
+            ("Authorization" . ,(concat "Bearer " (gptel--api-key))))))
     (push (format "-X%s" "POST") args)
```
minad, line 52: I am a bit unhappy with the code duplication of the two backends (url and curl). I think it would be okay if you restrict gptel to a single backend and require curl if that works much better for you.

karthink:
Thanks, I'll get to this soon!

gptel-curl.el
karthink: You're right, it's a bit of a mess right now. Let's just require curl for now -- but after we ensure that it works for everyone. Right now curl doesn't work for some users (#3), `url-retrieve` doesn't work for some (#2), and both backends don't work for others (#16).

PR #24, opened 2023-03-24
Add autoload for gptel-send

akirak:
Hello,

I use `gptel-send` command more frequently than `gptel`. It would be nicer if this command were autoloaded.


karthink:
Thank you!

PR #28, opened 2023-03-30
Add Doom Emacs installation instructions

AlessandroW:


karthink:
Thanks! Could you change "Doom" to "Doom Emacs" and move this below the manual section (and above Spacemacs)?

M1ndo:
In *packages.el*
```elisp
(package! gptel :recipe (:host github :repo "karthink/gptel"))
```
In *bash*
```sh
doom sync
```
In *Emacs*
```elisp
M-x doom/reload-packages
M-x gptel
```

karthink:
> In packages.el since some will experience "package not found"

I don't use Doom so I'm not sure about this -- but since gptel is on MELPA and Doom uses Straight, why would this happen?

AlessandroW:
It won't afaik and I think it's better to pull the package from melpa than from github.

AlessandroW:
> Thanks! Could you change "Doom" to "Doom Emacs" and move this below the manual section (and above Spacemacs)?

done

karthink:
Cheers @AlessandroW

M1ndo:
> > In packages.el since some will experience "package not found"
> 
> I don't use Doom so I'm not sure about this -- but since gptel is on MELPA and Doom uses Straight, why would this happen?

I honestly have no idea, i wasn't able to pull it out of melpa (other packages worked).  I just left that comment there, in-case somebody else stumbled upon the same issue.

PR #29, opened 2023-03-31
Resolved #4: Handles multi-byte characters with url-retrieve (i.e., without the external curl command)

### LINKED ISSUE #4: Multi-byte char handling with url-retrieve is broken
The [proposed solution](https://lists.gnu.org/archive/html/emacs-devel/2020-06/msg00513.html) for Emacs bug 23750 is to call `encoding-coding-string` on the JSON request and send `utf-8` text, but this does not seem to work.  This bug and workaround is also explained [here](https://github.com/magit/ghub/issues/12).

``` emacs-lisp
(let (((url-request-method "POST")
       (url-request-extra-headers
        '(("Content-Type" . "application/json")))
       (url-request-data
        (encode-coding-string
         (json-encode
          '(:messages [(:role "user"
                        :content "qual é o melhor editor, vim ou emacs?")]))
         'utf-8))))
  (url-retrieve-synchronously "https://api.openai.com/v1/chat/completions"))
```

`url-http-create-request` still complains about the message length and message size (in bytes) not lining up.

algal:
These two commits fix handling of multi-byte characters in the prompt send to OpenAI and in the response received from it.

As a result, you can now use non-Latin alphabets like Greek, or emojis.

To make requests work, it was necessary to encode the loaded API key as utf-8, for reasons that remain obscure, but may have to do with some subtlety concerning how the `concat` function combines multibyte and non-multibyte strings. To make replies work, it was necessary to decode the response body as utf-8 before, not after, parsing it as json.

I have tested this lightly on Emacs 28.1 running on Linux.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -417,16 +417,15 @@ INFO is a plist with the following keys:
       (if-let* ((status (buffer-substring (line-beginning-position) (line-end-position)))
                 (json-object-type 'plist)
                 (response (progn (forward-paragraph)
-                                 (condition-case nil
-                                         (json-read)
-                                       (json-readtable-error 'json-read-error)))))
+                                 (let ((json-str (decode-coding-string
+                                                  (buffer-substring-no-properties (point) (point-max))
+                                                  'utf-8)))
+                                   (condition-case nil
+                                       (json-read-from-string json-str)
+                                     (json-readtable-error 'json-read-error))))))
           (cond
            ((string-match-p "200 OK" status)
-            (list :content (string-trim
-                            (decode-coding-string
-                             (map-nested-elt
-                              response '(:choices 0 :message :content))
-                             'utf-8))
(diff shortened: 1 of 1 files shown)


algal:
This is intended to resolve #4

karthink:
Appears to work for me too. This is great! I'm also on Emacs 28.2 on Linux, I'm going to make the assumption here that explicitly encoding all parts of the request as utf-8 is the conservative approach and thus unlikely to cause issues in other versions or platforms.

> To make requests work, it was necessary to encode the loaded API key as utf-8, for reasons that remain obscure, but may have to do with some subtlety concerning...

Ah, I can see why this could be a problem.

Pushed.

narendraj9:
Wouldn't it be better to always encode all headers (including `Authorization`) at a lower level to `utf-8` instead of just modifying the auth source function? @algal Or am I missing something about the issue here?

PR #38, opened 2023-04-08
Give pulse-momentary-highlight-region a fallback END argument

twashing:
If there is an error in the response, instead of giving the response error, it's giving an error of `pulse-momentary-highlight-region`.


karthink:
Thanks, I've had trouble generating erroneous responses for testing.

When there's an error or no response, it shouldn't pulse anything at all. Could you change it instead to

```lisp
(when tracking-marker
   (pulse-momentary-highlight-region ...))
```

twashing:
> EDIT: I rewrote the process sentinel to better handle errors, so this fix should not be needed any more. Thanks for figuring out the problem! 

Ah ok 👍🏿 I've pushed a commit of your suggestion, in case you need it. 
Thanks!

PR #46, opened 2023-04-23
README: Update instructions for setting key

tmr08c:
* README.org Update the instructions for getting =gptel-api-key= to include using the .authinfo file after support was added in 6f951ed.


karthink:
Looks good, thanks @tmr08c!

PR #49, opened 2023-04-28
Make OpenAI API endpoint configurable

christianromney:
Also encode stream :json-false when stream is nil
since the API requires a boolean value (https://platform.openai.com/docs/api-reference/chat).

Signed-off-by: Christian Romney <christian.romney@nubank.com.br>

### DIFF
--- a/README.org
+++ b/README.org
@@ -114,6 +114,11 @@ If you want custom behavior, such as
 
 GPTel provides a general =gptel-request= function that accepts a custom prompt and a callback to act on the response. You can use this to build custom workflows not supported by =gptel-send=.  See the documentation of =gptel-request=, and the [[https://github.com/karthink/gptel/wiki][wiki]] for examples.
 
+** Additional Configuration
+
+- You can override the OpenAI API endpoint by customizing ~gptel-openai-endpoint~.
+
+
 ** Why another ChatGPT client?
 
 Existing Emacs clients don't /reliably/ let me use it the simple way I can in the browser.  They will get better, but I wanted something for now.
--- a/gptel.el
+++ b/gptel.el
@@ -487,7 +487,7 @@ there."
   (let ((prompts-plist
          `(:model ,gptel-model
            :messages [,@prompts]
-           :stream ,(and gptel-stream gptel-use-curl))))
+           :stream ,(or (and gptel-stream gptel-use-curl) :json-false))))
     (when gptel-temperature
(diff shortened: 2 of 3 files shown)


christianromney:
Hi! I really like the user experience of your package. I prefer it to the chatgpt package from emacs-openai. I had a problem where I needed to send requests through a (non-system-wide) proxy which led me to add the base uri as a custom var. I also force the json-encoding of stream to a boolean value as indicated by the API documentation. Thanks for sharing your package!

gptel.el
```diff
@@ -487,7 +487,7 @@ there."
   (let ((prompts-plist
          `(:model ,gptel-model
            :messages [,@prompts]
-           :stream ,(and gptel-stream gptel-use-curl))))
+           :stream ,(or (and gptel-stream gptel-use-curl) :json-false))))
```
karthink, line 490: Thanks, I was fuzzy about the difference between `nil` and `:json-false`.

Could you make just this change the first commit?

README.org
```diff
@@ -114,6 +114,11 @@ If you want custom behavior, such as
 
 GPTel provides a general =gptel-request= function that accepts a custom prompt and a callback to act on the response. You can use this to build custom workflows not supported by =gptel-send=.  See the documentation of =gptel-request=, and the [[https://github.com/karthink/gptel/wiki][wiki]] for examples.
 
+** Additional Configuration
+
+- You can override the OpenAI API endpoint by customizing ~gptel-openai-endpoint~.
```
karthink, line 119: Replace with `=gptel-openai-endpoint=`, as I use `=` delimiters for verbatim commands and `~` delimiters for notating keybindings. (There is no distinction between them in the Github renderer but these are different things when viewing this file in Emacs.)

gptel.el
```diff
@@ -522,6 +522,9 @@ BUFFER is the interaction buffer for ChatGPT."
     ('org-mode (gptel--convert-markdown->org content))
     (_ content)))
 
+(defcustom gptel-openai-endpoint "https://api.openai.com/v1"
```
karthink, line 525: This definition needs a `:type 'string` annotation. While it's not required by Emacs, please also add a `:group 'gptel` annotation for consistency with the other definitions in this package.

karthink (changes requested):
Thank you! Please see the review comments.

karthink:
This has been superceded by #67.

PR #69, opened 2023-06-06
feature/curl proxy support

PalaceChan:
This is an amazing package. 

In many organizations the openai api can only be accessed via proxy. This is something easily supported by curl. This PR proposes an (optionally empty) path to a proxy to enable gptel in these cases.

Implemented in two commits purely for expository reasons (easily squashed into one)

The first commit minimally shows how simple the change is, the crux of that diff being:

```
@@ -54,6 +61,12 @@ PROMPTS is the data to send, TOKEN is a unique identifier."
     ;; (push (format "--keepalive-time %s" 240) args)
     (push (format "-m%s" 60) args)
     (push "-D-" args)
+    (when (not (string-empty-p gptel-proxy-path))
+      (push "--proxy" args)
+      (push (format "%s" gptel-proxy-path) args)
+      (push "--proxy-negotiate" args)
+      (push "--proxy-user" args)
+      (push ":" args))
```

The second commit is a proposed cleanup of the overall function (from a bunch of single pushes) to leveraging `list` and `append` and also avoiding a reverse at the end. Both commits worked with and without a proxy server.

(GPT-4, together with `gptel`, actually helped with parts of the second commit)

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -36,6 +36,13 @@
 (defvar gptel-curl--process-alist nil
   "Alist of active GPTel curl requests.")
 
+(defcustom gptel-proxy-path ""
+  "Path to a proxy to use for gptel interactions.
+Passed to curl via --proxy arg, for example 'proxy.yourorg.com:80'
+Leave it empty if you don't use a proxy."
+  :group 'gptel
+  :type 'string)
+
 (defun gptel-curl--get-args (prompts token)
   "Produce list of arguments for calling Curl.
 
@@ -54,6 +61,12 @@ PROMPTS is the data to send, TOKEN is a unique identifier."
     ;; (push (format "--keepalive-time %s" 240) args)
     (push (format "-m%s" 60) args)
     (push "-D-" args)
+    (when (not (string-empty-p gptel-proxy-path))
+      (push "--proxy" args)
+      (push (format "%s" gptel-proxy-path) args)
+      (push "--proxy-negotiate" args)
+      (push "--proxy-user" args)
+      (push ":" args))
     (pcase-dolist (`(,key . ,val) headers)
       (push (format "-H%s: %s" key val) args))
     (push (format "-d%s" data) args)
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -56,8 +56,8 @@ PROMPTS is the data to send, TOKEN is a unique identifier."
            "-D-"
            (format "-d%s" data)
            url)
-     (when (not (string-empty-p gptel-proxy-path))
-       (list "--proxy" gptel-proxy-path
+     (when (not (string-empty-p gptel-proxy))
+       (list "--proxy" gptel-proxy
              "--proxy-negotiate"
              "--proxy-user" ":"))
      (cl-loop for (key . val) in headers
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -29,6 +29,7 @@
 (require 'gptel)
 
 (eval-when-compile
+  (require 'cl-lib)
   (require 'subr-x))
 (require 'map)
 (require 'json)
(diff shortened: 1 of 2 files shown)


gptel.el
```diff
@@ -74,6 +74,13 @@
   :group 'gptel
   :type 'string)
 
+(defcustom gptel-proxy-path ""
```
karthink, line 77: Please rename to just `gptel-proxy`, for consistency with `gptel-host` (which is not `gptel-host-path`.)

gptel-curl.el
```diff
@@ -29,48 +29,39 @@
...
-    (push (format "-d%s" data) args)
-    (nreverse (cons url args))))
+    (append
+     (list "--location" "--silent" "--compressed" "--disable"
+           (format "-X%s" "POST")
+           (format "-w(%s . %%{size_header})" token)
+           (format "-m%s" 60)
+           "-D-"
+           (format "-d%s" data)
+           url)
```
karthink, line 58: To aid with debugging, it will help to move `url` to the end of the argument list, like in the previous code.

So the curl invocation will look like

```sh
curl --location --silent ... --proxy ... -H%s: %s -H%s: %s url
```

karthink: You can do this by moving the cl-loop up in the list of lists being appended.

PalaceChan: makes sense, you mean like below?

```
-           (format "-d%s" data)
-           url)
-     (when (not (string-empty-p gptel-proxy-path))
-       (list "--proxy" gptel-proxy-path
+           (format "-d%s" data))
+     (when (not (string-empty-p gptel-proxy))
+       (list "--proxy" gptel-proxy
              "--proxy-negotiate"
              "--proxy-user" ":"))
      (cl-loop for (key . val) in headers
-              collect (format "-H%s: %s" key val)))))
+              collect (format "-H%s: %s" key val))
+     (list url))))
```

karthink:
Looks good, thank you.

> The second commit is a proposed cleanup of the overall function (from a bunch of single pushes) to leveraging list and append and also avoiding a reverse at the end.

`push` + `nreverse` is how `cl-loop` generates lists, so this is not the case, BTW. The old and new versions expand to almost the same code:

Old, `pcase-dolist` version expansion:
```emacs-lisp
(cl-block nil
  (let
      ((--dolist-tail-- headers)
       x1262)
    (while --dolist-tail--
      (setq x1262
            (car --dolist-tail--))
      (progn
        (ignore
         (consp x1262))
        (let*
            ((x1263
              (car-safe x1262))
             (x1264
              (cdr-safe x1262)))
          (let
              ((key x1263)
               (val x1264))
            (push
             (format "-H%s: %s" key val)
             args))))
      (setq --dolist-tail--
            (cdr --dolist-tail--)))))
(nreverse (cons url args))
```

New, `cl-loop` version expansion
```emacs-lisp
(cl-block nil
  (let*
      ((--cl-var-- headers)
       (val nil)
       (key nil)
       (--cl-var-- nil))
    (while
        (consp --cl-var--)
      (setq val
            (car --cl-var--)
            key
            (pop val))
      (push
       (format "-H%s: %s" key val)
       --cl-var--)
      (setq --cl-var--
            (cdr --cl-var--)))
    (nreverse --cl-var--)))
```

Also a bunch of single pushes is no different from using list + append (performance-wise). I structured it like that originally because I wasn't sure if those Curl arguments would need to be able to be turned off individually depending on the needs of the package.  As it turned out, there was no such need.

That said, the cleaned up version looks good for now, thanks.

gptel-curl.el
karthink: Yup this will work :+1:

PalaceChan:
TIL on the `cl-loop` macro expand vs the `pcase-dolist` version

karthink:
Okay, we are good to go, merging.

> TIL on the `cl-loop` macro expand vs the `pcase-dolist` version

(`push` + `nreverse` is the most efficient way to generate a list in elisp, so every accumulator will use these two)

karthink:
Cheers @PalaceChan

PR #77, opened 2023-06-14
Add new turbo 0613 models

fmguerreiro:
Based on the new model update that got announced recently:

スクリーンショット 2023-06-14 15 15 10

Not sure how welcome this is, might get a bit tricky to keep up to date.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -237,11 +237,11 @@ will get progressively longer!"
   :class 'transient-lisp-variable
   :variable 'gptel-model
   :key "m"
-  :choices '("gpt-3.5-turbo-16k" "gpt-3.5-turbo" "gpt-4" "gpt-4-32k")
+  :choices '("gpt-3.5-turbo" "gpt-3.5-turbo-16k" "gpt-4" "gpt-4-32k")
   :reader (lambda (prompt &rest _)
             (completing-read
              prompt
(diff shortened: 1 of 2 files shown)


karthink:
Thanks, this looks good.

Could you also update the documentation for the user option `gptel-model` with the new choices?

fmguerreiro:
> Could you also update the documentation for the user option gptel-model with the new choices?

done in [f43562e](https://github.com/karthink/gptel/pull/77/commits/f43562e41d3951e1a3ebca2538e71f080745724d)

PalaceChan:
in about [a week](https://platform.openai.com/docs/models/gpt-3-5) (june 27th) the 3.5 and 4 models will be the same as the 0613 variants while the march ones will be deprecated and retired in september. Is it worth adding these dangling "symlink" models? On the other hand the more interesting permanent ones are the 16k 3.5 and the 32k 4.0 variants so might it be a better idea to add those instead and link to the openai models page?

karthink:
@PalaceChan Didn't know that, thanks for pointing it out. I think it's better to add only the new permanent variants in this case. @fmguerreiro Could you make the change?

fmguerreiro:
> @fmguerreiro Could you make the change?

Done!

- Kept gpt-3.5-turbo and gpt-3.5-turbo-16k https://platform.openai.com/docs/models/gpt-3-5
- Kept gpt-4 and gpt-4-32k https://platform.openai.com/docs/models/gpt-4

karthink:
Chers @fmguerreiro, @PalaceChan

PR #78, opened 2023-06-15
Update gptel.el: Needs :require for auth-source-search

marcuskammer:
To read from .authinfo.gpg the key parameter :require for auth-source-search is needed.


marcuskammer:
System: Windows 10, Emacs 28.1

karthink:
Thanks @marcuskammer!

rennsax:
If the user only have `~/.authinfo` and do not want to encrypt it, with this selector hard-coded, `gptel-api-key-from-auth-source` cannot get the apikey. Now I have to pass a function to `:key` when customizing my gpt backend.

PR #85, opened 2023-07-04
Fix read temperature from minibuffer

tshu-w:
The returned result now is a string instead of a number, and the code is from `set-variable`.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -242,7 +242,8 @@ include."
   :variable 'gptel--num-messages-to-send
   :key "n"
   :prompt "Number of past messages to include for context (leave empty for all): "
-  :reader 'transient-read-number-N0)
+  :reader (lambda (prompt initial-input history)
+            (read-from-minibuffer prompt initial-input read-expression-map t history)))
 
 (transient-define-infix gptel--infix-max-tokens ()
   "Max tokens per response.
@@ -259,7 +260,8 @@ will get progressively longer!"
   :variable 'gptel-max-tokens
   :key "<"
   :prompt "Response length in tokens (leave empty: default, 80-200: short, 200-500: long): "
-  :reader 'transient-read-number-N+)
+  :reader (lambda (prompt initial-input history)
+            (read-from-minibuffer prompt initial-input read-expression-map t history)))
 
 (transient-define-infix gptel--infix-model ()
   "AI Model for Chat."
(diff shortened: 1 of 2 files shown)


karthink:
By setting the `read` argument to `t`, the returned result is now a _number_, not a string.  This is useful in that we can get rid of `gptel--numberize`, but then reading in`gptel-max-tokens` should also be fixed.

tshu-w:
Hi, I have updated my PR with `gptel--numberize` removement.

gptel-transient.el
```diff
@@ -268,20 +272,17 @@ will get progressively longer!"
   :variable 'gptel-model
   :key "m"
   :choices '("gpt-3.5-turbo" "gpt-3.5-turbo-16k" "gpt-4" "gpt-4-32k")
-  :reader (lambda (prompt &rest _)
-            (completing-read
-             prompt
-             '("gpt-3.5-turbo" "gpt-3.5-turbo-16k" "gpt-4" "gpt-4-32k"))))
+  :reader 'gptel--transient-read-variable)
```
tshu-w, line 275: I accident change this, will remove tomorrow

gptel.el
```diff
@@ -253,18 +253,14 @@ By default, `gptel-host' is used as HOST and \"apikey\" as USER."
...
-;; FIXME Should we utf-8 encode the api-key here? 
+;; FIXME Should we utf-8 encode the api-key here?
 (defun gptel--api-key ()
   "Get api key from `gptel-api-key'."
   (pcase gptel-api-key
     ((pred stringp) gptel-api-key)
     ((pred functionp) (funcall gptel-api-key))
     (_ (error "`gptel-api-key' is not set"))))
 
-(defsubst gptel--numberize (val)
```
karthink, line 264: Please add `gptel--numberize` back in, as I need this elsewhere in the code.  The rest of the changes look good!

karthink (changes requested):

gptel.el
tshu-w: Hello, it seems that I have already removed all the places where `gptel--numberize` was used. Is this function no longer being used in the entire project?

karthink: Yes, it is no longer being used. I will need to use it again in the future though, so it will be better to just leave it in.

tshu-w: OK, I have added this function back ;)

karthink:
Cheers @tshu-w

PR #88, opened 2023-07-12
Add another key retrieval method to README.org

tony-dwire-paradigm:
I typically use this method for storing the key, easier than filling out ~/.authinfo  but more secure than keeping it in config. Just another option, feel free to decline PR if you think it's not valuable.

### DIFF
--- a/README.org
+++ b/README.org
@@ -85,6 +85,14 @@ Optional: Set =gptel-api-key= to the key. Alternatively, you may choose a more s
   #+begin_src authinfo
 machine api.openai.com login apikey password TOKEN
   #+end_src
+- Storing in any file and manually reading it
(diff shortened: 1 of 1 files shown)


karthink:
1. I don't want to encouraging storing the API key in a plain text file.  If using `insert-file-contents` on a gpg (or otherwise) encrypted file, the code should be prepended with `(epa-file-enable)`.
2. These instructions use macros from Doom emacs, please rewrite without `use-package!` and `setq!`.
3. I would like to keep the README short, but this is a useful addition to the [wiki](https://github.com/karthink/gptel/wiki).  Could you add it there after fixing 1 and 2?

karthink:
Closing as this is not planned.

PR #89, opened 2023-07-12
add optional system message argument to gptel--create-prompt

neilfulwiler:
not sure if this was a deliberate choice but I assumed it wasn't, I was playing around with gtpel-request and wanted to use the builtin prompt functionality but with a custom system message

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -402,7 +402,9 @@ Model parameters can be let-bound around calls to this function."
             (set-marker (make-marker) position buffer))))
          (full-prompt
          (cond
-          ((null prompt) (gptel--create-prompt start-marker system))
+          ((null prompt) 
+           (let ((gptel--system-message system))
+             (gptel--create-prompt start-marker)))
           ((stringp prompt)
            `((:role "system" :content ,system)
              (:role "user"   :content ,prompt)))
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for catching that, I hadn't considered a `nil` prompt + custom system message.

I'd like to avoid modifying `gptel--system-prompt` for now, as it needs to be rewritten to support more LLMs soon. Please see the code review for an alternative.

neilfulwiler:
assuming you mean an upcoming code review, cause I don't see one now (but maybe I don't know how to github). if you haven't gotten around to it yet the other way I could do this is just get the prompt list and pop off the system message and add the custom one back, not sure how stable that data format is though

karthink:
2023-07-13-142516_930x886_scrot
Not sure why it's not showing up.

Reproduced here:

Instead of modifying `gptel--create-prompt`, please let-bind gptel--system-message around the call to `gptel--create-prompt` in `gptel-request`. Something like this:

```elisp
          ((null prompt)
           (let ((gptel--system-message
                  (or system gptel--system-message)))
             (gptel--create-prompt start-marker)))
```

neilfulwiler:
ah its pending you gotta hit submit, also that makes way more sense than what I said

gptel.el
```diff
@@ -402,7 +402,9 @@ Model parameters can be let-bound around calls to this function."
             (set-marker (make-marker) position buffer))))
          (full-prompt
          (cond
-          ((null prompt) (gptel--create-prompt start-marker system))
+          ((null prompt) 
+           (let ((gptel--system-message (or system gptel--system-message)))
```
karthink, line 406: I just realized you don't need `(or system gptel--system-message)` here, just `system` should do since it's set to `gptel--system-message` in the `cl-defun` call.

neilfulwiler: makes sense, thanks. updated

karthink:
Cheers @neilfulwiler.

PR #104, opened 2023-08-26
Add support for Azure-OpenAI API

doctorguile:
Summary of changes in gptel-curl.el:
- Added conditional logic to generate the URL based on whether Azure-Openai API is being used or not
- Added support for passing the Azure-Openai API key as a header

Summary of changes in gptel.el:
- Added customization variables for Azure-Openai API
- Added support for generating the URL based on whether Azure-Openai API is being used or not
- Added support for passing the Azure-Openai API key as a header

PR #111, opened 2023-10-05
Allow setting query protocol

janEbert:
`https` was previously hardcoded; this allows other protocols to be used (useful for hosting locally) while keeping the default.


karthink:
@janEbert Thanks. I'm closing this now as multiple LLM backend support is now underway (see #104 for an update).

PR #112, opened 2023-10-15
gptel: Use save-excursion when major-mode is org-mode

codeasone:
I was using `gptel-mode` when editing `org-roam` files and noticed point was moving to the top of the buffer on save, not returning to where I was last editing the file as expected

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -258,7 +258,7 @@ The current options are
 - \"gpt-3.5-turbo-16k\"
 - \"gpt-4\" (experimental)
 - \"gpt-4-32k\" (experimental)
- 
+
 To set the model for a chat session interactively call
 `gptel-send' with a prefix argument."
   :local t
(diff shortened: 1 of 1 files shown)


karthink:
Hmm, I can't seem to reproduce this.  I already manage the excursion and restriction with `org-with-wide-buffer` when saving the buffer.  This macri is included with Org:

```elisp
(defmacro org-with-wide-buffer (&rest body)
  "Execute body while temporarily widening the buffer."
  (declare (debug (body)))
  `(save-excursion
     (save-restriction
       (widen)
       ,@body)))
```

But it looks like it's not working for you.  Does adding an additional save-excursion on top fix it?

codeasone:
@karthink I re-evaluated with a [minimal dotemacs](https://gist.github.com/codeasone/4cf214c1645dbd8a979d4598da97e6c9) and couldn't reproduce the problem either, meaning it's an own goal related to the peculiarities of my broader configuration.  

Apologies for not assessing that before raising the PR. Thank you for developing `gptel` it's fitting into my workflow really well!

PR #126, opened 2023-11-02
Add custom gptel-update-destination

codeasone:
This PR proposes a customisable option `gptel-update-destination` to enable status messages to appear strictly in the minibuffer and never appear in the header line.

### DIFF
--- a/README.org
+++ b/README.org
@@ -344,7 +344,7 @@ Other Emacs clients for LLMs prescribe the format of the interaction (a comint s
 | =gptel-default-mode=          | Major mode for dedicated chat buffers. |
 | =gptel-prompt-prefix-alist=   | Text inserted before queries.          |
 | =gptel-response-prefix-alist= | Text inserted before responses.        |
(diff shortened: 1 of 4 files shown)


karthink:
Thank you for the PR, this looks useful!

My suggestions are minor: `gptel-update-destination` could be better named, and it would be better for the options to be symbols.  Perhaps `gptel-status-method`, or `gptel-display-method`?

While we're at it, we could also add  `'mode-line` as an option, and make the `gptel-mode` lighter dynamic.  Of course, many Emacs users hide all minor-mode mode-line indicators so this might not get much use.

karthink:
@codeasone I tweaked the idea slightly and merged. The user option is renamed to `gptel-use-header-line`, defaults to `t`. When set to `nil`, the mode-line is used to show in-progress requests (via `mode-line-process`, which is what the variable is meant for) and the echo area for error messages (as in your commit).

Thank you for the PR!

PR #127, opened 2023-11-03
Increased curl timeout

### LINKED ISSUE #125: Slow responses from GPT4All cause error parsing HTTP response
I am using GPT4All Desktop with `Vulkan on NVIDIA RTX A2000 8GB Laptop GPU`.

When I send a query with `gptel-send` I most often get a message like this in my `*Messages*` buffer.

```
ChatGPT response error: (((a5fc6c7e12b4c77eccc028d8aa342ca1 . 0)) Could not parse HTTP response.) Could not parse HTTP response.
```

I thought that errors I was seeing in GPT4All Desktop console logs indicated the query failed to be processed:

```
Error allocating memory ErrorOutOfDeviceMemory
error loading model: Error allocating vulkan memory.
llama_load_model_from_file: failed to load model
LLAMA ERROR: failed to load model from /home/nickanderson/.cache/gpt4all/mistral-7b-instruct-v0.1.Q4_0.gguf
```

However I since found the `Server Chat` in the UI and can see the queries sent from `gptel` and the responses that should come back.

image

```
Prompt: You are a Nexus, Nick Anderson's helpful assistant. You have dry humor. You always respond using org-mode syntax.

Complete this sentence: "I've got a lovely bunch of coconuts, " 
Response: and I'm ready to crack them open! Now if only someone could teach me to make piña colada without a C64 emulator...
Prompt: You are a Nexus, Nick Anderson's helpful assistant. You have dry humor. You always respond using org-mode syntax.

Complete this sentence: "I've got a lovely bunch of coconuts, " 
Response: "(inserting dry humor) I bet they all hate themselves."
```

I get responses back in emacs only when the response if fairly quick. With more context the response takes longer (especially if it offloads to CPU) and the response doesn't come back to emacs.

Digging around a bit I found the option for curls max-time here: https://github.com/karthink/gptel/blob/d7a89d7575e5d0f5148d7498536e470fda628f1e/gptel-curl.el#L58

I tweaked that up to 300 evaluated, tried a longer context message and it came back to emacs.

I think it would be nice if there were a variable that we could tweak for the timeout. Also it would be nice if we could set the default number of messages to send, currently it seems I need to set this via `gptel-menu` in each buffer I am using.

nickanderson:
Often local LLMs will offload a query to CPU if there is not enough VRAM or in
the case of an unsupported GPU. When a query is offloaded to the CPU responses
can be significantly slower. If curl times out early the user will not get the
response from the LLM back in Emacs.

This change increases the timeout for curl from 60s to 300s to make gptel usable
in slower environments.

Closes #125

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -55,7 +55,7 @@ PROMPTS is the data to send, TOKEN is a unique identifier."
      (list "--location" "--silent" "--compressed" "--disable"
            (format "-X%s" "POST")
            (format "-w(%s . %%{size_header})" token)
-           (format "-m%s" 60)
+           (format "-m%s" 300)
            "-D-"
            (format "-d%s" data))
      (when (not (string-empty-p gptel-proxy))


karthink:
Cheers @nickanderson!

PR #142, opened 2023-12-02
Add multiline prefixes & AI response prefixes

daedsidog:
This makes it so that prefixes with newlines are respected s.t. the point is moved to the end of the prefix:

image

Before, the point would be at the end of the first line. This saves two key presses.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -206,13 +206,24 @@ defaults to `text-mode'."
 
 ;; TODO: Handle `prog-mode' using the `comment-start' variable
 (defcustom gptel-prompt-prefix-alist
-  '((markdown-mode . "### ")
-    (org-mode . "*** ")
-    (text-mode . "### "))
+  '((markdown-mode . "## ")
+    (org-mode . "** ")
+    (text-mode . "## "))
   "String inserted after the response from ChatGPT.
 
-This is an alist mapping major modes to the prefix strings. This
+This is an alist mapping major modes to the prefix strings.  This
 is only inserted in dedicated gptel buffers."
+  :group 'gptel
+  :type '(alist :key-type symbol :value-type string))
+
+(defcustom gptel-response-prefix-alist
+  '((markdown-mode . "")
+    (org-mode . "")
+    (text-mode . ""))
+  "String inserted after the response from ChatGPT.
+
+This is an alist mapping major modes to the reply prefix strings.  This
+is only inserted in dedicated gptel buffers before the AI's response."
   :group 'gptel
   :type '(alist :key-type symbol :value-type string))
 
@@ -393,8 +404,11 @@ and \"apikey\" as USER."
      (skip-syntax-forward "w.")
      ,@body))
 
-(defun gptel-prompt-string ()
+(defun gptel-prompt-prefix-string ()
   (or (alist-get major-mode gptel-prompt-prefix-alist) ""))
+
+(defun gptel-response-prefix-string ()
+  (or (alist-get major-mode gptel-response-prefix-alist) ""))
 
 (defun gptel--restore-state ()
   "Restore gptel state when turning on `gptel-mode'.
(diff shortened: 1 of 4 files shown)


karthink (commented):
This buffer might already exist when running this code, so you don't want to go to `(point-max)`.  Instead, you can just remove  "\n" from line 955, so that it doesn't skip forward over newlines.

daedsidog:
Would it be better to export those chars to a symbol that can be configurable?

daedsidog:
I took the time to add support for more complex prefixes for the dedicated `gptel-mode`. I wanted to have prefixes for the AI responses as well as the user prompts so that it would be easier to organize a document from my end. It looks like this:

image

I made sure the parsing of the buffer works as expected when using either backend:

```
((:role "system" :content "You are a large language model living in Emacs and a helpful assistant. Respond concisely.") (:role "user" :content "Hello") (:role "assistant" :content "Hello! How can I assist you today?") (:role "user" :content "How are you?"))
```

Behavior is maintained for normal, inline prefixes:

image

```
((:role "system" :content "You are a large language model living in Emacs and a helpful assistant. Respond concisely.") (:role "user" :content "Hello") (:role "assistant" :content "Hello! How can I assist you today?") (:role "user" :content "How are you?"))
```

By default, responses don't have prefixes, and the behavior is unchanged unless configured:

image

```
((:role "system" :content "You are a large language model living in Emacs and a helpful assistant. Respond concisely.") (:role "user" :content "Hello") (:role "assistant" :content "Hello! How can I assist you today?") (:role "user" :content "How are you?"))
```

I tested with and without `curl`, though I did not test the other backends. I expect them to behave the same, since the only thing I changed is how the messages are being trimmed.

karthink:
Thanks for this PR, it's looking good!

I'm testing it right now.  We can merge  it soon if you think it's ready.

karthink:
Is there some reason you insert the response-prefix after the response is fully inserted instead of before?

daedsidog:
The original reason was that it was mostly a hack from my own configuration file that had that behavior with post hooks. The latest commit changes the behavior to insert the response prefix before the message starts.

karthink:
> Seems like you don't have [c6dd3e5](https://github.com/karthink/gptel/commit/c6dd3e5ddcc2f737028bb58fc81b72be02854283), then?

I do. Is [this](https://github.com/karthink/gptel/pull/142/commits/c6dd3e5ddcc2f737028bb58fc81b72be02854283#diff-3b65c1cc47ce5206ac9e29d41bac39285eb1fe9dc192603cb6120c752cb6b57bR222) the change you are referring to?

This inserts the response prefix after the response has been fully received.  Is there some reason you don't insert it before the response starts?

karthink:
> The latest commit changes the behavior to insert the response prefix before the message starts.

I'm not seeing this -- could you point me to it?

daedsidog:
The comment on that commit is wrong, I carelessly copied it from the previous location. It puts the prefix before the message has been received, not after.

daedsidog:
I was also not sure how to deal with the [skip-chars-backward](https://github.com/daedsidog/gptel/blob/be1d78bec19ca2bf12daf684a95f8d120b1db5b3/gptel.el#L973C6-L973C6), as I was not sure what's its purpose, so I just exported them to an internal variable without changing the default value. However, for multiline prefixes to be respected on startup, the `\n` needs to be removed, which I suggest doing, unless it would affect something else I don't know about.

karthink:
> It puts the prefix before the message has been received, not after. The current behavior is like this:

My bad, I had pulled [c6dd3e5](https://github.com/karthink/gptel/commit/c6dd3e5ddcc2f737028bb58fc81b72be02854283) but hadn't reloaded the file yet.

It works as expected.

> I was also not sure how to deal with the [skip-chars-backward](https://github.com/daedsidog/gptel/blob/be1d78bec19ca2bf12daf684a95f8d120b1db5b3/gptel.el#L973C6-L973C6), as I was not sure what's its purpose

It's there to clean things up when inserting the output when redirecting it from another buffer, usually from the transient menu.

- To fix the problem of it gobbling up whitespace from multiline prefixes, we can simply move the skip statement to before we insert the prefix:

```elisp
(goto-char (point-max))
(skip-chars-backward gptel--prefix-chars-to-skip)
(if (bobp) (insert (or initial (gptel-prompt-prefix-string))))
```

- There are no other uses of `gptel--prefix-chars-to-skip`, so I think making it a variable is unnecessary.

daedsidog:
Done. I should note that the README needs to be updated to mention the response prefixes.

gptel.el
```diff
@@ -950,9 +966,9 @@ buffer created or switched to."
       (visual-line-mode 1))
      (t (funcall gptel-default-mode)))
     (unless gptel-mode (gptel-mode 1))
-    (if (bobp) (insert (or initial (gptel-prompt-string))))
-    (goto-char (point-max))
     (skip-chars-backward "\t\r\n")
+    (if (bobp) (insert (or initial (gptel-prompt-prefix-string))))
+    (goto-char (point-max))
```
karthink, line 971: The `(goto-char (point-max))` should come first, before the `(skip-chars-backward...`.

daedsidog:
Done

karthink:
Looks good, thank you for the PR @daedsidog!

PR #145, opened 2023-12-07
gptel: Allow setting custom curl path

rleppink:
This commit allows the user to set a custom path to the curl executable.

Windows has some strange notions of what "curl" is, and this can't always be changed easily. With this variable, a user can let gptel use the curl executable of their choosing -- allowing full response streaming.

The default is still set to the same value it was: `"curl"`.

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -106,9 +106,9 @@ the response is inserted into the current buffer after point."
          (args (gptel-curl--get-args (plist-get info :prompt) token))
          (stream (and gptel-stream (gptel-backend-stream gptel-backend)))
          (process (apply #'start-process "gptel-curl"
-                         (generate-new-buffer "*gptel-curl*") "curl" args)))
+                         (generate-new-buffer "*gptel-curl*") gptel-curl-path args)))
     (when (eq gptel-log-level 'debug)
-      (gptel--log (json-encode (cons "curl" args))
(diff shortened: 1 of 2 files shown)


karthink:
@rleppink Thank you for the PR, I don't have a Windows PC right now to test `gptel`.

I think we can combine the two variables (`gptel-curl-path` and `gptel-use-curl`)  into one (`gptel-use-curl`), which will be set to the curl path (`(executable-find "curl")` by default instead of `t`.  Do you see any problems with this?

rleppink:
From an end user perspective, I personally think it's more clear that `use-curl` is a boolean value indicating whether to use it, and `curl-path` is a string to curl's path. 

Overloading `use-curl` to a string containing the path to curl seems a bit confusing for an end user.

karthink:
> From an end user perspective, I personally think it's more clear that `use-curl` is a boolean value indicating whether to use it, and `curl-path` is a string to curl's path.
> 
> Overloading `use-curl` to a string containing the path to curl seems a bit confusing for an end user.

Okay. We can go the other way and replace `gptel-use-curl` with `gptel-curl-path` then.  We can define `gptel-use-curl` as an `obsolete-variable-alias` of `gptel-curl-path`.

karthink:
@rleppink Are you still interested in adding a gptel-curl-path?

PR #146, opened 2023-12-07
Update available OpenAI GPT models to match API

moritzschaefer:
Since a month, OpenAI offers access to the faster and cheaper GPT4 turbo model. This PR replaces the non-existent GPT-4-32k model with that new existent model


karthink:
@moritzschaefer Thank you for keeping up with the API changes!

Issue #149, opened 2023-12-17
Add gemini backend support?


mrdylanyin:
Gemini is Google’s AI model and free for personal usage. User can get an API key via their [page on Google AI Studio](https://makersuite.google.com/app/apikey)

karthink:
Sure. Would you be interested in adding it?

mrdylanyin:
I really want to make this feature happen, but truth be told, I'm a newbie when it comes to writing Emacs Lisp. Spent a few hours digging through the source code, and I'm still clueless on where to begin. Apologies for that.

karthink:
@mrdylanyin Almost all of the code is LLM agnostic, so it will not need changing.  You also do not need to understand these parts to add a new backend. Assuming gemini has a http API and replies with JSON, there are only three API-specific functions (technically implementations of generic methods) needed:

- `gptel--parse-buffer`: This scans the buffer where you type and assembles a list of user prompts and LLM responses.
- `gptel--request-data`: This creates the full payload (to be JSON encoded and sent) including all model parameters required by the API.

If streaming responses are supported:
- `gptel-curl--parse-stream`: This extracts the content of the response from the JSON stream returned by the API.
If streaming responses are not supported:
- `gptel--parse-response`: This extracts the content of the response from the full JSON structure returned by the API.

You can look at the implementation of these methods (as `cl-defmethod`) in the files `gptel-openai.el` and `gptel-ollama.el`.  You might be able to get gemini working just by tweaking one of these.

PR #151, opened 2023-12-20
Add Gemini backend support

mrdylanyin:
#149


karthink:
This is lovely! Thank you for your work. I squashed your commits, worked on it a bit and pushed to master. I'm not sure why Github doesn't recognize the PR as approved. [Here's your commit](https://github.com/karthink/gptel/commit/84cd7bf5a467e279e676e2c71ae30b899fa20e60).

I added streaming (Google sends slower, bigger chunks of data than ChatGPT so it's not a huge improvement over the non-streaming API), support for temperature and max_tokens, and simplified the user configuration.

PR #159, opened 2023-12-27
Update README.org to set api key

vxe:


README.org
```diff
@@ -118,6 +118,9 @@ Optional: Set =gptel-api-key= to the key. Alternatively, you may choose a more s
 machine api.openai.com login apikey password TOKEN
   #+end_src
 - Setting it to a function that returns the key.
+  #+begin_src elisp
```
karthink, line 121: Did you mean to write a function that returns the key? This sets `gptel-api-key` to the API key string, not a function.

Issue #76, opened 2023-06-13
Add support for function calling handling


colonelpanic8:
See https://platform.openai.com/docs/guides/gpt/function-calling

would be cool to somehow add some support for this. Some inversion of control style kind of thing where you could set up the system prompt to be able to call emacs functions or something like that might be cool.

karthink:
Maybe I'm lacking imagination -- could you provide an example of the kinds of things you could do in Emacs with this? (Keeping in mind that this requires sending openAI the full type signatures of all the available functions.)

IvanMalison:
The main use case I have in mind is sort of cleaner manipulation of existing buffer content.

Right now, it can occasionally be kind of difficult to get chat gpts response to be something that is just a code completion, even if you prompt it with stuff like:

"You are a large language model and a careful programmer. Provide code that completes what is provided and only code as output without any additional text, prompt or note."

it still does not always only give the completion.

Furthermore, it could do pretty interesting stuff on larger segments of code, given the proper api, that could avoid it having to send back the entire text.

It might be possible to do something like provide it with editing commands that allow it to make edits to specific parts of a buffer or a selected region.

This is something that I think will become increasingly important as the context window is made larger and it can operate over larger and large segments of code bases.

karthink:
IIUC, you're describing the case of supplying ChatGPT the descriptions and type signatures of elisp functions, and then asking it to combine those (along with its general knowledge of elisp) to make edits to the buffer. I'm still unable to think of an actual use case in this context, since it already understands the elisp API primitives quite well. The fact that buffer editing elisp functions have to work by side-effect doesn't help either. Can you think of a more specific example like the weather report function in the documentation?

IvanMalison:
> IIUC, you're describing the case of supplying ChatGPT the descriptions and type signatures of elisp functions, and then asking it to combine those (along with its general knowledge of elisp) to make edits to the buffer.

Nope.

Did you read the text I wrote above? I don't think there would be any expectation that it directly call elisp arbitrary elisp functions. You would probably give it a relatively limited API that would simply allow it to specify edits to existing code:

As an example, you might give chat gpt functions like:

delete(start_line: int, start_character: int, end_line: int, end_character: int)
insert(start_line: int, start_character: int, text: string)

You would then handle these requests from chat gpt internally with elisp code.

The advantage would be that:

[...]

PR #209, opened 2024-02-07
[Looking for feedback] Add support for Function Calls

isaacphi:
# Demo

I made a quick screen recording showing the proof of concept:
https://screenapp.io/app/#/shared/6d146099-4bc3-4b34-afd7-bff3d16ee0ed

# Overview

I've added support for [Function Calling](https://platform.openai.com/docs/guides/function-calling) to gptel. This is just a proof of concept that I wanted to share before going any further.

First I want to thank you for making this package! It's become a seamless part of my emacs workflow because of how well it blends in, in particular using `gptel-send` for some of my own scripts. I think that supporting function calls would fit in well with the design philosophy of this package, allowing users to define their own callable functions for their own use cases. Some use cases I can think of off hand:
* Asking GPT to create a file or files for you based on the context of an above chat
* Requesting a diff of only specific lines of a file and interactively merging the diff
* Orchestrating "Retrieval Augmented Generation" including context from other files in your project or from the internet.
I think that because of how flexible emacs is, this could be the basis of a copilot that's more capable than any other editor.

I should note that function calls are not specific to OpenAI, they're available many if not all of the other models you support.

Before moving on, I'd like to know if this is even something you'd be interested in including in gptel. I'd also like to know your thoughts on how the user would use this feature. So far this is what I'm thinking:
* User defines a list of global functions that will be provided to the function api. Gptel can introspect these functions and automatically generate the structured information about them that the API requires.
* In the transient, function calls can be toggled on or off, or a specific function can be forced to be called (the API allows for this)

# How to test this PR
1. Run all the elisp in the file I included called `test-user-config.el`
2. Open a chat using the model I called "OpenAI with function calls" (I turned off streaming mode because I only support that for now)
3. Ask the chat to create new files or to have a cow say something. If it doesn't recognize your prompt as relating to either of those topics, it would work regularly.

Let me know what you think! Feel free to open a discussion if you think that's a more appropriate venue for this.
Related issue: #76

### DIFF
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -84,7 +84,7 @@
 (cl-defmethod gptel--request-data ((_backend gptel-openai) prompts)
   "JSON encode PROMPTS for sending to ChatGPT."
   (let ((prompts-plist
-         `(:model gptel-model
+         `(:model ,gptel-model
            :messages [,@prompts]
            :stream ,(or (and gptel-stream gptel-use-curl
                              (gptel-backend-stream gptel-backend))
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -84,9 +84,8 @@
 (cl-defmethod gptel--request-data ((_backend gptel-openai) prompts)
   "JSON encode PROMPTS for sending to ChatGPT."
   (let ((prompts-plist
-         `(:model ,gptel-model
+         `(:model gptel-model
            :messages [,@prompts]
-           ;; :response_format (:type "json_object")
            :stream ,(or (and gptel-stream gptel-use-curl
                              (gptel-backend-stream gptel-backend))
                         :json-false))))
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -127,7 +127,7 @@
     (name &key models stream key
           (header
            (lambda () (when-let (key (gptel--get-api-key))
-                        `(("Authorization" . ,(concat "Bearer " key))))))
+                    `(("Authorization" . ,(concat "Bearer " key))))))
           (host "api.openai.com")
           (protocol "https")
           (endpoint "/v1/chat/completions"))
@@ -170,7 +170,7 @@ function that returns the key."
     (prog1 backend
       (setf (alist-get name gptel--known-backends
                        nil nil #'equal)
-            backend))))
+                  backend))))
 
 ;;; Azure
 ;;;###autoload
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -72,46 +72,14 @@
        (goto-char (match-beginning 0))))
     (apply #'concat (nreverse content-strs))))
 
-;; (cl-defmethod gptel--parse-response ((_backend gptel-openai) response _info)
-;;  (map-nested-elt response '(:choices 0 :message :content)))
-
 (cl-defmethod gptel--parse-response ((_backend gptel-openai) response _info)
+  ;; If the reply specifies a function call, parse and return it instead of the message
   (let* ((choices-path '(:choices 0 :message))
          (tool-calls (map-nested-elt response (append choices-path '(:tool_calls))))
          (content (map-nested-elt response (append choices-path '(:content)))))
     (if tool-calls
         (prin1-to-string tool-calls)
       content)))
-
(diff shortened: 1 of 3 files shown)


gptel.el
```diff
@@ -386,6 +386,9 @@ To set the model for a chat session interactively call
           (const :tag "GPT 4 32k" "gpt-4-32k")
           (const :tag "GPT 4 1106 (preview)" "gpt-4-1106-preview")))
 
+(defcustom gptel-callable-functions nil
```
isaacphi, line 389: I wasn't sure where to put this

gptel-openai.el
```diff
@@ -73,16 +74,24 @@
     (apply #'concat (nreverse content-strs))))
 
 (cl-defmethod gptel--parse-response ((_backend gptel-openai) response _info)
-  (map-nested-elt response '(:choices 0 :message :content)))
+  ;; If the reply specifies a function call, parse and return it instead of the message
```
isaacphi, line 77: If GPT decides a function should be called, the structured data is included in `tool_calls` instead of `content` so that parsing needs to be updated.
For now I'm only parsing the full response, streaming is not supported yet (but could be)

gptel-openai.el
```diff
@@ -73,16 +74,24 @@
...
   (let ((prompts-plist
          `(:model ,gptel-model
            :messages [,@prompts]
            :stream ,(or (and gptel-stream gptel-use-curl
-                         (gptel-backend-stream gptel-backend))
-                     :json-false))))
+                             (gptel-backend-stream gptel-backend))
+                        :json-false))))
+    (when gptel-callable-functions
+      (plist-put prompts-plist :tools gptel-callable-functions))
```
isaacphi, line 94: This is the main important change. This could be toggled on or off based on a setting in the transient. A certain function call can also be forced by using the `tool_choice` parameter

test-user-config.el
```diff
@@ -0,0 +1,93 @@
...
+      (with-temp-buffer
+        (insert contents)
+        (write-file filename)))
+  (split-window)
+  (other-window 1)
+  (find-file filename))
+
+;;; Callable function schema
+(setq! gptel-callable-functions
+       ;; Hard coded variable specifying callable functions. This could be defined in a user's configuration
```
isaacphi, line 24: Given a list of emacs functions, we should be able to introspect them to generate this schema data if that's preferable.

test-user-config.el
```diff
@@ -0,0 +1,93 @@
...
+(defun cons-list-to-plist (cons-list)
+  (let ((plist '()))
+    (dolist (item cons-list)
+      (setq plist (plist-put plist (if (keywordp (car item))
+                                       (car item)
+                                     (intern (concat ":" (symbol-name (car item)))))
+                             (cdr item))))
+    plist))
+
+(cl-defun gptel-run-function-on-region (beg end)
```
isaacphi, line 71: For now this is just a post-response hook function, but this would likely be more appropriate in the core gptel source. I'm not sure exactly where though.

karthink:
This is fantastic -- and quite a small change too! Thanks for the PR, this looks quite promising.

I'm not familiar with the OpenAI function-calling API.  I'll take a look at it soon and then think about the UI, it'll probably be a few days from now.

cosmicz:
This PR would be nice to have for my use case. Anything I can do to push it along?

karthink:
> This PR would be nice to have for my use case. Anything I can do to push it along?

Function calling is planned to be part of the features I'm working on in the `feature-capabilities` branch.  This branch adds the infrastructure needed for per-model capability specification in gptel, as well as image support for supported models. The design is not final yet.

If you're interested, you could switch to the `feature-capabilities` branch and add a `tool-use` capability to the model specification.  Then we need to handle the different function calling APIs (as provided by OpenAI, Anthropic, Gemini, Ollama) and create a uniform interface for gptel.  I'm not sure how much of this PR can carry over as I haven't looked at it in a while.

karthink:
I've now implemented function calling - details in #514.  Looking for testers!

karthink:
Closing this PR as function calling is not available in gptel for all supported models.

PR #213, opened 2024-02-08
:recycle:: Untangle Gemini model and endpoint

cashpw:
gptel-gemini.el: There may soon be more than one Gemini model. This PR changes how the URL is built to support future additional model options.


karthink:
@cashpw Cheers!

PR #219, opened 2024-02-15
Avoid clashes with the custom directive key and use an arbitrary unused key in the worst case

martenlienen:
At the moment, the transient can assign `h` to one of your directives and then you need to use the mouse to customize directives on the fly. Also, if all characters of your directive name were already assigned, it would not assign any. With this PR, an arbitrary unused key will be assigned instead.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -185,22 +185,18 @@ which see."
   (transient-parse-suffixes
    'gptel-system-prompt
    (cl-loop for (type . prompt) in gptel-directives
-       with taken
+       ;; Avoid clashes with the custom directive key
+       with unused-keys = (delete ?h (number-sequence ?a ?z))
        with width = (window-width)
        for name = (symbol-name type)
-       for key =
-       (let ((idx 0) pos)
-         (while (or (not pos) (member pos taken))
-           (setq pos (substring name idx (1+ idx)))
-           (cl-incf idx))
-         (push pos taken)
-         pos)
+       for key = (seq-find (lambda (k) (member k unused-keys)) name (seq-first unused-keys))
+       do (setq unused-keys (delete key unused-keys))
        ;; The explicit declaration ":transient transient--do-return" here
        ;; appears to be required for Transient v0.5 and up.  Without it, these
        ;; are treated as suffixes when invoking `gptel-system-prompt' directly,
        ;; and infixes when going through `gptel-menu'.
        ;; TODO: Raise an issue with Transient.
-       collect (list (key-description key)
+       collect (list (key-description (list key))
                      (concat (capitalize name) " "
                              (propertize " " 'display '(space :align-to 20))
                              (propertize


gptel-transient.el
```diff
@@ -185,22 +185,18 @@ which see."
...
        for name = (symbol-name type)
-       for key =
-       (let ((idx 0) pos)
-         (while (or (not pos) (member pos taken))
-           (setq pos (substring name idx (1+ idx)))
-           (cl-incf idx))
-         (push pos taken)
-         pos)
+       for key = (seq-find (lambda (k) (member k unused-keys)) name (seq-first unused-keys))
+       do (setq unused-keys (delete key unused-keys))
```
karthink, line 193: This is going to cons up a ~26-element sequence as many times as there are directives, creating a bunch of garbage.  It's not much but it adds up (with all the other allocation gptel does), and this is why I chose to not do it this way.

gptel-transient.el
```diff
@@ -185,22 +185,18 @@ which see."
...
        with width = (window-width)
        for name = (symbol-name type)
-       for key =
-       (let ((idx 0) pos)
-         (while (or (not pos) (member pos taken))
-           (setq pos (substring name idx (1+ idx)))
-           (cl-incf idx))
-         (push pos taken)
-         pos)
+       for key = (seq-find (lambda (k) (member k unused-keys)) name (seq-first unused-keys))
```
karthink, line 192: The current method tries to provide mnemonic keybindings, using an alternative only in the event of a clash. IIUC this just assigns keys randomly to the directives?

Mnemonic keybindings are useful here (as opposed to just assigning them to 1 - 20) as they can be used by power users without requiring the transient menu to be shown. `(setq transient-show-popup nil)`

gptel-transient.el
```diff
@@ -185,22 +185,18 @@ which see."
   (transient-parse-suffixes
    'gptel-system-prompt
    (cl-loop for (type . prompt) in gptel-directives
-       with taken
+       ;; Avoid clashes with the custom directive key
+       with unused-keys = (remove ?h (number-sequence ?a ?z))
```
karthink, line 189: Thanks for catching this case! I forgot about `h` being taken.

karthink:
@martenlienen Thanks for the PR!

> Also, if all characters of your directive name were already assigned

Can you give me an example of how this can happen under the current scheme? I'm imagining that a combination (> 2) of similar directive names can lead to this, but I couldn't come up with an example.

gptel-transient.el
martenlienen: The doc says that `delete` works in-place like `delq` but you are supposed to reassign the result value anyway to make clear what happens, so that is what I did. I can also change it to `delq` if you prefer.

gptel-transient.el
martenlienen: This does the same but assigns a random (the first) unused key if all characters of the name have already been assigned to other directives, i.e. `seq-find` iterates through the characters of the name and returns the first that is `(member k unused-keys)`, defaulting to `(seq-first unused-keys)` if no key in `name` is unused.

One could think about if it makes sense to work with small and capital letters here, so that two directives `chat` and `complete` can be assigned `c` and `C` instead of `c` and `o`.

martenlienen:
Consider the case where you have three directives `a`, `b` and `ab`. When you get to `ab`, both characters have already been taken, so it continues to loop until you get
```elisp
ELISP> (substring "ab" 2 3)
*** Eval error ***  Args out of range: "ab", 2, 3
```
so it would actually error, I believe, instead of assigning no key.

gptel-transient.el
karthink: Just checked the implementation of `delete`, I think we're good here -- it appears to do additional consing if `unused-keys` is a vector, not a list.

gptel-transient.el
karthink: Thanks for the clarification.

I'm happy with using small letters consistently here.

gptel-transient.el
karthink: Could you use `delete` instead of `remove` here since `remove` just does `(delete elt (copy-sequence seq))`?  We don't need to generate a copy of the sequence.

karthink:
> Consider the case where you have three directives `a`, `b` and `ab`. When you get to `ab`, both characters have already been taken, so it continues to loop until you get
> 
> ```emacs-lisp
> ELISP> (substring "ab" 2 3)
> *** Eval error ***  Args out of range: "ab", 2, 3
> ```
> 
> so it would actually error, I believe, instead of assigning no key.

Ah right, thanks for clearing it up.

gptel-transient.el
martenlienen: Done

karthink:
@martenlienen Thanks for the PR! Merging now.

BTW I think there's one edge case still unaddressed: Consider the directives `a`, `b`, `ab` and `c`.  Per the current logic,

- `a` is assigned `a`
- `b` is assigned to `b`
- `ab` is assigned to `c` (the first unused letter)
- `c` is assigned to `d`

Ideally `c` would be assigned to `c` and `ab` to something else.  But I think we can live with this, it's exceedingly unlikely to happen.

PR #221, opened 2024-02-19
Add curl-args slot to gptel-backend

r0man:
This slot is used to append backend specific options to the curl arguments used to make HTTP requests. The backend specific arguments are appended after `gptel-curl--common-args`.

My use case for this feature is to set the `--cert` and `--key` options in a custom backend that uses mutal TLS to communicate with an OpenAI proxy/gateway.

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -66,6 +66,7 @@ PROMPTS is the data to send, TOKEN is a unique identifier."
       (gptel--log data "request body"))
     (append
      gptel-curl--common-args
+     (gptel-backend-curl-args gptel-backend)
      (list (format "-w(%s . %%{size_header})" token))
      (if (length< data gptel-curl-file-size-threshold)
          (list (format "-d%s" data))
--- a/gptel-gemini.el
+++ b/gptel-gemini.el
@@ -111,7 +111,7 @@
 
 ;;;###autoload
 (cl-defun gptel-make-gemini
-    (name &key header key (stream nil)
+    (name &key curl-args header key (stream nil)
           (host "generativelanguage.googleapis.com")
           (protocol "https")
           (models '("gemini-pro"))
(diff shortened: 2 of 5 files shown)


gptel-curl.el
```diff
@@ -66,6 +66,7 @@ PROMPTS is the data to send, TOKEN is a unique identifier."
       (gptel--log data "request body"))
     (append
      gptel-curl--common-args
+     (gptel-backend-curl-args gptel-backend)
```
r0man, line 69: We could also use the backend OR the gptel-curl--common-args. That way the backend has full control over the options, but also the liability to add all required options that are in gptel-curl--common-args. Wdyt?

karthink: Most of the arguments in `gtepl-curl--common-args` are critical for the functioning  of the callbacks/filter functions gptel uses. I think it's fine to leave these as "mandatory".

karthink:
Thank you, this looks good.  Unfortunately, we'll have to update the docstrings and signatures of all the constructors to include `curl-args`:
1. `gptel-make-openai` and `gptel-make-azure` in `gptel-openai.el`
2.`gptel-make-ollama` in `gptel-ollama.el`.
3. `gptel-make-kagi` in `gptel-kagi.el`.
4. `gptel-make-gemini` in `gptel-gemini.el`.

i can update 2 through 4 after the PR is merged.  Could you add it to the constructors in `gptel-openai.el`?

r0man:
Hi @karthink,

I now added an optional `curl-args` to all constructors and initialize the backends with it.

karthink:
Cheers @r0man!

r0man:
Hi @karthink, thanks for merging this. Can I ask you for 1 last thing: Could you please cut a release (tag a new version) with this change included? I would like to release the custom back-end I built on top of this to my colleagues. Some of them might use MELPA stable, and I would like them to be able to try it without any friction.
Thanks in advance, Roman.

r0man:
Perfect, thank you! :heart:

PR #258, opened 2024-03-17
Comment out responses in prog-mode

hexmode:
In any mode that is derived from prog-mode insert comment-start followed by a space.

Also add a hook to do this while streaming the response.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -281,12 +281,13 @@ defaults to `text-mode'."
   :group 'gptel
   :type 'symbol)
 
-;; TODO: Handle `prog-mode' using the `comment-start' variable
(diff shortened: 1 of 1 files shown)


karthink:
@hexmode Thanks for the PR.  I'm afraid I don't understand the intent: the prompt-prefix alist only applies to chat buffers (i.e. `gptel-mode`), and `gptel-mode` is not supported in `prog-mode` buffers.

karthink:
Closing as `gptel-prompt-prefix-alist` does not apply to `prog-mode` buffers.

For commenting out responses in `prog-mode` please use `gptel-post-response-functions` or `gptel-post-stream-hook`.

PR #259, opened 2024-03-18
gptel-transient: Also remove "-ts" suffix from major mode

fbergroth:
I just noticed the prompt saying "You are a python-ts programmer." :)


karthink:
@fbergroth Thank you!

But please note that the current "refactor" menu is obsolete and will be removed in the next release.  You can now rewrite/refactor and Ediff from the main gptel menu, and reuse an extensive system prompt along with a per-query refactor message.

To Ediff or switch between previous states you just have to place your cursor anywhere inside the response.
gptel-tweak-response

PR #267, opened 2024-03-22
Description in system directives; a summary for complex prompts

gregoryg:
I  add descriptions in my growing list of long complex gptel directives.

I find it helpful to use as a memory jog or summary of what long complex prompts are supposed to accomplish.

It has proven useful in a custom non-transient-mode function that uses annotations and looks like this:
completing-read-gptel-directives-with-description

Note that any prompt with no description set shows as "NO DESCRIPTION".

This change includes the definition needed for the customization framework, and still works with the existing transient selection framework.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -353,9 +353,9 @@ transient menu interface provided by `gptel-menu'."
 
 ;; Model and interaction parameters
 (defcustom gptel-directives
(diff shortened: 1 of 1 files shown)


gregoryg:
please hold off on this - in current version I see that transient mode selection is not working - I will investigate

gregoryg:
closing in order to do more testing with gptel-transient.el

PR #281, opened 2024-04-06
README: New pacakge based on gptel, magit-gptcommit

douo:
Add a reference to [magit-gptcommit](https://github.com/douo/magit-gptcommit) in the README's extensions section. magit-gptcommit is a package that depends on GPTel to generate Git commit messages within the magit-status buffer.


karthink:
Cheers, thanks @douo!

Issue #288, opened 2024-04-19
Make `gptel-expert-commands` customizable


swflint:
The variable `gptel-expert-commands` is described in #249, but is not discoverable (not described in `README.org` or shown in `M-x customize-group RET gptel`), but seems quite useful.  Would it make sense to make this a) customizable, and b) document it somewhere in README?

karthink:
`gptel-expert-commands` is not documented because it was a temporary measure and not intended to be used except for troubleshooting issues.  I plan to deprecate it in favor of transient levels `(info "(transient) Enabling and Disabling Suffixes")`, which will require no additional settings in gptel.

PR #289, opened 2024-04-19
gptel: make gptel-expert-commands customizable (#288)

swflint:
* gptel.el (gptel-expert-commands): Expert command availability is now discoverable.


swflint:
Given comment on #288, closing.

PR #256, opened 2024-03-16
LLM context aggregator

daedsidog:
This adds the contexter I've been working on for my own personal use to GPTel. As of now, this pull request is not complete because I'm unsure what's the best path to integrate it, and also how it will fit with the new UI refactor. I synced up with the master branch of GPTel, and just added that in, but my transient menu seems to be vastly different from the previews I've seen.

Because this started as part of my config & grew organically, there are still some features that are part of my config and not in GPTel. Those parts are added to the end of this pull request in the form of a snippet from my config.

In a nutshell, I personally find this to be an **incredibly** useful programming tool for productivity, more so than any other AI programming tool I've tried. It pampered me to the point where I don't think I can go back to programming without it.

Here is a short guide how to use it with my config (this will change the more things I'll offload to GPTel):

1. Select various regions in buffers and add them to the context buffer with `C-c g p`. You can press `c` to remove selected context snippets from the context buffer directly, or `C-c g p` again to remove context snippets from the code buffers.

context1

2. Write an instruction comment in code, then select it and use `C-c g r` to send it to GPTel.

context2

With Lisp, the contexter automatically collapses s-expressions in areas that they are not required in, such as functions with docstrings, thus saving tokens:

contexter3

Here is the part of my config file which sets all this up:

```emacs-lisp
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
;;; -------------------------------- GPTEL --------------------------------- ;;;
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

(use-package gptel-contexter
  :commands (gptel-context-string
             gptel-context-substring
             gptel-context-in-region
             gptel-pop-or-push-context
             gptel-remove-context))

(defun my/code-refactoring-prompt-string (code-regions)
  "Create a prompt for an LLM to refactor CODE-REGIONS in the current buffer."
  (let* ((context (gptel-context-string))
         (have-context (not (zerop (length context)))))
    (concat "I need you to refactor and/or write code. Write carefully with\
 great respect to existing syntax, styling, & conventions, this also includes\
 column width."
            (when have-context
              (concat "Listed below is context that may help you. Use it as a\
 read-only reference. You are strictly prohibited from changing it in any\
 way, even if asked directly."
[...]

PR #292, opened 2024-04-21
gptel: Add `gptel-add-context` function

munen:
Hi @karthink 

Coming back to our short discussion [5 hours ago](https://github.com/karthink/gptel/pull/256#issuecomment-2067771421). 

> I haven't looked around to see what other LLM clients offer since ~May 2023. Surely someone must have written something more featureful and seamless than gptel? A quick google search indicates that there are multiple (paid) commercial offerings for VSCode, I can't imagine them doing worse than gptel. This package is just a fancy wrapper around Curl.

Maybe I'm just an old, grumpy dude, but those commercial offerings that I checked out didn't hold their water. [Cody](https://github.com/sourcegraph/cody) was quite OK, but I didn't get it to work with my LLM API keys (for example OpenAI). It does work well with Claude 2.0 and ollama, though. 

Having said so, I prefer the functionality of gptel. I appreciate that it has a small and simple API surface which nonetheless includes all actually interesting interaction patterns. When adding Emacs and the GPL into the mix, it's a hard to beat mixture. There was only one important feature missing for me, so I built it. Couldn't have done that to any of those proprietary systems(;

> The main reason is that gptel is a one-weekend-per-month project for me and that time is eaten up fixing bugs! 

Thank you for your effort and wonderful project :pray: :bow:  

> There are three features on the roadmap that I haven't had the time to work on: attaching context (this PR), supporting function calling, and multimodal support (mainly vision) in chats. 

Happy to hear :+1: Maybe you'll like this implementation of attaching context. I tried to keep the implementation small, but flexible. I don't have much experience using it, so far, but I did dogfood on this PR from the very beginning(; Also, I'm looking very much forward to using it to tackle some bigger projects next week.

> Copilot-style completion-at-point, a fourth feature, seems very difficult to do via the Chat APIs so I've shelved that plan for now.

Agreed. I think it's hard to do with the chat APIs. In my tests of the proprietary alternatives, they also didn't really yield better results than actually using the chat modality (when given enough context). So I don't mind intentionally omitting this this feature.

> Yes, keeping gptel simple and focused is the only way for it to remain maintainable given my time constraints.

That's a smart choice! Apart from time, these AI based projects tend to have a lot of feature creep and a short half-life. I appreciate your effort to keeping this maintainable while at the same time making it all the more usable. Thank you :pray: 

> @munen Love your Emacs work and presentations! Thank you for organice as well, it's very handy!

Thank you for your kind words :pray: I didn't have too much time for FOSS lately, but this time was needed so that I can come back to it again in the long term. I'm looking forward to that time!


daedsidog:
Could perhaps benefit from the features I have on my PR, i.e. removing context, highlighting context, dedicated context buffer, etc? I think those don't really conflict with the vision of this library just so long as they are opt-in.

daedsidog:
#256 is close to fruition and I think it adds everything you want, making this PR obsolete.

karthink:
@munen A pretty expansive version of this feature has now been added, thanks to @daedsidog's efforts.  Please let me know after trying it if it's missing any features you might want.  If you have trouble finding/using the available features (after updating gptel), that will also serve as very useful feedback for us.

I will close this now, but I'm looking forward to how you find the feature, so please let us know in this thread!

karthink:
Dropping by here to mention that the "snapshot" context style favored by @munen is now also available in gptel.  You can specify ranges of text in `gptel-context` as (live) overlays, or as (static) bounds or line number ranges.  There is no UI for this, `gptel-add` continues to add overlays, but it is trivial to do this.  The docstring for `gptel-context` explains.

`gptel-context` can also be set buffer-locally, providing buffer-local context sets.

Finally, an example of specifying context on the fly, for use with on-the-fly presets in prompts:

```elisp
(gptel-make-preset 'visible-text
  :description "Include visible text from all windows in the frame."
  :context
  '(:eval
    (mapcar (lambda (win) ;; Create (<buffer> :bounds ((start . end)))
              (list (window-buffer win)
                    :bounds (list (cons (window-start win)
                                        (window-end win)))))
     ;; From all windows except the selected one and minibuffer
     (delq (unless (minibufferp) (selected-window))
           (window-list)))))
```

This adds (a snapshot of) the visible text in all windows on frame to the context, using simple position bounds.  Usable in prompts as, for example:

``` text
@visible-text This is an Emacs frame.  What do you see?
```

PR #301, opened 2024-05-01
Replace window-width with frame-width

RNAer:
To avoid truncate of the text for the vertically split window like this:

Screenshot 2024-05-01 at 12 29 56 PM

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -270,7 +270,7 @@ Also format its value in the Transient menu."
      (string-replace
       "\n" "⮐ "
       (truncate-string-to-width
-       gptel--system-message (max (- (window-width) 12) 14) nil nil t)))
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR!

Actually neither `window-width` nor `frame-width` are correct here, since the width of the transient window depends on the value of `transient-display-buffer-action`.  For example, transient menus appear below the active window for me, which means they are only as wide as the window, not the frame:

2024-05-02-113244_1427x744_scrot

So a better fix is needed.

RNAer:
Thanks! good to learn a new thing `transient-display-buffer-action`. `org-roam` has a similar problem with `org-roam-node-find` cmd, even though it is minibuffer not transient.

karthink:
Created an issue so this PR can be closed.

Issue #305, opened 2024-05-05
Interacting with privateGPT (specifically for RAG)


Aquan1412:
Hello, I'm looking for a way to use privateGPT as a backend for gptel, in order to use its simple RAG pipeline. 
Specifically, I'm looking for a way to provide additional keywords to the backend (such as "use_context" and "include_sources").

I can generally use privateGPT as an "openai"-like backend with the following configuration:

```
(gptel-make-openai "privateGPT"
  :protocol "http"
  :host "localhost:8001"
  :models '("private-gpt"))
```

I tried simply adding the keyword to the configuration:

```
(gptel-make-openai "privateGPT"
  :protocol "http"
  :host "localhost:8001"
  :use_context t
  :models '("private-gpt"))
```

However, if I do that, I get the following error message:

>   Debugger entered--Lisp error: (error "Keyword argument :use_context not one of (:curl-ar...")
>   signal(error ("Keyword argument :use_context not one of (:curl-ar..."))
>   error("Keyword argument %s not one of (:curl-args :models..." :use_context)
>   gptel-make-openai("privateGPT-Context" :protocol "http" :host "localhost:8001" :use_context t :models ("private-gpt"))
>   (progn (gptel-make-openai "privateGPT-Context" :protocol "http" :host "localhost:8001" :use_context t :models '("private-gpt")))
>   eval((progn (gptel-make-openai "privateGPT-Context" :protocol "http" :host "localhost:8001" :use_context t :models '("private-gpt"))) t)
>   elisp--eval-last-sexp(nil)
>   eval-last-sexp(nil)
>   funcall-interactively(eval-last-sexp nil)
>   call-interactively(eval-last-sexp nil nil)
>   command-execute(eval-last-sexp)

So, is there a way to get this to work? Or am I approaching it completely wrong?

karthink:
You're going to have to provide more context for me to understand what you're looking for.  What do you expect `use_context` and `include_sources` to do?

Aquan1412:
I want to use privateGPT to generate responses based on embeddings I previously generated by "ingesting" several PDFs (in privateGPT lingo). According to the [API reference](https://docs.privategpt.dev/api-reference/api-reference/contextual-completions/chat-completion-v-1-chat-completions-post-stream), if I set `use_context = true`, it should use the embeddings to generate responses based on the ingested papers. Additionally, if I set `include_sources = true`, it should include the source chunks, on which the generated answers are based.

karthink:
[...]

PR #312, opened 2024-05-12
Support for privateGPT, including support for RAG functions like rspecting context and parsing of sources

Aquan1412:
Hi, as suggested in our recent discussion #305, here's a PR that adds support for interacting with privateGPT. Its based on the openai-backend, with some additions. The focus was to enable the use of two specific key words, as described in the [API reference](https://docs.privategpt.dev/api-reference/api-reference/contextual-completions/chat-completion-v-1-chat-completions-post-stream):
1) `use_context`: directs the LLM to use the context of documents that have been "ingested"
2) `include_sources`: directs the LLM to also return information about the sources of the context (so far: file name and page number, if appropriate).

I changed the code of `gptel-openai.el`to send the additional keywords to the llm server, and to correctly parse the answer if sources are provided. 

I tested the code with different queries, and so far it works for my use case. However, there is at least one missing feature: currently, the two new keywords `use_context` and `include_sources` are hardcoded to `t`. It would probably be better to make them configurable when creating the backend. Unfortunately, I'm not sure how to do that correctly.

### DIFF
--- a/gptel-privategpt.el
+++ b/gptel-privategpt.el
@@ -1,0 +1,163 @@
+;;; gptel-privategpt.el ---  Privategpt AI suppport for gptel  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2023  Karthik Chikmagalur
+
+;; Author: Karthik Chikmagalur <karthikchikmagalur@gmail.com>
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; This file adds support for Privategpt's Messages API to gptel
+
+;;; Code:
+(require 'cl-generic)
+(eval-when-compile
+  (require 'cl-lib))
+(require 'map)
+(require 'gptel)
+
+(defvar json-object-type)
+
+(declare-function prop-match-value "text-property-search")
+(declare-function text-property-search-backward "text-property-search")
+(declare-function json-read "json" ())
+
+
+
+;;; Privategpt (Messages API)
+(cl-defstruct (gptel-privategpt (:constructor gptel--make-privategpt)
+                               (:copier nil)
+                               (:include gptel-openai))
+  context sources)
+
+(defun gptel--privategpt-parse-sources (response)
+  (cl-loop with source-alist
+           for source across (map-nested-elt response '(:choices 0 :sources))
+           for name = (map-nested-elt source '(:document :doc_metadata :file_name))
+           for page = (map-nested-elt source '(:document :doc_metadata :page_label))
+           do (push page (alist-get name source-alist nil nil #'equal))
+           finally return
+           (cl-loop for (file-name . file-pages) in source-alist
+                    for pages = (delete-dups (delq nil file-pages))
+                    if pages
+                    collect (format "- %s (page %s)" file-name (mapconcat #'identity pages ", "))
+                    into source-items
+                    else collect (format "- %s" file-name) into source-items
+                    finally return (mapconcat #'identity (cons "\n\nSources:" source-items) "\n"))))
+
+(cl-defmethod gptel-curl--parse-stream ((_backend gptel-privategpt) info)
+  (let* ((content-strs))
+    (condition-case nil
+        (while (re-search-forward "^data:" nil t)
+          (save-match-data
+            (if (looking-at " *\\[DONE\\]")
+                (when-let ((sources-string (plist-get info :sources)))
+                  (push sources-string content-strs))
+              (let ((response (gptel--json-read)))
+		(unless (or (plist-get info :sources)
+                            (not (gptel-privategpt-sources (plist-get info :backend))))
+                  (plist-put info :sources (gptel--privategpt-parse-sources response)))
+		(let* ((delta (map-nested-elt response '(:choices 0 :delta)))
+		       (content (plist-get delta :content)))
+		  (push content content-strs))))))
+    (error
+     (goto-char (match-beginning 0))))
+  (apply #'concat (nreverse content-strs))))
+
+(cl-defmethod gptel--parse-response ((_backend gptel-privategpt) response info)
+  (let ((response-string (map-nested-elt response '(:choices 0 :message :content)))
+        (sources-string (and (gptel-privategpt-sources (plist-get info :backend))
+                             (gptel--privategpt-parse-sources response))))
+    (concat response-string sources-string)))
+
+(cl-defmethod gptel--request-data ((_backend gptel-privategpt) prompts)
+  "JSON encode PROMPTS for sending to ChatGPT."
+  (let ((prompts-plist
+         `(:model ,gptel-model
+	   :messages [,@prompts]
+	   :use_context ,(or (gptel-privategpt-context gptel-backend) :json-false)
+	   :include_sources ,(or (gptel-privategpt-sources gptel-backend) :json-false)
+           :stream ,(or (and gptel-stream gptel-use-curl
+                         (gptel-backend-stream gptel-backend))
+                     :json-false))))
+    (when gptel-temperature
+      (plist-put prompts-plist :temperature gptel-temperature))
+    (when gptel-max-tokens
+      (plist-put prompts-plist :max_tokens gptel-max-tokens))
+    prompts-plist))
+
+
+;;;###autoload
+(cl-defun gptel-make-privategpt
+    (name &key curl-args stream key
+          (header
+           (lambda () (when-let (key (gptel--get-api-key))
+			`(("Authorization" . ,(concat "Bearer " key))))))
+          (host "localhost:8001")
+          (protocol "http")
+	  (models '("private-gpt"))
+          (endpoint "/v1/chat/completions")
+          (context t) (sources t))
+  "Register an Privategpt API-compatible backend for gptel with NAME.
+
+Keyword arguments:
+
+CURL-ARGS (optional) is a list of additional Curl arguments.
+
+HOST (optional) is the API host, \"api.privategpt.com\" by default.
+
+MODELS is a list of available model names.
+
+STREAM is a boolean to toggle streaming responses, defaults to
+false.
+
+PROTOCOL (optional) specifies the protocol, https by default.
+
+ENDPOINT (optional) is the API endpoint for completions, defaults to
+\"/v1/messages\".
+
+HEADER (optional) is for additional headers to send with each
+request. It should be an alist or a function that retuns an
+alist, like:
+((\"Content-Type\" . \"application/json\"))
+
+KEY is a variable whose value is the API key, or function that
+returns the key.
+
+CONTEXT and SOURCES: if true (the default), use available context
+and provide sources used by the model to generate the response."
+  (declare (indent 1))
+  (let ((backend (gptel--make-privategpt
+                  :curl-args curl-args
+                  :name name
+                  :host host
+                  :header header
+                  :key key
+                  :models models
+                  :protocol protocol
+                  :endpoint endpoint
+                  :stream stream
+                  :url (if protocol
+                           (concat protocol "://" host endpoint)
+                         (concat host endpoint))
+                  :context context
+                  :sources sources)))
+    (prog1 backend
+      (setf (alist-get name gptel--known-backends
+                       nil nil #'equal)
+                  backend))))
+
+(provide 'gptel-privategpt)
+;;; gptel-backends.el ends here
--- a/README.org
+++ b/README.org
@@ -4,23 +4,24 @@
 
 gptel is a simple Large Language Model chat client for Emacs, with support for multiple models and backends.
 
-| LLM Backend        | Supports | Requires                  |
-|--------------------+----------+---------------------------|
-| ChatGPT            | ✓      | [[https://platform.openai.com/account/api-keys][API key]]                   |
-| Azure              | ✓      | Deployment and API key    |
-| Ollama             | ✓      | [[https://ollama.ai/][Ollama running locally]]    |
-| GPT4All            | ✓      | [[https://gpt4all.io/index.html][GPT4All running locally]]   |
-| Gemini             | ✓      | [[https://makersuite.google.com/app/apikey][API key]]                   |
-| Llama.cpp          | ✓      | [[https://github.com/ggerganov/llama.cpp/tree/master/examples/server#quick-start][Llama.cpp running locally]] |
-| Llamafile          | ✓      | [[https://github.com/Mozilla-Ocho/llamafile#quickstart][Local Llamafile server]]    |
-| Kagi FastGPT       | ✓      | [[https://kagi.com/settings?p=api][API key]]                   |
-| Kagi Summarizer    | ✓      | [[https://kagi.com/settings?p=api][API key]]                   |
-| together.ai        | ✓      | [[https://api.together.xyz/settings/api-keys][API key]]                   |
-| Anyscale           | ✓      | [[https://docs.endpoints.anyscale.com/][API key]]                   |
-| Perplexity         | ✓      | [[https://docs.perplexity.ai/docs/getting-started][API key]]                   |
-| Anthropic (Claude) | ✓      | [[https://www.anthropic.com/api][API key]]                   |
-| Groq               | ✓      | [[https://console.groq.com/keys][API key]]                          |
-| OpenRouter         | ✓      | [[https://openrouter.ai/keys][API key]]                          |
+| LLM Backend        | Supports | Requires                   |
+|--------------------+----------+----------------------------|
+| ChatGPT            | ✓        | [[https://platform.openai.com/account/api-keys][API key]]                    |
+| Azure              | ✓        | Deployment and API key     |
+| Ollama             | ✓        | [[https://ollama.ai/][Ollama running locally]]     |
+| GPT4All            | ✓        | [[https://gpt4all.io/index.html][GPT4All running locally]]    |
+| Gemini             | ✓        | [[https://makersuite.google.com/app/apikey][API key]]                    |
+| Llama.cpp          | ✓        | [[https://github.com/ggerganov/llama.cpp/tree/master/examples/server#quick-start][Llama.cpp running locally]]  |
+| Llamafile          | ✓        | [[https://github.com/Mozilla-Ocho/llamafile#quickstart][Local Llamafile server]]     |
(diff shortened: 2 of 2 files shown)


karthink:
@Aquan1412  Thanks for the PR.

I had a look at the code, It uses free/global variables and uses some `cl-lib` functions that aren't available at runtime.  I can clean up the code but I need some sample privategpt output to be sure I don't break anything.

Could you run `(setq gptel-log-level 'info)`, use privategpt so it generates some sources, and paste the results of the `*gptel-log*` buffer here?

Aquan1412:
Thanks for cleaning up the code, I'm unfortunately not very experienced writing elisp code.

I ran the command you asked for, however, the output is quite large, which is why I appended it as a file to this comment.

The user query and the expected response are as follows:

> *** What are the Navier-Stokes equations?

> The Navier-Stokes equations are a set of partial differential equations that describe the conservation of mass, momentum, and energy for a viscous fluid. They are given by equation (2.1) for mass conservation, equation (2.2) for momentum conservation, and equation (2.3) for energy conservation in compressible form. The pressure p, density ρ, velocity components ui, vi, wi, specific internal energy e, specific enthalpy h, temperature T, effective stress tensor τij, and convective heat flux qj are involved in these equations. For gases, the ideal gas law (2.4) relates pressure p to density ρ and specific volume RT/γ, where R is the gas constant and γ is the ratio of specific heats. The specific internal energy e and enthalpy h can be expressed as functions of temperature T using equations (2.5).
> 
> Sources:
> - Beneddine-2017-Characterization_of_unsteady_flow_behavior_by_linear_stability_analysis.pdf (page 62)
> - Iorio-2015-Global_Stability_Analysis_of_Turbulent_Transonic_Flows_on_Airfoil_geometries.pdf (page 33) 
> 
[gptel-privategpt-output.txt](https://github.com/karthink/gptel/files/15362594/gptel-privategpt-output.txt)

karthink:
From what I can see, you're not actually buffering the sources from each response.  `persistent-sources` in your code is set to the sources included in the last-but-one response chunk.  Is this intentional?

Aquan1412:
As far as I'm aware,the list of sources is the same across all chunks. I didn't find an explicit confirmation in the privateGPT API description, but in all my tests it was like that.
Therefore I only always reset the chunk, to avoid parsing the list of sources for each chunk.

karthink:
Wow, that's really wasteful API design. It's sending about 200-1000x as
much JSON as required then.

karthink:
I cleaned up the code and pushed a couple of commits to your branch.  I also made "use_context" and "include_sources" configurable.  Please check if this introduced any bugs, as I can't test it.

Aquan1412:
Great, thanks! Unfortunately, I'm currently travelling and won't have access to my computer for the next two weeks. But afterwards I'll test it as soon as possible.

Aquan1412:
I finally found some time to test your changes. There are currently two issues:
1) If `:stream` is set to `t`, there is an error with concatenating the sources to the response (it works for `:stream nil`):

> error in process filter: apply: Wrong type argument: characterp, "

This should be straight forward enough to fix,I think, as soon as I find a bit of time.
2) If either `:context` or `:sources` is set to `nil`, I get the following error (independent of the value of `:stream`):

>  gptel response error: ((HTTP/1.1 422 Unprocessable Entity) Could not parse HTTP response.) Could not parse HTTP response.

Here, I'm not sure what the reason is. Any idea?

karthink:
> 2) If either `:context` or `:sources` is set to `nil`, I get the following error (independent of the value of `:stream`):
> 
> > gptel response error: ((HTTP/1.1 422 Unprocessable Entity) Could not parse HTTP response.) Could not parse HTTP response.
> 
> Here, I'm not sure what the reason is. Any idea?

Can you generate a log for this case? `(setq gptel-log-level 'info)`, then produce this error and check `*gptel-log*`.

karthink:
Made a couple of changes based on guesses.  Please test again and try to use the gptel log buffer to get more details on the errors.

gptel-privategpt.el
```diff
@@ -0,0 +1,163 @@
...
+    (name &key curl-args stream key
+          (header
+           (lambda () (when-let (key (gptel--get-api-key))
+			`(("Authorization" . ,(concat "Bearer " key))))))
+          (host "localhost:8001")
+          (protocol "http")
+	  (models '("private-gpt"))
+          (endpoint "/v1/chat/completions")
+          (context t) (sources t))
+  "Register an Privategpt API-compatible backend for gptel with NAME.
```
RFDonovan, line 113: Just noticed a typo : "_a PrivateGPT_" . Probably because of "**API**"

Aquan1412:
> Made a couple of changes based on guesses. Please test again and try to use the gptel log buffer to get more details on the errors.

With these new changes I get an "empty response". I attached the output of the gptel-log-buffer, as it is once again quite lengthy.
[gptel-privategpt-output_latest-commit.txt](https://github.com/user-attachments/files/15777793/gptel-privategpt-output_latest-commit.txt)

Aquan1412:
> > 2. If either `:context` or `:sources` is set to `nil`, I get the following error (independent of the value of `:stream`):
> > 
> > > gptel response error: ((HTTP/1.1 422 Unprocessable Entity) Could not parse HTTP response.) Could not parse HTTP response.
> > 
> > 
> > Here, I'm not sure what the reason is. Any idea?
> 
> Can you generate a log for this case? `(setq gptel-log-level 'info)`, then produce this error and check `*gptel-log*`.

I also went back and reran the command while generating a log. Here the output was quite short:

> {
  "gptel": "request body",
  "timestamp": "2024-06-10 20:36:33"
}
{
  "model": "private-gpt",
  "messages": [
    {
      "role": "system",
      "content": "You are a large language model living in Emacs and a helpful assistant. Respond concisely."
    },
    {
      "role": "user",
      "content": "What are the Navier-Stokes equations?"
    }
  ],
  "use_context": true,
  "include_sources": null,
  "stream": true,
  "temperature": 1.0
}
{
  "gptel": "response body",
  "timestamp": "2024-06-10 20:36:33"
}
{
  "detail": [
    {
      "type": "bool_type",
      "loc": [
        "body",
        "include_sources"
      ],
      "msg": "Input should be a valid boolean",
      "input": null,
      "url": "https://errors.pydantic.dev/2.5/v/bool_type"
    }
  ]
}

karthink:
Are you sure the changes were applied?

```json
{
    "model": "private-gpt",
    "messages": [
        {
            "role": "system",
            "content": "You are a large language model living in Emacs and a helpful assistant. Respond concisely."
        },
        {
            "role": "user",
            "content": "What are the Navier-Stokes equations?"
        }
    ],
    "use_context": true,
    "include_sources": null,
    "stream": true,
    "temperature": 1.0
}
```

In this request, `"include_sources"` should be set to `false`, not `null`, if the changes are applied.

karthink:
I made a dummy privategpt backend and tested it, and `"include_sources"` is indeed set to `false`.

Aquan1412:
> > 2. If either `:context` or `:sources` is set to `nil`, I get the following error (independent of the value of `:stream`):
> > 
> > > gptel response error: ((HTTP/1.1 422 Unprocessable Entity) Could not parse HTTP response.) Could not parse HTTP response.
> > 
> > 
> > Here, I'm not sure what the reason is. Any idea?
> 
> Can you generate a log for this case? `(setq gptel-log-level 'info)`, then produce this error and check `*gptel-log*`.

I also went back and reran the command while generating a log. Here the output was quite short:

> {
  "gptel": "request body",
  "timestamp": "2024-06-10 20:36:33"
}
{
  "model": "private-gpt",
  "messages": [
    {
      "role": "system",
      "content": "You are a large language model living in Emacs and a helpful assistant. Respond concisely."
    },
    {
      "role": "user",
      "content": "What are the Navier-Stokes equations?"
    }
  ],
  "use_context": true,
  "include_sources": null,
  "stream": true,
  "temperature": 1.0
}
{
  "gptel": "response body",
  "timestamp": "2024-06-10 20:36:33"
}
{
  "detail": [
    {
      "type": "bool_type",
      "loc": [
        "body",
        "include_sources"
      ],
      "msg": "Input should be a valid boolean",
      "input": null,
      "url": "https://errors.pydantic.dev/2.5/v/bool_type"
    }
  ]
}


> Are you sure the changes were applied?
> 
> ```json
> {
>     "model": "private-gpt",
>     "messages": [
>         {
>             "role": "system",
>             "content": "You are a large language model living in Emacs and a helpful assistant. Respond concisely."
>         },
>         {
>             "role": "user",
>             "content": "What are the Navier-Stokes equations?"
>         }
>     ],
>     "use_context": true,
>     "include_sources": null,
>     "stream": true,
>     "temperature": 1.0
> }
> ```
> 
> In this request, `"include_sources"` should be set to `false`, not `null`, if the changes are applied.

Sorry, I wasn't clear enough: I went back to the previous commit and ran the command with logging turned on, in case it would help.

Aquan1412:
Also, I found (and fixed, I hope) an error in your latest commit. Now it is possible to correctly toggle `:context` and `:sources`.

karthink:
> Also, I found (and fixed, I hope) an error in your latest commit. Now it is possible to correctly toggle `:context` and `:sources`.

It looks like you reverted a change I made, to use the `backend` that is passed to the parsing functions.  You are using `gptel-backend` instead -- this is an error.

`gptel-backend` can be set buffer-locally, and the parsing code runs in the HTTP request buffer, so it can have the wrong value.  This is why the `backend` is explicitly passed to the parsing functions.

karthink:
> Sorry, I wasn't clear enough: I went back to the previous commit and ran the command with logging turned on, in case it would help.

I'm confused -- when `:sources` is set to `nil`, is `"include_sources"` set to `null` or `false` in the latest commit?

gptel-privategpt.el
karthink: Thanks, will fix this before merging.

Aquan1412:
> > Sorry, I wasn't clear enough: I went back to the previous commit and ran the command with logging turned on, in case it would help.
> 
> I'm confused about "previous commit" now. As per the latest commit, when `:sources` is set to `nil` in the privategpt backend definition, is `"include_sources"` set to `null` or `false` when sending the request?

It is set to `false`.

Aquan1412:
Now I also fixed the problem that the `sources-string` was not properly appended to the response if `:stream t`.
So now everything seems to work as it should, all backend-settings (`:stream`, `:context` and `:sources`) can be toggled on and off without problems.

gptel-privategpt.el
```diff
@@ -0,0 +1,163 @@
...
+  (let* ((content-strs))
+    (condition-case nil
+        (while (re-search-forward "^data:" nil t)
+          (save-match-data
+            (if (looking-at " *\\[DONE\\]")
+                (when-let ((sources-string (plist-get info :sources)))
+                  (push sources-string content-strs))
+              (let ((response (gptel--json-read)))
+		(unless (or (plist-get info :sources)
+                            (not (gptel-privategpt-sources gptel-backend)))
```
karthink, line 70: This should use the passed argument `backend` (renamed from `_backend`), not `gptel-backend`, for reasons explained in the main thread.

gptel-privategpt.el
```diff
@@ -0,0 +1,163 @@
...
+		(let* ((delta (map-nested-elt response '(:choices 0 :delta)))
+		       (content (plist-get delta :content)))
+		  (push content content-strs))))))
+    (error
+     (goto-char (match-beginning 0))))
+  (apply #'concat (nreverse content-strs))))
+
+(cl-defmethod gptel--parse-response ((_backend gptel-privategpt) response _info)
+  (let ((response-string (map-nested-elt response '(:choices 0 :message :content)))
+        (sources-string (and (gptel-privategpt-sources gptel-backend)
```
karthink, line 81: This should use the passed argument `backend` (renamed from `_backend`), not `gptel-backend`, for reasons explained in the main thread.

gptel-privategpt.el
```diff
@@ -0,0 +1,163 @@
...
+                             (gptel--privategpt-parse-sources response))))
+    (concat response-string sources-string)))
+
+(cl-defmethod gptel--request-data ((_backend gptel-privategpt) prompts)
+  "JSON encode PROMPTS for sending to ChatGPT."
+  (let ((prompts-plist
+         `(:model ,gptel-model
+	   :messages [,@prompts]
+	   :use_context ,(or (gptel-privategpt-context gptel-backend) :json-false)
+	   :include_sources ,(or (gptel-privategpt-sources gptel-backend) :json-false)
```
karthink, line 91: This should use the passed argument `backend` (renamed from `_backend`), not `gptel-backend`, for reasons explained in the main thread.

Aquan1412:
> > Also, I found (and fixed, I hope) an error in your latest commit. Now it is possible to correctly toggle `:context` and `:sources`.
> 
> It looks like you reverted a change I made, to use the `backend` that is passed to the parsing functions. You are using `gptel-backend` instead -- this is an error.
> 
> `gptel-backend` can be set buffer-locally, and the parsing code runs in the HTTP request buffer, so it can have the wrong value. This is why the `backend` is explicitly passed to the parsing functions.

The problem with using `backend` (or `_backend`)  is that I get the following error: 

> error in process sentinel: Wrong type argument: gptel-privategpt, nil

This is why I changed it back to `gptel-backend`.  I also looked at some of the other implementations (specifically `gptel-openai` and `gptel-ollama`, and there you also still use `gptel-backend`, so I thought it was OK. 

The full log output is as follows:
> {
  "gptel": "request body",
  "timestamp": "2024-06-16 21:26:55"
}
{
  "model": "private-gpt",
  "messages": [
    {
      "role": "system",
      "content": "You are a large language model living in Emacs and a helpful assistant. Respond concisely."
    },
    {
      "role": "user",
      "content": "What are the Navier-Stokes equations?"
    }
  ],
  "use_context": true,
  "include_sources": true,
  "stream": false,
  "temperature": 1.0
}
{
  "gptel": "response body",
  "timestamp": "2024-06-16 21:26:59"
}
{
  "id": "fe28e561-95e0-4e37-91c0-1bf2defc1173",
  "object": "completion",
  "created": 1718566019,
  "model": "private-gpt",
  "choices": [
    {
      "finish_reason": "stop",
      "delta": null,
      "message": {
        "role": "assistant",
        "content": " The Navier-Stokes equations are a set of partial differential equations that describe the motion of viscous fluid substances, including air and water. They were formulated by Claude Louis Marie Henri Navier and George Gabriel Stokes in the early 19th century. These equations are fundamental to fluid dynamics and are widely used in various fields such as meteorology, aeronautics, hydrodynamics, and many more."
      },
      "sources": [
        {
          "object": "context.chunk",
          "score": 0.5979778911410051,
          "document": {
            "object": "ingest.document",
            "doc_id": "f0fa7ff2-9c83-46f7-9412-0752af9d96ac",
            "doc_metadata": {
              "window": "Treaty Org.\n ˚Akervik E, Brandt L, Henningson DS, Hœpffner J, Marxen O, Schlatter P.  2006.  Steady solutions of the\nNavier-Stokes equations by selective frequency damping.  Phys.  Fluids 18:068102\n˚Akervik E, Hœpffner J, Ehrenstein U, Henningson DS.  2007. ",
              "original_text": "Steady solutions of the\nNavier-Stokes equations by selective frequency damping. ",
              "page_label": "28",
              "file_name": "Theofilis-2011-Global_Linear_Instability.pdf",
              "doc_id": "f0fa7ff2-9c83-46f7-9412-0752af9d96ac"
            }
          },
          "text": "Treaty Org.\n ˚Akervik E, Brandt L, Henningson DS, Hœpffner J, Marxen O, Schlatter P.  2006.  Steady solutions of the\nNavier-Stokes equations by selective frequency damping.  Phys.  Fluids 18:068102\n˚Akervik E, Hœpffner J, Ehrenstein U, Henningson DS.  2007. ",
          "previous_texts": null,
          "next_texts": null
        },
        {
          "object": "context.chunk",
          "score": 0.573735513098284,
          "document": {
            "object": "ingest.document",
            "doc_id": "080a7865-1a7f-49d8-8781-e991f04563c0",
            "doc_metadata": {
              "window": "J. Fluid Mech.  493:31–58\nOosterlee CW, Wesseling P, Segal A, Brakkee E.  1993.  Benchmark solutions for the incompressible Navier-\nStokes equations in general coordinates on staggered grids.  Int.  J. Numer.  Methods Fluids 17:301–21\nPawlowski RP, Salinger AG, Shadid JN, Mountziaris TJ. ",
              "original_text": "Benchmark solutions for the incompressible Navier-\nStokes equations in general coordinates on staggered grids. ",
              "page_label": "32",
              "file_name": "Theofilis-2011-Global_Linear_Instability.pdf",
              "doc_id": "080a7865-1a7f-49d8-8781-e991f04563c0"
            }
          },
          "text": "J. Fluid Mech.  493:31–58\nOosterlee CW, Wesseling P, Segal A, Brakkee E.  1993.  Benchmark solutions for the incompressible Navier-\nStokes equations in general coordinates on staggered grids.  Int.  J. Numer.  Methods Fluids 17:301–21\nPawlowski RP, Salinger AG, Shadid JN, Mountziaris TJ. ",
          "previous_texts": null,
          "next_texts": null
        }
      ],
      "index": 0
    }
  ]
}

gptel-privategpt.el
Aquan1412: See comment in the main thread.

gptel-privategpt.el
Aquan1412: See comment in the main thread.

gptel-privategpt.el
Aquan1412: See comment in the main thread.

karthink:
> The problem with using `backend` (or `_backend`) is that I get the following error:
> 
> > error in process sentinel: Wrong type argument: gptel-privategpt, nil
> 
> This is why I changed it back to `gptel-backend`. I also looked at some of the other implementations (specifically `gptel-openai` and `gptel-ollama`, and there you also still use `gptel-backend`, so I thought it was OK.

Those are errors too, actually, I need to fix that.  However, you'll notice that the backend is not used (as either `backend` or `gptel-backend`) when parsing responses for any of the other APIs, in `gptel--parse-response` or `gptel--curl-parse-stream`.  This is where things can go wrong if the wrong backend is used, and it's why this bug has gone unnoticed. It's only used when creating the request, in `gptel--request-data`, where `backend` and `gptel-backend` typically have the same value.

> error in process sentinel: Wrong type argument: gptel-privategpt, nil

Anyway, I'm trying to figure out why you're getting that error :thinking:

karthink:
I fixed the bug that was causing `backend` to be `nil`, rewrote that part of the response handling (`gptel--parse-buffer` and `gptel-curl--parse-stream`), and rebased on master.  Could you update and check if it works as intended now?

Aquan1412:
> I fixed the bug that was causing `backend` to be `nil`, rewrote that part of the response handling (`gptel--parse-buffer` and `gptel-curl--parse-stream`), and rebased on master. Could you update and check if it works as intended now?

I just checked it, and everything works as intended. Thanks for the effort!

karthink:
That's great, we can merge it then.  Can you add an entry to the README with instructions for using PrivateGPT?  You can copy the text from one of the other sections (like Perplexity or OpenRouter) and modify it, with additional notes on using `:sources` and `:context`.

Aquan1412:
Done. I added the usage instructions and also included a link to the privateGPT repository for reference.

karthink:
Cheers @Aquan1412! Thanks for the PR.

PR #313, opened 2024-05-13
Add new model: GPT-4o

axelknock:
OpenAI has released a new model called GPT-4o, or "GPT-4 Omni". More information [here](https://openai.com/index/hello-gpt-4o/).

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -401,6 +401,7 @@ The current options for ChatGPT are
 - \"gpt-3.5-turbo\"
 - \"gpt-3.5-turbo-16k\"
 - \"gpt-4\"
+- \"gpt-4o\"
 - \"gpt-4-turbo\"
 - \"gpt-4-turbo-preview\"
 - \"gpt-4-32k\"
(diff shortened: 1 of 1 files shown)


karthink:
@axelknock thanks for the PR!  I'm aware of gpt-4o but haven't added it because I'm not sure about this.  As I understand, gpt-4o can respond to text queries with images, like Dall-E etc.  Is this true?  gptel is not set up to handle base64-encoded images in responses, so this can break things.

axelknock:
Currently image generation is unavailable, and I could find no definitive answer on its future broad availability. I looked at the Dalle-3 documentation and found that the API can respond with both an image or a URL for an image, depending on the [response format parameter](https://platform.openai.com/docs/guides/images/usage?lang=curl).

If this parameter were available on the 4o response, would always sending it as URL be an option?

karthink:
Okay, we'll worry about it when image generation with gpt-4o is turned on.  If we can't fix it easily I can make gptel display a warning when the model is chosen.

Thanks again for the PR!

PR #329, opened 2024-06-18
new flag introduced: it forces prompt context that is sent to backend, be more explicit

vibrys:
@karthink, thank you very much for the great gptel package. It's great value is noticeable.

I'm issuing PR to prevent users from accidental sending too many data during requesting from org mode buffer.

The idea is to force both branching and topic be present on the path. If they both are included, then one may assume the user set these two intentionally thus sends that part of buffer (s)he wants.

the setting by default is nil (nothing changes to existing users), but IMO it may be considered to be set to true as default.

please let me know what you think about the change.

kind regards,
Vibrys

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -114,6 +114,12 @@ This makes it feasible to have multiple conversation branches."
   :type 'boolean
   :group 'gptel)
 
+(defcustom gptel-org-force-branching-context-with-topic nil
+  "force the existence both the topic and branching context"
+  :local t
+  :type 'boolean
+  :group 'gptel)
+
 
 ;;; Setting context and creating queries
 (defun gptel-org--get-topic-start ()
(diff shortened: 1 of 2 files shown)


vibrys:
maybe the idea of explicit context should be spread to other major modes as well.
If yes, then if the `gptel-org-force-branching-context-with-topic` flag is true, then:
1. for org-mode the behavior is just as in this pull request
2. for other modes it's enough that some region is marked.

if this is good idea, then `gptel-org-force-branching-context-with-topic` should be renamed, for example to: `gptel-force-context`   ?

regards,
Vibrys

vibrys:
I've just made the setting more generic, just as described in the message above. Flag name: gptel-force-prompt-context-be-explicit

karthink:
This restriction doesn't make sense to me.  gptel is intended to be used freely in any buffer in Emacs.

- If you are unsure about what will be sent, you can use gptel's dry-run to examine the query. `(setq gptel-expert-commands t)`.
- If you want to limit the size of the text to be sent, you can advise gptel-send, count the number of words in the prompt and cancel the query if required.

Closing as not planned.

PR #331, opened 2024-06-20
Add `claude-3-5-sonnet-20240620` to gptel-anthropic.el

erwald:
This model was recently released by Anthropic, and is now Anthropic's strongest model.
See https://docs.anthropic.com/en/docs/about-claude/models


karthink:
@erwald Thanks!

Issue #249, opened 2024-03-13
Separate system prompts and directives


jwr:
This is a feature request based on several days of using gptel with various models. I am not trying to implement this and submit a pull request, because I think this needs designing first, and it's likely my design choices would be different.

Current problems:

* there is no separation between the task/directive prompt (e.g. "Rewrite this paragraph") and the system prompt (providing context)
* there is no way to pick from one of system prompts (contexts) and separately pick one of several tasks to perform
* large prompts are not easy to handle (for example, they expand the minibuffer after modifying them)

My suggestion:

Make the system prompt(s) separate from task/directive prompts, so that I can pick a system prompt and then easily work with several directives in the context of that system prompt.

Context:

I usually provide a large system prompt with a lot of context, containing for example a full description of my application, along with the requested writing style. The directives are fairly simple: expand this, rewrite that, summarize this, etc. and they change very often, almost every invocation requires editing them. This is difficult if directives are combined with the context.

When working with code, I also have a general set of rules which do not change often, and then the exact directive is something I edit on the spot.

In both cases, having a separate set of system prompts and directives would make working through gptel more convenient.

karthink:
> * there is no separation between the task/directive prompt (e.g. "Rewrite this
>   paragraph") and the system prompt (providing context)
> 
> * there is no way to pick from one of system prompts (contexts) and separately
>   pick one of several tasks to perform

Yes, this has been bothering me for months now, it was just low on the priority list.

> * large prompts are not easy to handle (for example, they expand the minibuffer
>   after modifying them)

Fixed.  Are there any other issues with handling large prompts?

> Context: ...

I agree, the system prompt is for shaping the conversation as a whole, and we need an easy way to specify per-message instructions.  So far I've just been adding this before/after the buffer contents of interest.  For instance, when rewriting/refactoring I just add a comment above the region of interest explaining what I want it to do.

This was my original plan, to just use the buffer.  But it doesn't always work, like when using read-only buffers, and even when it's possible I often have to manually erase my instructions from the buffer afterwards.  So it might be worth adding a directives option.  There are many UI considerations, however.  Here are some questions, let me know what you think.

[...]

PR #339, opened 2024-07-03
Allow disabling system message

janEbert:
Means that it can be set to `nil` and won't be supplied in requests.

(Requested here: https://github.com/karthink/gptel/issues/249#issuecomment-2205680676)

### DIFF
--- a/gptel-gemini.el
+++ b/gptel-gemini.el
@@ -110,7 +110,10 @@
                   (list :text (string-trim
                                (buffer-substring-no-properties (point-min) (point-max)))))
             prompts))
-    (cl-callf (lambda (msg) (concat gptel--system-message "\n\n" msg))
+    (cl-callf (lambda (msg)
+                (if gptel--system-message
+                    (concat gptel--system-message "\n\n" msg)
+                  msg))
         (thread-first (car prompts)
                       (plist-get :parts)
                       (plist-get :text)))
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -58,6 +58,9 @@ Meant to be called when `gptel-menu' is active."
                        (concat "\n\n" (substring s 1))))
                   args))
 
+(defconst gptel--no-system-message-description "[No system message]"
+  "Description displayed when no system message is active.")
+
 (defun gptel--instructions-make-overlay (text &optional ov)
   "TODO"
   (save-excursion
@@ -282,10 +285,12 @@ Also format its value in the Transient menu."
   ;; :incompatible '(("-m" "-n" "-k" "-e"))
   [:description
    (lambda ()
-     (string-replace
-      "\n" "⮐ "
-      (truncate-string-to-width
-       gptel--system-message (max (- (window-width) 12) 14) nil nil t)))
+     (if gptel--system-message
+         (string-replace
+          "\n" "⮐ "
+          (truncate-string-to-width
+           gptel--system-message (max (- (window-width) 12) 14) nil nil t))
+       gptel--no-system-message-description))
    [""
     "Instructions"
     ("s" "Set system message" gptel-system-prompt :transient t)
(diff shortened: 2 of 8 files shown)


tshu-w:
@karthink I just wanted to kindly remind you about this PR, in case it may have slipped your attention.

gptel-gemini.el
```diff
@@ -110,7 +110,10 @@
                   (list :text (string-trim
                                (buffer-substring-no-properties (point-min) (point-max)))))
             prompts))
-    (cl-callf (lambda (msg) (concat gptel--system-message "\n\n" msg))
+    (cl-callf (lambda (msg)
```
karthink, line 113: Instead of this, you can avoid the whole `cl-callf` statement if no system message is available.

gptel-transient.el
```diff
@@ -397,20 +402,27 @@ Also format its value in the Transient menu."
...
-                             (propertize " " 'display '(space :align-to 20))
-                             (propertize
-                              (concat
-                               "("
-                               (string-replace
-                                "\n" " "
-                                (truncate-string-to-width prompt (- width 30) nil nil t))
-                               ")")
-                              'face 'shadow))
+                     (if prompt
```
karthink, line 405: I think it should be possible to disable the system message from the transient menu.  It should probably be a standalone option in the menu, bound to `<delete>` or something.

gptel-transient.el
```diff
@@ -806,7 +821,8 @@ Or in an extended conversation:
              :position position
              :in-place (and in-place (not output-to-other-buffer-p))
              :stream stream
-             :system (concat gptel--system-message system-extra)
+             :system (when gptel--system-message
```
karthink, line 824: If there's no system message, it should still be possible to add an additional directive from the transient menu (the `d` option captured in `system-extra` here).  This will now function as the whole system message.

gptel.el
```diff
@@ -374,7 +374,8 @@ transient menu interface provided by `gptel-menu'."
   '((default . "You are a large language model living in Emacs and a helpful assistant. Respond concisely.")
     (programming . "You are a large language model and a careful programmer. Provide code and only code as output without any additional text, prompt or note.")
     (writing . "You are a large language model and a writing assistant. Respond concisely.")
-    (chat . "You are a large language model and a conversation partner. Respond concisely."))
+    (chat . "You are a large language model and a conversation partner. Respond concisely.")
+    (none))
```
karthink, line 378: Why is `none` an option here?  There appears to be no way to actually choose this to select `nil` as the value of `gptel--system-message`.

gptel.el
```diff
@@ -968,7 +970,8 @@ Model parameters can be let-bound around calls to this function."
   (declare (indent 1))
   (let* ((gptel--system-message
           ;Add context chunks to system message if required
-          (if (and gptel-context--alist
+          (if (and system
```
karthink, line 973: If there's no system message but there are context chunks (and `gptel-use-context` is set to `system`), they should still be supplied as the context.

karthink (commented):
There are three sources that combine to form the system message in gptel:

1. `gptel--system-message`,
2. the "additional directive" option for quick instructions from the transient menu, and
3. the context chunks added via `gptel-add`, assuming `gptel-use-system` is set to `system`.

Sources 2 and 3 have to be explicitly specified, and source 1 is automatic.  This PR should not disable 2 and 3 when disabling 1.

janEbert:
Hey, just a quick heads-up, I haven't forgotten about this, but vacation off the computer and other priorities kept me from implementing your improvements. Thank you already for the review!

guibor:
Is there anything one can do to help to test this?

karthink:
> Is there anything one can do to help to test this?

It would be best to rebase this PR onto the `feature-capabilities` branch, which affects a fair bit of gptel's system message handling.  This branch (which adds image support and the requirements for supporting other features) will be merged into master soon.

karthink:
I've added the ability to set the system message to `nil`.  It was quite a bit more work than this PR since the transient menus, context handling and "additional directive" logic all had to be changed.

Anyway, if you're still interested in this feature please test it, and create an issue if it does not work as expected.

Closing this PR now.

PR #341, opened 2024-07-15
Make markdown to org conversion optional

ultronozm:
This adds an option `gptel-curl-convert-markdown-to-org` that allows the user to disable the conversion from markdown to org, which can be convenient in some situations (e.g., when the responses are streamed to part of an existing org buffer)

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -92,6 +92,10 @@ REQUEST-DATA is the data to send, TOKEN is a unique identifier."
               collect (format "-H%s: %s" key val))
      (list url))))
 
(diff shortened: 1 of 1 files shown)


karthink:
@ultronozm Thanks for the PR.  I think a better way to do this is to use a processing hook that the user can customize.  Then they can include/exclude the Org conversion, and also add additional processing functions to transform the incremental responses however they would like.

karthink:
I've added an option `gptel-org-convert-response` that does this, so this PR is no longer required.  This is still an interim solution while I figure out how to provide a filtering hook where the user can specify whatever processing they (don't) want applied to the response.

PR #345, opened 2024-07-22
Clean up response before ediffing

### LINKED ISSUE #344: Filter out language header for ediff session
It seems gptel produces an ediff session that is not usable because it doesn't filter out programming language header/footer, e.g "```javascript" from the ChatGPT output. See the screenshot.

image

hron:
ChatGPT response contains three backticks with programming language name at the first and last lines which prevents ediff from working properly.

Fixes #344

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -100,7 +100,7 @@ Or is it the other way around?"
 (defun gptel--rewrite-message ()
   "Set a generic refactor/rewrite message for the buffer."
   (if (derived-mode-p 'prog-mode)
-      (format "You are a %s programmer. Refactor the following code. Generate only code, no explanation."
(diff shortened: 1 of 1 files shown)


karthink:
Thanks! Not all models and backends do this, so unconditionally deleting lines will not work.  It's better to simply change the system message used with the request to specify that only code/prose should be returned, without any code fences or delimiters.

hron:
> Thanks! Not all models and backends do this, so unconditionally deleting lines will not work. It's better to simply change the system message used with the request to specify that only code/prose should be returned, without any code fences or delimiters.

I see. I have updated the generic rewrite message for programming modes. It works well!

karthink:
Cheers @hron!

PR #349, opened 2024-07-26
Add better default context string

daedsidog:
While gptel provides the tools to make any arbitrary context format, I think it's important to have a very general and all-encompassing default one.

I find that the old one was too confusing for many models and a lot of the times they included the line information and ellipses in the result, simply because those were embedded within the code block.

This patch makes the context string slightly more general and removes any embedded information that might be confusing to the models while still providing line information:

image

This one also supports context with nested Markdown backticks.

### DIFF
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -254,48 +254,92 @@ START and END signify the region delimiters."
 
 (defun gptel-context--insert-buffer-string (buffer contexts)
   "Insert at point a context string from all CONTEXTS in BUFFER."
-    (let ((is-top-snippet t)
-          (previous-line 1))
-      (insert (format "In buffer `%s`:" (buffer-name buffer))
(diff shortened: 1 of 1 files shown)


karthink:
Thanks.  From a cursory look at the changes I don't understand what it's doing.  I'll take a detailed look later, I'm pretty sure you don't need all all those flags and checks to do what you want.

karthink:
Also is the column number information ever useful?

daedsidog:
> Also is the column number information ever useful?

It's only there if the context is actually not from the start of the line. I imagine that it could be confusing if you select multiple contexts from the same line and it would show the same line number (confusing for the model, not the user). Not that it wouldn't be able to deal with it or anything, but at the same time I didn't really think it would hurt.

I have no real way to gauge usefulness beyond that.

I do know however that having non-embedded line numbers is vastly better, because now the models stopped returning them in their results.

PR #357, opened 2024-08-06
Utilizing the original buffer as ediff buffer in eglot-mode

luluman:
Eglot employs track-changes to monitor buffer modifications. However, track-changes is unable to account for cloned buffers, resulting in inaccurate status updates when changes occur in a cloned buffer.

This patch directly uses the original buffer in ediff mode, effectively resolving the issue. While this approach narrows the buffer, potentially causing confusion for some users, the buffer is restored to its original state after exiting ediff.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -1046,6 +1046,7 @@ When LOCAL is non-nil, set the system message only in the current buffer."
             (lambda ()
               (when (window-configuration-p cwc)
                 (set-window-configuration cwc))
+              (ediff-kill-buffer-carefully "*gptel-rewrite-Region.B-*")
               (remove-hook 'ediff-quit-hook gptel--ediff-restore))))
     (message "Waiting for response... ")
     (gptel-request
@@ -1071,7 +1072,9 @@ When LOCAL is non-nil, set the system message only in the current buffer."
              (add-hook 'ediff-quit-hook gptel--ediff-restore)
              (apply
               #'ediff-regions-internal
-              (get-buffer (ediff-make-cloned-buffer gptel-buffer "-Region.A-"))
+              (if (and (require 'eglot nil t) (eglot-managed-p))
+                  gptel-buffer
+                (get-buffer (ediff-make-cloned-buffer gptel-buffer "-Region.A-")))
               (car gptel-bounds) (cdr gptel-bounds)
               new-buf new-beg new-end
               nil


luluman:
I use GPTel to refactor my code with the eDiff interactive patch. The default implementation of GPTel refactoring employs a cloned buffer of the original region. However, since my project is tracked by Eglot and Flymake, changes in the cloned buffer disrupt Flymake status and trigger numerous errors. Worse, Flymake cannot recover from these errors, forcing me to restart Eglot, which can take a considerable amount of time to get Clangd back up and running.

To address this, I propose a solution that uses the original buffer if it is managed by Eglot. The patch I’m providing is not optimal, as it checks Eglot status, introducing a dependency. However, considering Eglot is already a built-in package in Emacs, this dependency is trivial for some Emacs users. Alternatively, we could remove this check altogether and always use the original buffer in all situations.

karthink:
@luluman reviewing this is on the docket, but I need to do a fair bit of work first to improve gptel's rewrite-and-diff feature.  If you're still interested please reopen the PR.

karthink:
I've redesigned the refactoring interface from scratch, this problem with LSP should not happen any more.  Please update and try when you can.

luluman:
Thank you for the new design! The updated interface is incredibly cool and powerful, particularly with its increased "async" capabilities for refactoring. It offers more flexibility for handling multiple rewrites simultaneously and allows for finer-grained tuning.

However, I noticed that when there are many rewritten parts in the buffer, I can only apply changes one at a time before moving to the next and triggering the review system. Could this process be made more automatic? For instance, allowing users to review all changes at once or providing additional keybindings for navigating to the next or previous refactor part would be immensely helpful.

Additionally, the new design has resolved the LSP issues I encountered with `ediff-region`. Thanks again for your hard work!

karthink:
> However, I noticed that when there are many rewritten parts in the buffer, I can only apply changes one at a time before moving to the next and triggering the review system. Could this process be made more automatic? For instance, allowing users to review all changes at once

Just open the refactor menu again when there are reviews pending, you can do a buffer-wide review of all the changes then. See the options available in the transient menu:

screenshot_20240905T172052

> providing additional keybindings for navigating to the next or previous refactor part would be immensely helpful.

I've made a note, I'll add some keybindings.

karthink:
Added navigation commands.  They're bound to `C-c C-n` and `C-c C-p` inside the overlays by default, and unbound elsewhere in the buffer.  The eldoc and tooltip hints don't list these keybindings since the message looks very busy, but users looking for it shouldn't have much trouble finding them.

PR #358, opened 2024-08-07
Fix gptel-context-add handling of region boundary when choosing a buffer

jdormit:
When gptel-context-add was called with a positive prefix argument to select a buffer to add to the context, it was using `point-min` and `point-max` of the _current buffer_ to determine the region of context, instead of the target buffer. This meant the context didn't contain the whole target buffer. The fix is straightforward: just make sure to use the target buffer's `point-min` and `point-max` instead.


karthink:
Cheers @jdormit!

PR #369, opened 2024-08-28
Change order of Refactor sentences in gptel--rewrite-message

nfedyashev:
This change I believe is more useful for a typical user in the following common use case

I as a user would like not to refactor but to rewrite in a specific way mentioning what is important. With the existing default prompt I need to formulate it like "Refactor/rewrite in a way that {details}" repeating "Refactor" twice in different places. With "Refactor the following code." being a suffix, I just need to press Backspace once removing dot and start typing {details} right away.

"Generate only code, no explanation" chunk is a part that rarely gets updated so it might be better to keep it in the middle because users won't need to access it often.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -100,7 +100,7 @@ Or is it the other way around?"
 (defun gptel--rewrite-message ()
   "Set a generic refactor/rewrite message for the buffer."
   (if (derived-mode-p 'prog-mode)
-      (format "You are a %s programmer. Refactor the following code. Generate only code, no explanation, no code fences."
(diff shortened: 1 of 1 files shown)


karthink:
Looks like a good change, thank you.

Note that the whole refactoring interface is due for an overhaul since I'd like to make it (i) more streamlined in how the changes are applied, and (ii) make it easier to add context when refactoring.  This way of specifying refactoring instructions was always meant to be a stopgap and will probably change.

karthink:
...and the refactoring menu has been overhauled.  The refactor-message prompt is still unchanged, I will update it too.  But please try updating gptel and trying some refactoring (from the transient menu as usual), I'd appreciate your feedback for the redesign.

PR #386, opened 2024-09-16
Add Github Models

gs-101:
Besides supporting the Azure SDK, Github Models also supports OpenAI. I tried the [same setup](https://github.com/karthink/gptel/discussions/385#discussioncomment-10664014) with `gptel-make-azure` but had no success.


karthink:
Cheers @gs-101!

PR #387, opened 2024-09-18
README: Update reference to consult-web

gs-101:
consult-web is now deprecated in favor of [consult-omni](https://github.com/armindarvish/consult-omni).


karthink:
Thanks @gs-101.  While you're at it, could you change the description to something like "Consult Omni is a versatile multi-source search package for Emacs.  It includes gptel as one of the sources for queries."

gs-101:
I also noticed that some of the bullet points (-) are missing commas. Don't think it's a big issue but it would be nice to standardise them in some way (but in a separate commit, since this is for only consult-web/omni).

karthink:
Thank you @gs-101!

Issue #237, opened 2024-03-07
Weird interplay with purpose mode


xuan-w:
When I invoke (gptel) directly, I got the following error:

It seems that this has something to do with purpose mode (I use Spacemacs, it includes purpose mode). Since purpose mode hasn't been updated for 3 years and the only thing I updated was gptel, I guess gptel introduced some changes? 

Here is the error I encountered: 

```
Debugger entered--Lisp error: (void-function user-action-sequence)
  user-action-sequence(#<buffer *ChatGPT*> ((user-action-sequence user-action-sequence pop-to-buffer)))
  purpose--action-function(#<buffer *ChatGPT*> nil)
  #<subr display-buffer>(#<buffer *ChatGPT*> ((user-action-sequence pop-to-buffer)) nil)
  purpose-display-buffer-advice(#<subr display-buffer> #<buffer *ChatGPT*> ((user-action-sequence pop-to-buffer)) nil)
  apply(purpose-display-buffer-advice #<subr display-buffer> (#<buffer *ChatGPT*> ((user-action-sequence pop-to-buffer))))
  display-buffer(#<buffer *ChatGPT*> ((user-action-sequence pop-to-buffer)))
  #<subr pop-to-buffer>(#<buffer *ChatGPT*> ((user-action-sequence pop-to-buffer)) nil)
  purpose-pop-to-buffer-advice(#<subr pop-to-buffer> #<buffer *ChatGPT*> ((user-action-sequence pop-to-buffer)) nil)
  apply(purpose-pop-to-buffer-advice #<subr pop-to-buffer> (#<buffer *ChatGPT*> ((user-action-sequence pop-to-buffer))))
  pop-to-buffer(#<buffer *ChatGPT*> ((user-action-sequence pop-to-buffer)))
  purpose--action-function(#<buffer *ChatGPT*> nil)
  #<subr display-buffer>(#<buffer *ChatGPT*> (pop-to-buffer) nil)
  purpose-display-buffer-advice(#<subr display-buffer> #<buffer *ChatGPT*> (pop-to-buffer) nil)
  apply(purpose-display-buffer-advice #<subr display-buffer> (#<buffer *ChatGPT*> (pop-to-buffer)))
  display-buffer(#<buffer *ChatGPT*> (pop-to-buffer))
  gptel("*ChatGPT*" "sk-***********************************************..." nil t)
  funcall-interactively(gptel "*ChatGPT*" "sk-***********************************************..." nil t)
  command-execute(gptel record)
[...]

PR #412, opened 2024-10-10
Update Spacemacs installation instructions

jaeyeom:
Spacemacs has an `llm-client` layer that supports `gptel`. This layer also resolves (in Spacemacs) GH-237 by disabling purpose mode for `gptel`, providing a working out-of-the-box experience for Spacemacs users.


karthink:
That's great! Thanks for the PR @jaeyeom

PR #413, opened 2024-10-10
Consistently use "last user prompt" for where context may be included

tschwinge:
Sometimes we used unqualified "the user prompt" (or "the last prompt", once), which is imprecise in multi-turn conversations with several assistant/user messages.

Accordingly adjust 'gptel--wrap-user-prompt' -> 'gptel--wrap-last-user-prompt' (without backwards-compatibility fallback).

No change in behavior intended.

### DIFF
--- a/gptel-anthropic.el
+++ b/gptel-anthropic.el
@@ -146,8 +146,8 @@ format."
                                        &optional inject-media)
   "Wrap the last user prompt in PROMPTS with the context string.
 
-If INJECT-MEDIA is non-nil wrap it with base64-encoded media
-files in the context."
+If INJECT-MEDIA is non-nil wrap the first user prompt in PROMPTS
+with base64-encoded media files in the context."
   (if inject-media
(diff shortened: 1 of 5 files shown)


tschwinge:
> Accordingly adjust 'gptel--wrap-user-prompt' -> 'gptel--wrap-last-user-prompt' (without backwards-compatibility fallback).

That change now doesn't make sense anymore: as of commit 16da7eb4b70f1a5996a6974dad990e11101a92a3 "gptel-context: Include images via `gptel-add'", "If INJECT-MEDIA [...] Wrap **the first user prompt** with included media files/contexts".

tschwinge:
@karthink, let me know if you see value in any of the commits that I've now force-pushed.  Otherwise close this as "(largely) obsoleted by recent changes".

karthink:
@tschwinge Thanks for the PR.  The plan is to allow injecting the context into the first or last prompt as required, so the rename doesn't make sense.

Closing as not planned.

PR #414, opened 2024-10-10
Allow for prepending (instead of appending) context to the system message

tschwinge:
... via new 'system-prepend'.

Certain use cases/LLMs may prefer to first see the context, then the system message.

### DIFF
--- a/gptel-anthropic.el
+++ b/gptel-anthropic.el
@@ -101,7 +101,7 @@
             prompts))
     prompts))
 
-(cl-defmethod gptel--wrap-user-prompt ((_backend gptel-anthropic) prompts)
+(cl-defmethod gptel--wrap-last-user-prompt ((_backend gptel-anthropic) prompts)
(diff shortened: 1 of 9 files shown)


tschwinge:
This textually depends on #413 'Consistently use "last user prompt" for where context may be included' (and therefore includes those changes in this PR here).

tschwinge:
I'll gladly take any pointer about how to re-do `gptel--infix-use-context` in a more "visually pleasing" way.

tschwinge:
> This textually depends on #413 'Consistently use "last user prompt" for where context may be included' (and therefore includes those changes in this PR here).

I've removed this with the last force-push.

(Not yet properly tested.)

karthink:
@tschwinge Thanks for the PR.  I've exhausted my available time for gptel fixing some bugs in the last release, so it'll be a few days before I can look at this.

karthink:
gptel now prepends the context to the system message by default, so this PR is not required.  Closing.

Issue #330, opened 2024-06-18
How to pass additional options? Eg., num_ctx for Ollama


krvpal:
Hi, thank you for this excellent package!

I'm using Ollama and to be able to set the context length, ollama [provides this option](https://github.com/ollama/ollama/blob/main/docs/faq.md#how-can-i-specify-the-context-window-size):

```
"options": {
    "num_ctx": 4096
  }
```

I'm struggling to use this in Gptel. I understand there is a `curl-args` keyword, but any variation like the below I'm trying does not seem to be working. I've set `gptel-log-level` to `debug` and checking the logs, but the option does not show up.
```
  (defvar gptel--ollama
    (gptel-make-ollama
        "ollama"                             
      :host "localhost:12345"             
      :protocol "http"
      :models '("llama3")           
      :stream t
      :curl-args '("num_ctx:4096")
      ))                             
```

Please help!

karthink:
gptel uses Ollama's `/api/chat` endpoint, not `/api/generate`.  I'm not sure if `num_ctx` is meaningful for Chat-style requests.

In any case, `num_ctx` is not explicitly supported by gptel.  If you want to limit the context size, you can run the relevant `/set parameter ...` command in Ollama itself, or use the `-n` option in gptel's transient menu to limit the request to the last `n` responses:

![Screenshot_20240618_145812](https://github.com/karthink/gptel/assets/8607532/94cd77ac-cd0c-49a4-91ef-5702bf5caa78)

---
Only a subset of options common to most LLM APIs is exposed by gptel right now.  I plan to eventually cover backend-specific parameters, such as `num_ctx`, but this won't be any time soon.

krvpal:
Thanks for your response. 

I'm looking to increase the default context length. Without passing the `num_ctx` to Ollama, the context length is always set to 2048. This is a problem when I'm using Gptel to summarize text, which is longer and the response I receive only considers the last paragraph. 

I would like to increase context length to 4096 or higher, as there are fine tuned llama3 models that are capable of 8x the usual context length. 

This is a powerful use case to summarize long texts in org, markdown or any text file if one could set `num_ctx`. I'm using the `-n` parameter already as suggested. 

Is using `curl-args` out of the question? 

Thanks!

karthink:
> I'm looking to increase the default context length. Without passing the `num_ctx` to Ollama, the context length is always set to 2048. This is a problem when I'm using Gptel to summarize text, which is longer and the response I receive only considers the last paragraph.
[...]

PR #415, opened 2024-10-12
gptel-ollama: support customizing num_ctx

wangwb98:
Ollama num_ctx can be set in API request when modelfile default value is smaller than the model's max capability.

This patch removed the forced 8192 setting, and added customize-variable "gptel-ollama-num-ctx".

Main reason to name it as ollama specific variable instead of global gptel variable is it's only useful for ollama.
Other backends should keep using gptel-max-tokens which includes token count for both request and response.

Not adding a transient item in gptel-menu, also because it's ollama specific. Users should either customize it, or set it in elisp before calling gptel-request etc functions.

### DIFF
--- a/gptel-ollama.el
+++ b/gptel-ollama.el
@@ -30,6 +30,12 @@
 (declare-function gptel-context--wrap "gptel-context")
 (defvar json-object-type)
 
+(defcustom gptel-ollama-num-ctx nil
+  "Max input context, used to set num_ctx in Ollama."
+  :safe #'always
+  :type '(choice (natnum :tag "Specify count")
+                 (const :tag "Default" nil)))
+
 ;;; Ollama
 (cl-defstruct (gptel-ollama (:constructor gptel--make-ollama)
                             (:copier nil)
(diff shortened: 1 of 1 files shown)


wangwb98:
Hi @karthink ,
this pull request added a customize variable for ollama num_ctx, and removed the 8192 setting which prevents user to use models already has num_ctx > 8192.
Background discussion see #330 
Feel free to edit it or let me know what to change. THanks.

karthink:
@wangwb98 Since this is a backend-specific request parameter, i'd prefer to specify it along with the backend instead of as a top-level defvar.  Do you think adding a `:numctx` keyword option to `gptel-make-ollama` makes sense?  This way you can define a different Ollama backend with a different value of `:numctx` and switch on the fly, or use them simultaneously in different buffers.

Another question I had is whether this is something you want to be able to specify even more granularly, per _model_ instead of per _backend_.  Then you can set a different `num_ctx` for each model used in an Ollama backend.

I'm not using Ollama so I don't know how you use `num_ctx`.

karthink:
@wangwb98 please see [this comment](https://github.com/karthink/gptel/issues/330#issuecomment-2468732605) in #330.  I've added support for setting any Ollama request parameter per-backend or per-model.

See also the discussion in #471: defining variables like `gptel-ollama-num-ctx` does not scale, as there are hundreds of parameters (across all backends) that you might want to set.

If this approach is satisfactory we can close this PR.

wangwb98:
Totally agree with you, thanks!

PR #417, opened 2024-10-12
Update `gptel-model` docstring

benthamite:
I understand that `gptel-model` takes a symbol rather than a string now, so I believe the docstring should be updated to reflect this change.


karthink:
Oops, missed that during the merge.  Thank you.

PR #435, opened 2024-10-22
gptel: Make `gptel-expert-commands` a user option

benthamite:
I noticed that `gptel-expert-commands` was not a user option after I defined it via `:custom` in use-package and it failed to be set. As far as I can tell, there is no reason for it to be a defvar, but if this was intentional feel free to reject this PR.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -588,10 +588,12 @@ README for examples."
           (restricted-sexp :match-alternatives (gptel-backend-p 'nil)
 			   :tag "Other backend")))
 
-(defvar gptel-expert-commands nil
+(defcustom gptel-expert-commands nil
   "Whether experimental gptel options should be enabled.
 
-This opens up advanced options in `gptel-menu'.")
+This opens up advanced options in `gptel-menu'."
+  :safe #'always
+  :type 'boolean)
 
 (defvar-local gptel--bounds nil)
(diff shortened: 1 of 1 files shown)


karthink:
There's a little story here.

The right way to handle item visibility in transient menus is to assign them to different transient levels, which you can choose or set persistently by pressing `C-x l` in a transient menu.  (To show all levels temporarily you can use `C-x a`.)  This is how Magit handles "hidden" commands.

I tried that, but users were quite confused by the instructions.  Saying "Run `(setq gptel-expert-commands t)`" is simpler to understand, since everyone knows what `setq` does and the option name speaks for itself.  Saying "In the transient menu, press `C-x l` and assign the dry-run commands to level 4 or under" is confusing when you're already trying to debug a problem with gptel.  So I reverted this change.

But the `gptel-expert-commands` method isn't great either.  I just left it as a defvar for later, hoping to figure out a better system for showing advanced commands in the menu.

So here we are.  If we make it a defcustom that pretty much sets it in stone.  Switching to another method later will annoy users who have this set in their configurations.  As a defvar the setting is somewhat "internal".  I'm not sure what to do.

benthamite:
Ah, I see! No worries, I don’t think this is very important. Perhaps it could be added as an infix to the gptel menu that toggles the display of hidden commands on and off?

Feel free to close this.

PR #436, opened 2024-10-23
gptel-anthropic: Add upgraded Claude 3.5 Sonnet model

benthamite:
- Add `claude-3-5-sonnet-20241022`, the upgraded Claude 3.5 Sonnet model
released on 2024-10-22. “The upgraded Claude 3.5 Sonnet delivers
across-the-board improvements over its predecessor, with particularly
significant gains in coding—an area where it already led the field.”
(https://www.anthropic.com/news/3-5-models-and-computer-use)
- Replace a source with a more comprehensive one.

### DIFF
--- a/gptel-anthropic.el
+++ b/gptel-anthropic.el
@@ -184,7 +184,15 @@ files in the context."
 ;;         (plist-get (car (last prompts)) :content)))
 
 (defconst gptel--anthropic-models
-  '((claude-3-5-sonnet-20240620
+  '((claude-3-5-sonnet-20241022
+     :description "Highest level of intelligence and capability"
+     :capabilities (media tool)
+     :mime-types ("image/jpeg" "image/png" "image/gif" "image/webp")
+     :context-window 200
+     :input-cost 3
+     :output-cost 15
+     :cutoff-date "2024-04")
+    (claude-3-5-sonnet-20240620
      :description "Highest level of intelligence and capability"
      :capabilities (media tool)
      :mime-types ("image/jpeg" "image/png" "image/gif" "image/webp")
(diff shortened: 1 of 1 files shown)


karthink:
Cheers @benthamite!

algal:
Yay! I came here wondering and like magic it appeared within minutes of me looking.

benthamite:
@karthink, upon reflection, I wonder if we should simply replace `claude-3-5-sonnet-20240620` with `claude-3-5-sonnet-20241022` rather than adding the latter as a new model.

unuseless:
Thanks for the fast update.
I prefer that a user can choose which model to use, so I think it's better to have a choice for each model, and not replace claude-3-5-sonnet-20240620 with claude-3-5-sonnet-2024102.
Is it also possible to specify your own model, like adding a model spec like:
```
 (claude-3-5-sonnet-20240620           ;Full spec
   :description  \"Balance of intelligence and speed\"
   :capabilities (media tool json)
   :mime-types
   (\"image/jpeg\" \"image/png\" \"image/gif\" \"image/webp\")))

```
into your .emacs?

karthink:
@benthamite I think it's better to have both instead of replacing it, there are reasons to want to continue to use the old model.  We might actually also want to add `claude-3-5-sonnet-latest` and mention in the metadata that it's the same as `claude-3-5-sonnet-20241022` right now.  This way someone who always wants the latest model can set it to `-latest`.

karthink:
@unuseless 

> Is it also possible to specify your own model, like adding a model spec like

```elisp
(setq gptel-backend (gptel-make-claude :key ...))
(push
 '(claude-3-5-sonnet-20240620
   :description  \"Balance of intelligence and speed\"
   :capabilities (media tool json)
   :mime-types
   (\"image/jpeg\" \"image/png\" \"image/gif\" \"image/webp\"))
```

PR #438, opened 2024-10-26
Support adding files in directories to context

benthamite:
Currently, attempting to add directories to the context in Dired will throw an error. This PR provides support for adding files in directories to the context: if there is a directory at point, or the marked files include one or more directories, their files will recursively be added to the context.

### DIFF
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -111,10 +111,19 @@ context chunk.  This is accessible as, for example:
     (message "Current region added as context."))
    ;; If in dired
    ((derived-mode-p 'dired-mode)
-    (mapc (if (and arg (< (prefix-numeric-value arg) 0))
-              #'gptel-context-remove
-              #'gptel-context-add-file)
-          (dired-get-marked-files)))
+    (let* ((files (dired-get-marked-files))
+           (dirs (cl-remove-if-not #'file-directory-p files))
+           (remove-p (< (prefix-numeric-value arg) 0))
+	   (action-fn (if remove-p
+			  #'gptel-context-remove
+			#'gptel-context-add-file)))
+      (when (or remove-p
+		(null dirs)
+		(not (called-interactively-p 'any))
+		(y-or-n-p (format "Recursively add files from %d director%s? "
+				  (length dirs)
+				  (if (= (length dirs) 1) "y" "ies"))))
+	(mapc action-fn files))))
    ;; If in an image buffer
    ((and (derived-mode-p 'image-mode)
          (gptel--model-capable-p 'media;)
@@ -180,23 +189,19 @@ Return PATH if added, nil if ignored."
 (defun gptel-context--handle-directory (path action)
   "Process all files in directory at PATH according to ACTION.
 ACTION should be either `add' or `remove'."
-  (cl-block gptel-context--handle-directory
-    (when (eq action 'add)
-      (unless (y-or-n-p (format "Recursively add all files from directory %s? " path))
-        (cl-return-from gptel-context--handle-directory)))
-    (let ((files (directory-files-recursively path "." t)))
-      (mapc (lambda (file)
-              (unless (file-directory-p file)
-                (pcase-exhaustive action
-                  ('add
-                   (if (gptel--file-binary-p file)
-                       (gptel-context--handle-binary file)
-                     (cl-pushnew (list file) gptel-context--alist :test #'equal)
-                     (message "File \"%s\" added to context." file)))
-                  ('remove
-                   (setf (alist-get file gptel-context--alist nil 'remove #'equal) nil)
-                   (message "File \"%s\" removed from context." file)))))
-            files))))
+  (let ((files (directory-files-recursively path "." t)))
+    (mapc (lambda (file)
+            (unless (file-directory-p file)
+              (pcase-exhaustive action
+                ('add
+                 (if (gptel--file-binary-p file)
+                     (gptel-context--handle-binary file)
+                   (cl-pushnew (list file) gptel-context--alist :test #'equal)
+                   (message "File \"%s\" added to context." file)))
+                ('remove
+                 (setf (alist-get file gptel-context--alist nil 'remove #'equal) nil)
+                 (message "File \"%s\" removed from context." file)))))
+          files)))
 
 (defun gptel-context-add-file (path)
   "Add the file at PATH to the gptel context.
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -204,8 +204,7 @@ ACTION should be either `add' or `remove'."
                      (gptel-context--handle-binary file)
                    (gptel-context--add-text-file file)))
                 ('remove
-                 (setf (alist-get file gptel-context--alist nil 'remove #'equal) nil)
-                 (message "File \"%s\" removed from context." file)))))
+                 (gptel-context--remove-file file)))))
           files)))
 
 (defun gptel-context-add-file (path)
@@ -239,8 +238,7 @@ If CONTEXT is a directory, recursively removes all files in it."
    ((stringp context)                   ;file or directory
     (if (file-directory-p context)
         (gptel-context--handle-directory context 'remove)
-      (setf (alist-get context gptel-context--alist nil 'remove #'equal)
-            nil)))
+      (gptel-context--remove-file context)))
    ((region-active-p)
     (when-let ((contexts (gptel-context--in-region (current-buffer)
                                                    (region-beginning)
@@ -249,6 +247,12 @@ If CONTEXT is a directory, recursively removes all files in it."
    (t
     (when-let ((ctx (gptel-context--at-point)))
       (delete-overlay ctx)))))
+
+(defun gptel-context--remove-file (path)
+  "Remove the file at PATH from the gptel context."
+  (setf (alist-get path gptel-context--alist nil 'remove #'equal)
+	nil)
+  (message "File \"%s\" removed from context." path))
 
 (defun gptel-context--make-overlay (start end &optional advance)
   "Highlight the region from START to END.
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -91,8 +91,10 @@ context chunk.  This is accessible as, for example:
   context.  If there is already a gptel context at point, remove it
   instead.
 
-- If in Dired, add marked files or file at point to the context.
-  With negative prefix ARG, remove them from the context instead.
+- If in Dired, add marked files or file at point to the context. If
+  the file at point is a directory, or directories are marked,
+  recursively add all their files. With negative prefix ARG, remove
+  them from the context instead.
 
 - Otherwise add the current buffer to the context.  With positive
   prefix ARG, prompt for a buffer name and add it to the context.
@@ -110,10 +112,14 @@ context chunk.  This is accessible as, for example:
     (message "Current region added as context."))
    ;; If in dired
    ((derived-mode-p 'dired-mode)
-    (mapc (if (and arg (< (prefix-numeric-value arg) 0))
-              #'gptel-context-remove
+    (let ((files (cl-loop for file in (dired-get-marked-files)
+			  if (file-directory-p file)
+			  append (directory-files-recursively file "." t)
+			  else collect file)))
+      (mapc (if (and arg (< (prefix-numeric-value arg) 0))
+		#'gptel-context-remove
               #'gptel-context-add-file)
-          (dired-get-marked-files)))
+            (cl-remove-if #'file-directory-p files))))
    ;; If in an image buffer
    ((and (derived-mode-p 'image-mode)
          (gptel--model-capable-p 'media;)
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -84,23 +84,26 @@ context chunk.  This is accessible as, for example:
   :group 'gptel
   :type 'function)
 
-(defun gptel-context-add (&optional arg)
+(defun gptel-context-add (&optional arg confirm)
   "Add context to gptel in a DWIM fashion.
 
 - If a region is selected, add the selected region to the
   context.  If there is already a gptel context at point, remove it
   instead.
 
-- If in Dired, add marked files or file at point to the context.
-  Directories will have all their files added recursively. With
-  negative prefix ARG, remove them from the context instead.
+- If in Dired, add marked files or file at point to the context. With
+  negative prefix ARG, remove them from the context instead. If the
+  selection includes directories, add all their files recursively,
+  prompting the user for confirmation if called interactively or
+  CONFIRM is non-nil.
 
 - Otherwise add the current buffer to the context.  With positive
   prefix ARG, prompt for a buffer name and add it to the context.
 
-- With negative prefix ARG, remove all gptel contexts from the
-  current buffer."
-  (interactive "P")
+- With negative prefix ARG, remove all gptel contexts from the current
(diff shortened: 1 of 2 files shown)


karthink:
Looks good! I'll test it when I can and merge.

benthamite:
I just noticed this in the docstring of this function (emphasis added):

> If a region is selected, add the selected region to the context.  **If there is already a gptel context at point, remove it instead.**

I may not be understanding what this means, but as far as I can tell the function does not behave in the manner described: if I select a region, add it to the context, then re-select it and call `gptel-add` again, the region is not removed from the context. (The docstring talks about the "context at point" rather than the region, but I am not sure what that means in this, er, context. 😛)

karthink:
> I may not be understanding what this means, but as far as I can tell the function does not behave in the manner described: if I select a region, add it to the context, then re-select it and call `gptel-add` again, the region is not removed from the context.

It could probably be worded better.  Here is an updated description, let me know if it makes more sense:

```
gptel-add is an alias for ‘gptel-context-add’ in ‘gptel-context.el’.

(gptel-add &optional ARG)

Add context to gptel in a DWIM fashion.

- If a region is selected, add the selected region to the
  context.  (If point is in a context region already, it
  is removed first -- overlapping context regions are not
  supported.)

- If there is already a gptel context at point, remove it
  instead.

- If in Dired, add marked files or file at point to the context.
  With negative prefix ARG, remove them from the context instead.

- Otherwise add the current buffer to the context.  With positive
  prefix ARG, prompt for a buffer name and add it to the context.

- With negative prefix ARG, remove all gptel contexts from the
  current buffer.

```

karthink:
We should also add the ability to add directories to the context when running `gptel-add-file`, shouldn't we?

benthamite:
> We should also add the ability to add directories to the context when running
`gptel-add-file`, shouldn't we?

I wasn’t familiar with this command, but yeah, it seems it should mirror the behavior of `gptel-add` when the latter is used with a file or directory.

benthamite:
Perhaps we should change `gptel-context-add-file` so that this is the function called by `gptel-add` in Dired mode. Currently, the former doesn’t take a prefix argument.

karthink:
> Would you like me to make the changes?

Please go ahead.

benthamite:
@karthink, I undid the previous attempt and pushed a new version, which (1) works with both `gptel-add-file` and `gptel-add` and (2) provides the functionality to both add and remove files in directories. Let me know what you think.

benthamite:
Mmh, there is some issue with removing files in dirs. Let me look into it before you review the PR.

gptel-context.el
```diff
@@ -163,32 +165,57 @@ context chunk.  This is accessible as, for example:
...
+
+(defun gptel-context--handle-directory (path action)
+  "Process all files in directory at PATH according to ACTION.
+ACTION should be either `add' or `remove'."
+  (let ((files (directory-files-recursively path "." t)))
+    (mapc (lambda (file)
+            (unless (file-directory-p file)
+              (pcase action
+		('add (gptel--file-binary-p file)
+		      (gptel-context--handle-binary file)
```
karthink, line 189: Am I missing something here?  Seems like this should be `(if (gptel--file-binary-p file) (gptel-context--handle-binary file) ...`.

gptel-context.el
```diff
@@ -163,32 +165,57 @@ context chunk.  This is accessible as, for example:
...
+    (message "Ignoring unsupported binary file \"%s\"." path)
+    nil))
+
+(defun gptel-context--handle-directory (path action)
+  "Process all files in directory at PATH according to ACTION.
+ACTION should be either `add' or `remove'."
+  (let ((files (directory-files-recursively path "." t)))
+    (mapc (lambda (file)
+            (unless (file-directory-p file)
+              (pcase action
```
karthink, line 187: Use `pcase-exhaustive` to future-proof uses.

gptel-context.el
```diff
@@ -163,32 +165,57 @@ context chunk.  This is accessible as, for example:
...
-          (prog1 path
-            (cl-pushnew (list path :mime mime)
-                        gptel-context--alist :test #'equal)
-            (message "File \"%s\" added to context." path))
-        (message "Ignoring unsupported binary file \"%s\"." path))
-    ;; Add text file
-    (cl-pushnew (list path) gptel-context--alist :test #'equal)
-    (message "File \"%s\" added to context." path)
-    path))
+  (if (file-directory-p path)
```
karthink, line 202: Convert to a `cond` clause, it's easier to read here.

gptel-context.el
```diff
@@ -91,8 +91,10 @@ context chunk.  This is accessible as, for example:
   context.  If there is already a gptel context at point, remove it
   instead.
 
-- If in Dired, add marked files or file at point to the context.
-  With negative prefix ARG, remove them from the context instead.
+- If in Dired, add marked files or file at point to the context. If
```
karthink, line 94: Do you think we should ask for confirmation (`y-or-n-p`) when adding a directory with `gptel-add`?

gptel-context.el
```diff
@@ -91,8 +91,10 @@ context chunk.  This is accessible as, for example:
...
+  the file at point is a directory, or directories are marked,
```
karthink, line 95: This is difficult to read.  How about

```
- If in Dired, add marked files or file at point to the context.  Directories
  will have all their files added recursively.  With negative prefix ARG, remove
  them from the context instead.
```

karthink (changes requested):

gptel-context.el
benthamite: Yep, not sure what happened. Will make these and the other changes shortly.

gptel-context.el
benthamite: I weakly lean towards asking for confirmation. The prompt could include the number of files to be added.

gptel-context.el
benthamite: Looks good.

benthamite: Addressed: [3e5c1e6](https://github.com/karthink/gptel/pull/438/commits/8e5e3c867131f0c795c7eb62dd8ba28b25a9cf27)

gptel-context.el
benthamite: Addressed: https://github.com/karthink/gptel/pull/438/commits/2a40daef00c91620f142c0df8494883cdf3c66fe

gptel-context.el
benthamite: Addressed: https://github.com/karthink/gptel/pull/438/commits/6c6698210270a861001667ad5e798752c1194b66

gptel-context.el
benthamite: Addressed: https://github.com/karthink/gptel/pull/438/commits/bc9138a525120e8786d7390a436e44895e660c09

gptel-context.el
benthamite: Here’s an attempt: https://github.com/karthink/gptel/pull/438/commits/8e5e3c867131f0c795c7eb62dd8ba28b25a9cf27

Note that the user will be asked once for each directory. This seems like the correct behavior, though it can become tedious if the user is adding several dirs.

karthink:
Will take a look soon, thanks for the update!

gptel-context.el
```diff
@@ -163,32 +164,62 @@ context chunk.  This is accessible as, for example:
...
+        (message "File \"%s\" added to context." path))
+    (message "Ignoring unsupported binary file \"%s\"." path)
+    nil))
+
+(defun gptel-context--handle-directory (path action)
+  "Process all files in directory at PATH according to ACTION.
+ACTION should be either `add' or `remove'."
+  (cl-block gptel-context--handle-directory
+    (when (eq action 'add)
+      (unless (y-or-n-p (format "Recursively add all files from directory %s? " path))
```
karthink, line 185: `gptel-context--handle-directory` is a non-interactive function.  Running an interactive `y-or-n-p` check inside a non-interactive function is not a good idea, since `gptel-add` can be used non-interactively for scripting.  [Here's an example](https://github.com/meain/dotfiles/blob/master/emacs/.config/emacs/init.el#L4525).  So this check should happen in `gptel-context-add-file`.

gptel-context.el
```diff
@@ -163,32 +164,62 @@ context chunk.  This is accessible as, for example:
...
+                   (if (gptel--file-binary-p file)
+                       (gptel-context--handle-binary file)
+                     (cl-pushnew (list file) gptel-context--alist :test #'equal)
+                     (message "File \"%s\" added to context." file)))
+                  ('remove
+                   (setf (alist-get file gptel-context--alist nil 'remove #'equal) nil)
+                   (message "File \"%s\" removed from context." file)))))
+            files))))
+
 (defun gptel-context-add-file (path)
```
karthink, line 201: We can add an `&optional confirm-dir` argument to this function that the `(interactive ...)` form can set to `t`.

gptel-context.el
```diff
@@ -163,32 +164,62 @@ context chunk.  This is accessible as, for example:
...
-            (cl-pushnew (list path :mime mime)
-                        gptel-context--alist :test #'equal)
-            (message "File \"%s\" added to context." path))
-        (message "Ignoring unsupported binary file \"%s\"." path))
-    ;; Add text file
-    (cl-pushnew (list path) gptel-context--alist :test #'equal)
-    (message "File \"%s\" added to context." path)
-    path))
+  (cond ((file-directory-p path)
+	 (gptel-context--handle-directory path 'add))
```
karthink, line 207: Unless `conirm-dir` is `nil`, we can run the `y-or-n-p` check here before calling `gptel-context--handle-directory`.

gptel-context.el
benthamite: > The other advantage is that we only need to ask once per call to `gptel-add`, which is less annoying than asking per-directory.

Apologies if I’m misunderstanding something, but since `gptel-add` calls `gptel-context-add-file` once per directory, moving the logic to the latter would not by default have the effect you describe, I think. Can you clarify?

karthink: @benthamite Right on cue! https://github.com/karthink/gptel/issues/513#issuecomment-2560155884

benthamite: > You are not adding all files inside a directory recursively -- instead you're only descending one level

I don’t think this is correct: the current functionality adds all files inside a directory, regardless of whether these files are just one level up or in other subdirectories. If you think I’m mistaken, could you provide an example where the function doesn’t do the above?

benthamite: Separately, I have moved the `y-or-n-p` prompt from `gptel-context-add` to `gptel-context-add`. This also prompts once per call rather than once per directory.

karthink: > I don’t think this is correct: the current functionality adds all files inside a directory

Sorry, you're right -- I confused myself twice there.

gptel-context.el
```diff
@@ -110,10 +111,19 @@ context chunk.  This is accessible as, for example:
...
-          (dired-get-marked-files)))
+    (let* ((files (dired-get-marked-files))
+           (dirs (cl-remove-if-not #'file-directory-p files))
+           (remove-p (< (prefix-numeric-value arg) 0))
+	   (action-fn (if remove-p
+			  #'gptel-context-remove
+			#'gptel-context-add-file)))
+      (when (or remove-p
+		(null dirs)
+		(not (called-interactively-p 'any))
```
karthink, line 122: I advise against using `called-interactively-p`.  As the documentation explains:

```
This function is very brittle, it may fail to return the intended result when
the code is debugged, advised, or instrumented in some form.  Some macros and
special forms (such as ‘condition-case’) may also sometimes wrap their bodies
in a ‘lambda’, so any call to ‘called-interactively-p’ from those bodies will
indicate whether that lambda (rather than the surrounding function) was called
interactively.

Instead of using this function, it is cleaner and more reliable to give your
function an extra optional argument whose ‘interactive’ spec specifies
non-nil unconditionally ("p" is a good way to do this), or via
(not (or executing-kbd-macro noninteractive)).
```

gptel-context.el
```diff
@@ -163,32 +173,59 @@ context chunk.  This is accessible as, for example:
         (eq buffer-file-coding-system 'no-conversion))
     (file-missing (message "File \"%s\" is not readable." path))))
 
+(defun gptel-context--add-text-file (path)
```
karthink, line 176: We now have `--add-text-file`, `--handle-binary` and `--handle-directory`.  Can we unify the naming here to `--handle-text-file`, `--handle-binary-file` and `--handle-directory`?

I actually prefer `--add-text-file`, `--add-binary-file` and `--add-directory`.  It's okay that `--add-directory` both adds and removes directories, because `gptel-add` both adds and removes context.

karthink:
As expected, adding support for directories has been followed up by requests to ignore the `.git` directory and to support `.gitignore` specifications.  What do you think about adding these?

gptel-context.el
karthink: Leaving this comment open since I explain why we want to avoid `called-interactively-p` below.

gptel-context.el
benthamite: Ah, I didn’t know this. Thanks!

benthamite:
> As expected, adding support for directories has been followed up by requests to ignore the `.git` directory and to support `.gitignore` specifications. What do you think about adding these?

Yes, happy to add these. However, given what you say in [this comment](https://github.com/karthink/gptel/issues/513#issuecomment-2561369928), perhaps we should exclude this enhancement from the current PR?

karthink:
> perhaps we should exclude this enhancement from the current PR?

Yes, it's out of scope for this PR.  We can do it separately.

benthamite:
@karthink: I replaced the `called-interactively-p` call with an extra optional argument, as you proposed. I noticed that `gptel-context-add` already had a `y-or-n-p` confirmation prompt in the `(and arg (< (prefix-numeric-value arg) 0))` branch, so I thought we could use that extra argument to handle this case, too.

gptel-context.el
```diff
@@ -210,6 +252,12 @@ If selection is active, removes all contexts within selection."
     (when-let ((ctx (gptel-context--at-point)))
       (delete-overlay ctx)))))
 
+(defun gptel-context--remove-file (path)
+  "Remove the file at PATH from the gptel context."
+  (setf (alist-get path gptel-context--alist nil 'remove #'equal)
+	nil)
+  (message "File \"%s\" removed from context." path))
```
karthink, line 259: This is fine when removing individual files.  When removing a directory however, this will emit one message per removed file, which can be a lot.  I think this messaging needs to happen elsewhere.

karthink:
> @karthink: I replaced the `called-interactively-p` call with an extra optional argument, as you proposed. I noticed that `gptel-context-add` already had a `y-or-n-p` confirmation prompt in the `(and arg (< (prefix-numeric-value arg) 0))` branch, so I thought we could use that extra argument to handle this case, too.

Yup, this looks good!

gptel-context.el
benthamite: Addressed here: https://github.com/karthink/gptel/pull/438/commits/f73584eebfd6f483cd98e02e89f08b0e01f336b4

karthink: Perfect, I think we can merge this now, unless you plan to make any more changes?

benthamite: Great. Yes, let’s merge it.

karthink:
Thanks for the PR @benthamite.  For a relatively small addition there were a surprising number of things to sort out!

Issue #455, opened 2024-11-02
[Windows] Unsupported map type ‘symbol’: json-read-error (due to CR-LF issues)


ssafar:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [X] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

When parsing JSON responses, we get something like this:

```
Debugger entered--Lisp error: (error "Unsupported map type ‘symbol’: json-read-error")
  error("Unsupported map type `%S': %S" symbol json-read-error)
  map-elt(json-read-error :prompt_eval_count)
  (or (map-elt response :prompt_eval_count) 0)
  (+ (or (map-elt response :prompt_eval_count) 0) (or (map-elt response :eval_count) 0))
 (... left out some frames for conciseness)
  gptel-curl--parse-response((:token "bccdcda94429f3c6caee4658f8380f0c" :backend #s(gptel-ollama :name "Ollama" (... etc)
```

Upon looking at the `*gptel-curl*` buffer, we see a reasonable response; the cursor location in it is somewhere towards the beginning of the JSON response (but not quite at its start).

(PR coming soon)

### Backend

None

### Steps to Reproduce

1. open a chat window
2. try making any query

### Additional Context

Emacs 28.1, Windows 10,

```
curl 7.79.1 (Windows) libcurl/7.79.1 Schannel
Release-Date: 2021-09-22
Protocols: dict file ftp ftps http https imap imaps pop3 pop3s smtp smtps telnet tftp
Features: AsynchDNS HSTS IPv6 Kerberos Largefile NTLM SPNEGO SSL SSPI UnixSockets
```

karthink:
This makes sense, thanks for reporting it.  But now I'm confused about why no one has had this problem with gptel on Windows before!

ssafar:
I'm guessing "streaming" vs. "not streaming" might make a difference & many people might just be using APIs with streaming on by default? (e.g. Anthropic APIs worked for me before, this only surfaced when trying to use an Ollama instance)

Thanks for merging btw!

PR #456, opened 2024-11-02
gptel-curl: don't try converting CR-LF on Windows

ssafar:
We use curl's header size output to `goto-char` to the end of the response header (to parse the JSON that starts there).

On Windows, however, processes we launch will typically output Windows-style line endings, with CR-LF; by default, Emacs tries to convert this to just CR, to avoid seeing ugly `^M` characters at the end of lines.

 `curl` dumps what it gets from the server directly... but HTTP also specifies
 CR-LF line endings, at least at the end of headers. On Windows, Emacs converts
 these to CRs, removing the LFs, and thus causing our header size measurement to
 be slightly off (the buffer now includes fewer characters of header vs. what
 `curl` told us we'd have). So `goto-char` seeks too much ahead, missing the
 beginning of the JSON; we get a JSON parse error.

This PR fixes this by forcing the process coding system to the UNIX variant (with no CR-LF conversion).

We gate this behind an "is this Windows" check so it shouldn't affect other systems.

Fixes https://github.com/karthink/gptel/issues/455.

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -113,6 +113,10 @@ the response is inserted into the current buffer after point."
          (backend (buffer-local-value 'gptel-backend (plist-get info :buffer)))
          (process (apply #'start-process "gptel-curl"
                          (generate-new-buffer "*gptel-curl*") "curl" args)))
+    (when (memq system-type '(windows-nt ms-dos))
+      ;; Don't try to convert cr-lf to cr on Windows so that curl's "header size
+      ;; in bytes" stays correct
+      (set-process-coding-system process 'utf-8-unix 'utf-8-unix))
     (when (eq gptel-log-level 'debug)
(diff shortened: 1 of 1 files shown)


karthink:
Thanks.  Do you think we need the `memq` check?  It seems like we can set this unconditionally.

ssafar:
Just checked on linux, it's `(utf-8-unix utf-8-unix)` by default already, so it should be fine... but then the two windows systems had `(utf-8-dos utf-8-unix)` and `(iso-latin-2-dos iso-latin-2-unix)` respectively, so maybe there are other unix ones that have non-obvious defaults? Thus being conservative with this (... we're surely not breaking anyone else this way & we're making windows strictly better).

karthink:
Very well, let's see if this fixes some other bugs (i.e. not `json-read-error`) users are facing with gptel+Curl on Windows.

PR #461, opened 2024-11-05
Correct call to provided-mode-derived-p in gptel--strip-mode-suffix

nordlow:


karthink:
@nordlow Your change seems incorrect, `provided-mode-derived-p` takes a `&rest` argument.

```elisp
(provided-mode-derived-p MODE &rest MODES)
```

nordlow:
Ahh, the signature has changed on master. I prefer living on the edge.

karthink:
Per Nordlöw ***@***.***> writes:

> Ahh, the signature has changed on master. I prefer living on the edge. 

They broke all existing uses of the function?  They don't usually do that.

karthink:
It looks like the old calling convention will continue to work but is deprecated.  Since we support Emacs back up to 27.2, the current version cannot be changed, unless `compat-30` adds the new version.  I'll add a comment about updating this in the future.

PR #467, opened 2024-11-06
Added a section tell new users how to select a backend explicitly

blais:
It took me too long to search for this. It deserves an explicit callout like this IMHO.

Great mode, BTW!  This is amazing!


karthink:
Thank you @blais!

PR #470, opened 2024-11-09
fix(private-gpt): convert model name from symbol to string in request

rosenstrauch:
the switch to symbol for models means we need to convert them back to strings in json requests.


karthink:
@rosenstrauch Thanks for the fix!

PR #472, opened 2024-11-12
bug fix in Ollama backend: options-plist -> prompts-plit

andrewdea:
hey thanks for working on this package, it's amazing 💜

latest commit introduced a small bug in `gptel-ollama.el`: code was trying to modify `options-plist` without having setting it, resulting in "setq: Symbol’s value as variable is void: options-plist". 

Here's my fix: if I'm understanding things right, `options-plist` is not used anymore, so we just put the values directly into `prompts-plist`

### DIFF
--- a/gptel-ollama.el
+++ b/gptel-ollama.el
@@ -82,13 +82,11 @@ Intended for internal use only.")
                          (gptel-backend-stream gptel-backend))
                      :json-false))))
     (when gptel-temperature
-      (setq options-plist
-            (plist-put options-plist :temperature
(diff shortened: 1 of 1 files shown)


karthink:
Thank you for catching that! That's what I get for turning off the linter.  Unfortunately your fix won't work since `temperature`, `max_tokens` etc need to be inside an `:options` field.  I just "fixed" it -- I was in the middle of working on the code already -- but by punting a potential issue into the future (the commit message explains).

PR #483, opened 2024-11-24
Update `gptel--anthropic-models`

benthamite:
* Replace sources with link to a comprehensive comparison table.

* Reorder models to reflect the order in that table.

* Slightly tweak description of `claude-3-5-sonnet-20240620` to make it clear it
is an earlier version.


karthink:
Thanks for the update @benthamite!

Issue #473, opened 2024-11-12
(Ollama) temperature/max_token settings lost when using model-specific request parameters


andrewdea:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [X] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

setq: Symbol’s value as variable is void: options-plist

[see my fix here](https://github.com/karthink/gptel/pull/472)

### Backend

Ollama

### Steps to Reproduce

`M-x` gptel-send

### Additional Context

Emacs version: 29.4

### Backtrace

```emacs-lisp
Debugger entered--Lisp error: (void-variable options-plist)
  (plist-put options-plist :temperature gptel-temperature)
  (setq options-plist (plist-put options-plist :temperature gptel-temperature))
  (progn (setq options-plist (plist-put options-plist :temperature gptel-temperature)))
  (if gptel-temperature (progn (setq options-plist (plist-put options-plist :temperature gptel-temperature))))
  (let ((prompts-plist (list ':model (gptel--model-name gptel-model) ':messages (vconcat prompts) ':stream (or (and gptel-stream gptel-use-curl (progn (or ... ...) (aref gptel-backend 5))) :json-false)))) (if gptel-temperature (progn (setq options-plist (plist-put options-plist :temperature gptel-temperature)))) (if gptel-max-tokens (progn (setq options-plist (plist-put options-plist :num_predict gptel-max-tokens)))) (gptel--merge-plists prompts-plist (progn (or (progn (and (memq (type-of gptel-backend) cl-struct-gptel-backend-tags) t)) (signal 'wrong-type-argument (list 'gptel-backend gptel-backend))) (aref gptel-backend 10)) (gptel--model-request-params gptel-model)))
  (progn (let ((prompts-plist (list ':model (gptel--model-name gptel-model) ':messages (vconcat prompts) ':stream (or (and gptel-stream gptel-use-curl (progn ... ...)) :json-false)))) (if gptel-temperature (progn (setq options-plist (plist-put options-plist :temperature gptel-temperature)))) (if gptel-max-tokens (progn (setq options-plist (plist-put options-plist :num_predict gptel-max-tokens)))) (gptel--merge-plists prompts-plist (progn (or (progn (and (memq ... cl-struct-gptel-backend-tags) t)) (signal 'wrong-type-argument (list 'gptel-backend gptel-backend))) (aref gptel-backend 10)) (gptel--model-request-params gptel-model))))
[...]

PR #490, opened 2024-11-29
Possible bug fix 473

andrewdea:
See #473.
This simplifies `gptel--request-data`, ensuring that model or backend-specific parameters are preserved without overriding the global options


karthink:
Just a note: haven't lost track of this, will circle back after addressing some other issues.

karthink:
@andrewdea Thanks for the PR!

PR #498, opened 2024-12-07
Fix bug where dry run updates status message

daedsidog:
Before this, inspecting your query causes the status message to get stuck on "Waiting..."

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -900,7 +900,7 @@ Or in an extended conversation:
                           (get-char-property (point) 'read-only))
                 (insert reduced-prompt))
               (setq position (point))
-              (when gptel-mode
+              (when (and gptel-mode (not dry-run))
                 (gptel--update-status " Waiting..." 'warning)))))
          ;; Insert into new gptel session
          (t (setq buffer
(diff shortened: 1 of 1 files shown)


gptel-transient.el
```diff
@@ -900,7 +900,8 @@ Or in an extended conversation:
                           (get-char-property (point) 'read-only))
                 (insert reduced-prompt))
               (setq position (point))
-              (when gptel-mode
+              (when (and gptel-mode
```
karthink, line 903: Formatting issue -- please move this into one line: `when (and gptel-mode (not dry-run))`

karthink:
Thanks @daedsidog, I noticed this too but lost track of it.  Can merge with a tiny formatting fix.

gptel-transient.el
daedsidog: Done. Why is this a formatting issue, though? Just your own personal preference, or is this written down somewhere?

karthink: Personal preference, but I don't know how to enforce it on a project-wide level -- I would like to get github CI to do this automatically on PRs.  But setting up an auto-formatter for elisp with the required rules is a nontrivial task.

The rule is to have as many lines as possible on the screen, where each line is <100 chars long.  I don't think there's a point to writing this down somewhere in the repo without also providing a way to do it automatically.

karthink: Ditto for the commit message template, actually.  I'm following the same commit message format and guidelines as the Org mode project, and it would be good to get Github to generate this automatically.

Actually it might be possible to get gptel to generate the commit messages in the required format, and anyone submitting a PR is already running gptel by necessity. :thinking:

karthink:
Cheers @daedsidog

gptel-transient.el
daedsidog: I would just suggesting dropping a CONTRIBUTING file in the repo with the guidelines, and setting up a template for PRs asking if the submitter followed them. Not really anything enforceable, but better than nothing and takes almost no effort.

PR #499, opened 2024-12-08
README: add prefix string docs for gptel-org-branching-context

erganemic:
Discussion #457


karthink:
Perfect, thank you @erganemic!

PR #512, opened 2024-12-20
gptel-rewrite: Make text replacement preserve overlays

daedsidog:
I am working on gptel integration into my extension.

gptel-rewrite is very nice and already does a lot of work such as adding the ability to diff/ediff changes. Though most of the functions are internal to gptel, I am planning, for now, to use `gptel--rewrite-dispatch` to create the overlays, until a better solution.

I've played around with it and it integrates very nicely, except for one issue: since I am using pre-existing overlays (not in any way related to the overlays gptel-rewrite creates), they are deleted because they have `evaporate` set to true when gptel-rewrite replaces the accepted text.

To circumvent this, I made the text replacement unobtrusive.

This PR is also a question of whether or not this is a good idea to begin with. I am unaware of any compatibility problems that this may create. I can get it to work without having an unobtrusive replacement--this is just the easier path.


karthink:
> I've played around with it and it integrates very nicely, except for one issue:
> since I am using pre-existing overlays (not in any way related to the overlays
> gptel-rewrite creates), they are deleted because they have `evaporate` set to
> true when gptel-rewrite replaces the accepted text.

This is by design -- the rewrite "transaction" ends when you accept or reject
the text.  Otherwise there would be no way to actually get rid of the overlay.

> Besides the changes of convenience for my extension, it seems, to me, to also be
> desirable behavior, but I'll leave you to judge. My only concern is that it goes
> against some sort of standard convention that replacement always means total
> deletion and insertion.

Could you explain why keeping the overlay around after accepting the response is
desirable behavior?  I've actually been considering deleting the overlay after
running any operation, including diff/ediff.

daedsidog:
@karthink There is a misunderstanding, I am not talking about the gptel-rewrite overlay, which should rightly be removed.

I am talking about existing, unrelated overlays in the buffer, surrounding text. I want to replace the text of those overlays, and those overlays are evaporative.

karthink:
Just a note to say that I haven't lost track of this, I'll get to it soon.  I want to test it carefully as there are always unexpected edge cases with overlays.

PR #519, opened 2024-12-24
Add windows platform check for the curl file size thresh

### LINKED ISSUE #517: Different default for gptel-curl-file-size-threshold on windows
On windows arguments passed to the subprocess are limited to [32,767](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa?redirectedfrom=MSDN) characters.

It took me a bit to figure out what was going wrong when chat buffers would get a little long. I'm wondering if it would be possible to detect the users platform and if it's windows set the gptel-curl-file-size-threshold a little lower? If not maybe adding something FAQ in the readme to help other folks who come across this.

AustinMooreT:
This makes sure that windows doesn't bark at you when you pass arguments to the curl subprocess over window's limit of 32767.

closes #517

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -255,7 +255,9 @@ Adjusting this value may be necessary depending on the environment
 and the typical size of the data being sent in GPTel queries.
 A larger value may improve performance by avoiding the overhead of creating
 temporary files for small data payloads, while a smaller value may be needed
-if the command-line argument size is limited by the operating system."
+if the command-line argument size is limited by the operating system.
+The default for windows comes from Microsoft documentation located here:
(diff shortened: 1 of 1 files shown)


karthink:
The force push is fine -- we can merge it after addressing my above comment about the paragraph spacing.

AustinMooreT:
Sorry I don't see your comment about paragraph spacing. I'm more of a gitlab user, so I may just be missing where it's at on github.

karthink:
Reproducing the comment from above here:

> Could you add an additional newline before "The default for windows comes from..." so it's on a new paragraph?

karthink:
Thanks @AustinMooreT!

PR #522, opened 2024-12-26
Remove gptel-extensions package

gavinhughes:
Package is now read-only and not maintained.


karthink:
Thanks for the update!

PR #529, opened 2024-12-30
Instructions for adding to the list of Gemini models

jlcheng:


karthink:
Thanks for the PR!

The model metadata specification method you've used is incorrect.  The correct way would be

``` emacs-lisp
(setf (gptel-backend-models gptel-backend)
      (append
       (gptel--process-models '((gemini-2.0-flash-thinking-exp
                                 :context-window 32
                                 ...)))
```
where the active `gptel-backend` is of type Gemini. (Your code omits `gptel--process-models`.)

As you can see, this method is quite complicated, and requires the use of an internal function `gptel--process-models`.  This can be achieved more simply by just pushing a model onto the model list. _i.e._ you can do this:

``` emacs-lisp
(push 'gemini-2.0-flash-thinking-exp (gptel-backend-models gptel-backend))
```

where the active `gptel-backend` is of type Gemini.  (Replace `gptel-backend` with a handle to the Gemini backend instead.)

This works for all backends, so can we turn this PR into a generic section of the README that doesn't pertain to any backend?  Here's what I'm picturing:

```org
**** Adding models to gptel backends

While gptel tries to keep up with new models for the major LLM API providers,
you don't have to wait.  You can add compatiable models to any gptel backend in
the following way:

- Find the name of the model you want to add.  In this example we will add
  =gemini-2.0-flash-thinking-exp= to a Gemini backend.
- Obtain a reference to the backend.  If the backend is currently active, this
  is just =gptel-backend=.
- Push a model onto the backend's list of models with
  #+begin_src emacs-lisp
  (push 'gemini-2.0-flash-thinking-exp (gptel-backend-models gptel-backend))
  #+end_src

*Optional*:

You can specify model metadata and capabilities (like vision or tool-use) as
properties of the model symbol:

#+begin_src emacs-lisp
(put 'gemini-2.0-flash-thinking-exp :capabilities '(media tool-use))
(put 'gemini-2.0-flash-thinking-exp :description
"Thinking Mode is capable of stronger reasoning capabilities.")
(put 'gemini-2.0-flash-thinking-exp :context-window 32)
#+end_src

See =gptel-make-gemini= (or equivalent) for recognized metadata fields.
```

In addition to this section, we can add an FAQ entry that links to it.

jlcheng:
> The model metadata specification method you've used is incorrect, as you need to provide my-gemini-models to gptel-make-gemini. 

Good catch! So dumb of me to miss that. In trying to remove my API key, I overlooked passing `my-gemini-models` to `gptel-make-gemini`.

I got an error when I initially tried these instructions. I added this to my `init.el` file. Can you spot my bug?

```
(require 'gptel)
(gptel-make-gemini "Gemini" :key my-gemini-api-key :stream t)
;; Here I assumed `gptel-backend` will magically work
(push 'gemini-2.0-flash-thinking-exp (gptel-backend-models gptel-backend))
```

This however, worked.

```
(require 'gptel)
(gptel-make-gemini "Gemini" :key my-gemini-api-key :stream t)
(push 'gemini-2.0-flash-thinking-exp (gptel-backend-models (cdr (assoc "Gemini" gptel--known-backends))))
````

So maybe we change the part about "Obtain a reference to the backend..."? I updated the PR with your suggestions and this modification:

```
- Push a model onto the backend's list of models with
  #+begin_src emacs-lisp
  (defvar my-gemini-backend (gptel-make-gemini "Gemini"
                              :key my-gemini-api-key :stream t)
    "Reference to the Gemini gptel Backend")
  (push 'gemini-2.0-flash-thinking-exp
        (gptel-backend-models my-gemini-backend))
  #+end_src
```

LMK what you think

**Comment by karthink** [comment]
Thanks for the update.  I think gptel needs a better way to update backend attributes (like the model).  The process of obtaining a reference to the backend is going to confuse users.  Let me think about this some more.  If you have any ideas for how this can be made more user-friendly, let me know.

**Comment by karthink** [comment]
@jlcheng I've added a helper, `gptel-get-backend`, that should make this process easier.  You can use it to get the backend by name:

``` emacs-lisp
(gptel-get-backend "Gemini")            ;Returns the Gemini backend
```

So you can add to its models like this:

``` emacs-lisp
(push 'gemini-2.0-flash-thinking-exp (gptel-backend-models (gptel-get-backend "Gemini")))
```

or modify other attributes:

```emacs-lisp
(setf (gptel-backend-host (gptel-get-backend "Gemini")) ...)
```

I think the instructions in this PR can be made simpler now?

karthink:
@jlcheng Are you still interested in this PR?

jlcheng:
No, I won't have time to adequately address this. I'll go ahead and close it.

PR #531, opened 2024-12-31
fix(private-gpt): passing system directive

rosenstrauch:
followup to https://github.com/karthink/gptel/commit/0a2c07a2a41e5d96e8886fd7e761eefe55caa647 for privategpt


karthink:
@rosenstrauch Thanks for the PR!

Issue #416, opened 2024-10-12
gptel--system-message is not robustly initialised.


metachip:
After reviewing the README.org file, I don't see any explicit warning about renaming or removing the 'default' prompt in the `gptel-directives` alist. I think this is potential issue because:

1. The package relies on the presence of a 'default' key in `gptel-directives` to initialize `gptel--system-message`.
2. The customization interface allows one to modify this variable name without warning about the consequences.
3. Renaming, removing or setting as nil the 'default' prompt uncovers a latent bug.

My suggestion:

1. Add a clear warning in the README.org file about not removing or renaming the 'default' prompt in `gptel-directives` and/or
2. Modify the definition of this variable to be more resilient to changes in `gptel-directives`. For example:
    ```elisp
    (defvar gptel--system-message
      (or (alist-get 'default gptel-directives)
        "You are a helpful assistant living in Emacs. Respond concisely.")
      "The system message used by gptel.")
   ``` 
Alternatively or as well, it might be a good idea to modify the `defcustom` for `gptel-directives` to use a custom widget that would disallow the user from doing this in the first place. Perhaps something like:

```elisp
(defcustom gptel-directives
  '((default . "You are a helpful assistant living in Emacs. Respond concisely.")
    ;; ... other directives ...
    )
  "System prompts (directives) for the LLM."
  :type '(list (cons :format "%v"
                     (const :format "" default)
                     (string :tag "Default prompt"))
               (repeat (cons :format "%v"
                             (symbol :tag "Name")
                             (string :tag "Prompt"))))
  :group 'gptel)
```
This would ensure the first item is always named `default` and that it can't be renamed, it can't be deleted. It allows customisation of the prompts value and it permits adding, modifying and removing other named prompts.

I think there would also need to `:set` to ensure the default value is never nil.

```elisp
:set (lambda (sym val)
       (set-default sym
                    (cons (cons 'default (or (cdar val)
                                             "You are a helpful assistant living in Emacs. Respond concisely."))
                          (cdr val))))
```
I've also added `:group 'gptel` as the group. Shouldn't it have been there in the first place?

karthink:
Thanks for bringing this up -- the system message system needs some work on the whole, I'll work on this when I address the other issues.  (Specifically, some users are finding the way the system message is stored/restored/applied confusing, so I have to rework the presentation.  I also plan to make it possible to specify functions instead of strings so the system message can be dynamically generated by the user.)

[...]

PR #547, opened 2025-01-06
Remove fallback value for `gptel--system-message`

tshu-w:
The fallback message "You are a large language model..." is no longer
needed since the default value of `gptel-directives` can be `nil`.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1017,8 +1017,7 @@ MODE-SYM is typically a major-mode symbol."
 
 
 (defvar gptel--system-message
-  (or (alist-get 'default gptel-directives)
-      "You are a large language model living in Emacs and a helpful assistant. Respond concisely.")
+  (alist-get 'default gptel-directives)
   "The system message used by gptel.")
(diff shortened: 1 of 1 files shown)


karthink:
This fallback was added on purpose, see #416.  The problem is that the user might not realize that the `default` key in `gptel-directives` is special and remove it via the customize interface or `custom-set-variables`.  In this case we'd still like to have a generic system message when gptel is loaded.

tshu-w:
However, currently users need to use (setq gptel--system-message nil) to use an empty system message.
> the system message system needs some work on the whole
It seems you need to work on the system message system, so I'll close this PR for now.

karthink:
> However, currently users need to use (setq gptel--system-message nil) to use an empty system message

The system message can be unset from the transient menu:

image

tshu-w:
I know it can be done this way. What I want to express is that setting the default of gptel-directives to nil to disable the system message doesn't work:

```emacs-lisp
(setq gptel-directives
      `((default . nil))
      gptel--system-message nil)
```

PR #559, opened 2025-01-09
Add support for OpenAI o1 model

tillydray:
o1 has superceded o1-preview, the costs are the same, o1 is either identical or better than o1-preview

### DIFF
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -154,7 +154,7 @@ with differing settings.")
     (when gptel-max-tokens
       ;; HACK: The OpenAI API has deprecated max_tokens, but we still need it
       ;; for OpenAI-compatible APIs like GPT4All (#485)
-      (plist-put prompts-plist (if (memq gptel-model '(o1-preview o1-mini))
+      (plist-put prompts-plist (if (memq gptel-model '(o1 o1-preview o1-mini))
                                    :max_completion_tokens :max_tokens)
                  gptel-max-tokens))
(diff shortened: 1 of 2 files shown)


karthink:
@tillydray Thanks for the PR!

We can't replace `o1-preview` with `o1` as we need to give users time to update their configurations -- otherwise gptel will throw an error that `o1-preview` is not available, and switch to the first available model in the OpenAI backend, which is... `gpt-4o` right now.  Could you 
- add `o1` as an additional model to the list of models for now, 
- and change the `:description` for `o1-preview` to just "deprecated: please use o1" or something like that,
- with a deprecation note in the git-commit message?

In the next release of gptel, I'll mention that `o1-preview` will be removed, and in the one after that we can remove `o1-preview`.

tillydray:
@karthink done!

benthamite:
I noticed that o1 does not use Markdown/Org to format its responses. This can be fixed by adding "Please use Markdown" to the prompt (which gets converted to Org in `org-mode` buffers), but it would be nice if it was the default behavior, as is the case with the other models.

karthink:
> I noticed that o1 does not use Markdown/Org to format its responses. This can be fixed by adding "Please use Markdown" to the prompt (which gets converted to Org in `org-mode` buffers), but it would be nice if it was the default behavior, as is the case with the other models.

I'm not sure how to do this.  I don't think it's a good idea to modify the user's prompt?

Issue #451, opened 2024-11-01
Some characters are sometimes incorrectly converted from Markdown to Org


benthamite:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [X] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

This is just another instance of the behavior reported here: #81. @karthink asked to create a new issue to report new cases, hence this issue.

In this case, the issue arises when converting Claude responses in JavaScript code blocks:

``` org

:PROPERTIES:
:GPTEL_MODEL: claude-3-5-sonnet-20241022
:GPTEL_BACKEND: Claude
:GPTEL_SYSTEM: You are a large language model living in Emacs and a helpful assistant. Respond concisely.
:GPTEL_BOUNDS: ((1398 . 4372) (4373 . 4428) (4429 . 4563) (4564 . 4600) (4601 . 4633) (4634 . 4722) (4723 . 4746) (4748 . 4813) (4814 . 4835) (4836 . 4862) (4863 . 4910) (4911 . 4957) (4958 . 5809) (5957 . 7290) (7883 . 12429) (13398 . 14398) (14922 . 16218) (16796 . 17682) (18858 . 20140) (20626 . 24145) (24283 . 26588) (26842 . 30118) (38317 . 40674) (47408 . 51035) (57399 . 60437) (60785 . 60932) (61097 . 62584) (66370 . 69409) (80001 . 83666) (84242 . 86193) (86750 . 88789) (89581 . 91572) (91853 . 92668) (92669 . 92690) (92691 . 93125) (93126 . 93187) (93188 . 93803) (111393 . 115189) (115360 . 117420))
:END:

<prior discussion omitted>

I see the issue - there are two problems:

1. We're getting token ID "0" which shouldn't happen. We need to fix the token ID lookup logic.
2. The market update is being blocked by a lock that isn't being released properly.

Let's fix both issues:

1. First, modify the lock handling in =locks.js= to add a timeout for stale locks:

#+begin_src javascript
export async function acquireLock(
    lockName,
    debouncePeriod = 30 * 1000,  // 30 seconds
    staleLockTimeout = 15 * 60 * 1000  // 15 minutes
) {
    let result;
    const trx = await getTransaction();
    const lockRow = getFirstRow(
        await trx.query({
            text: "select is_locked, last_updated from process_lock where id = $1",
            values: [lockName],
        }),
    );
    if (lockRow?.lock_state !== undefined) {
[...]

PR #561, opened 2025-01-09
Skip reformatting markdown to org in src-blocks

ElleNajt:
When using org mode for chat buffers, gptel--convert-markdown->org is called on response, replacing markdown code blocks with source code blocks,  " ** "'s with "*" (for bold text), and "#" with "*"s, among other changes.

This creates problems for code returned by chatbots, because the meaning of the code inside of source blocks gets changed.

To solve this, this commit modifies gptel--convert-markdown->org in the following way:
1. the temp buffer created by is set to be org-mode
2. Then the org mode built in org-in-src-block-p is used to skips replacements if inside of source blocks.

This works because the backticks are replaced with source blocks before all of the other formatting replacement code. However, this could be made more explicit. 

The issue here can be tested by:
- asking gptel to return some python code with lines commented out by #'s
- asking gptel to return a function with x ** 2 (this is failing inconsistently, see photo)

See attached photo. 
image

Considering this further -- I think it would make sense to have the option to only apply the markdown block to source block adjustment. I personally don't have a reason to care about bold or italic text in the response, and losing meaning (as in the snippet outside of the source block) seems like a worse outcome. What are your thoughts?

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -436,52 +436,54 @@ elements."
   (interactive)
   (with-temp-buffer
     (insert str)
+    (org-mode)
     (goto-char (point-min))
     (while (re-search-forward "`+\\|\\*\\{1,2\\}\\|_\\|^#+" nil t)
-      (pcase (match-string 0)
-        ;; Handle backticks
-        ((and (guard (eq (char-before) ?`)) ticks)
-         (gptel--replace-source-marker (length ticks))
-         (save-match-data
-           (catch 'block-end
-             (while (search-forward ticks nil t)
-               (unless (or (eq (char-before (match-beginning 0)) ?`)
-                           (eq (char-after) ?`))
+      (unless (org-in-src-block-p)
+        (pcase (match-string 0)
+          ;; Handle backticks
+          ((and (guard (eq (char-before) ?`)) ticks)
+           (gptel--replace-source-marker (length ticks))
+           (save-match-data
+             (catch 'block-end
+               (while (search-forward ticks nil t)
+                 (unless (or (eq (char-before (match-beginning 0)) ?`)
+                             (eq (char-after) ?`))
                    (gptel--replace-source-marker (length ticks) 'end)
                    (throw 'block-end nil))))))
-        ;; Handle headings
-        ((and (guard (eq (char-before) ?#)) heading)
-         (when (looking-at "[[:space:]]")
-           (delete-region (line-beginning-position) (point))
-           (insert (make-string (length heading) ?*))))
-        ;; Handle emphasis
-        ("**" (cond
-               ;; ((looking-at "\\*\\(?:[[:word:]]\\|\s\\)")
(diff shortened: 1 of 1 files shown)


karthink:
> When using org mode for chat buffers, gptel--convert-markdown->org is called on response, replacing markdown code blocks with source code blocks, " ** "'s with "" (for bold text), and "#" with ""s, among other changes.
> 
> This creates problems for code returned by chatbots, because the meaning of the code inside of source blocks gets changed.

This is not exactly quite how the converter works.  Before I investigate the bug: was the above screenshot from a streaming response or a chunked response?  The converters used for the two are different.

karthink:
Okay.  It will help to compare against the JSON that was received.  Could you run `(setq gptel-log-level 'info)` and use gptel until you see the bug from your screenshot, then check the `*gptel-log*` buffer and paste the JSON from the response here?

karthink:
Thank you.  The problem is here:

```json
    {
      "role": "assistant",
      "content": "Here's a simple Python function that squares numbers with comments:\n\n#+begin_src python\ndef square_number(x):\n    # Calculate x raised to power 2\n    result = x ** 2\n    return result\n\n# Example usage\nnum = 5\nprint(f\"{num} squared is {square_number(num)}\")  # Output: 5 squared is 25\n#+end_src"
    },
```

Claude responded in Org syntax instead of Markdown.  See #451 for details.  There are two solutions: you can either turn off markdown -> org conversion (see `gptel-org-convert-response`) or ask Claude in your system message to only reply in markdown.  An LLM smart enough to automatically try to use Org markup based on previous output is hopefully also smart enough to follow your instructions.

Your proposed method of turning on org-mode is not a good idea for gptel (at least right now), as
- the text in the buffer is not in Org syntax, it's in Markdown
- turning on org-mode can be very slow and block Emacs

It might be possible to handle LLMs misbehaving like this with non-streaming responses, I'll leave this PR open while I investigate. (Handling this with streaming responses is out of scope for now.)

ElleNajt:
Interesting, thanks!

What about, as a third option, having the option for a lighter markdown->org conversion function that only replaces markdown code blocks with org source blocks, and otherwise doesn't touch any other formatting? 

The originally proposed commit does consistently solve the misbehavior, though I agree that turning on org-mode can slow things down. If there's no need to check that something is in an source block, then there's no need to turn on org-mode, so a lighter conversion function wouldn't need to do that.

>  An LLM smart enough to automatically try to use Org markup based on previous output is hopefully also smart enough to follow your instructions.

This has been tricky to ensure in my experimentation, I think it's more reliable to do a programmatic replacement for when markdown blocks creep in.

karthink:
I pushed a fix, can you try it now?  You can try streaming responses too, that should also be handled.

ElleNajt:
Working for me now, thanks! I'll go ahead and close this PR. :)

Issue #565, opened 2025-01-13
Use zero-width delimiters for role tracking in gptel-mode


lispy-ai:
# Title: Use zero-width delimiters for role tracking with overlay-based highlighting in gptel-mode buffers

This is a proposal for a new approach to tracking and visually distinguishing assistant/user roles in gptel buffers that addresses several long-standing issues (#321, #343) while maintaining compatibility with the existing system.

## The Problem

Currently gptel uses text properties to track which sections of text are assistant responses. This approach has proven problematic because:

1. Text properties don't interact naturally with standard Emacs editing operations
2. Property stickiness creates ambiguous cases during editing
3. Yanked text carries properties that can cause confusion
4. Visual feedback about roles is difficult to implement reliably

## Proposed Solution

Use zero-width Unicode characters as role delimiters with overlay-based highlighting, but only when gptel-mode is active:

- U+200B (zero-width space) marks response start
- U+200C (zero-width non-joiner) marks response end
- Overlays provide visual distinction for responses

Key aspects:
- Delimiters are invisible and don't affect buffer display
- Standard editing operations work naturally
- Cut/paste preserves role boundaries correctly
- Overlays provide clean visual feedback
- Works with all major modes

## Implementation

The solution uses two phases:

1. When gptel-mode is enabled:
   - Convert existing gptel text properties to delimiter pairs
   - Remove gptel properties
   - Enable delimiter-based role tracking
   - Create overlays for responses

2. When gptel-mode is disabled:
   - Convert delimiters back to gptel properties
   - Remove delimiters
   - Remove overlays
   - Restore property-based tracking

### Response Highlighting

Use overlays for visual distinction:
- Clean visual distinction
- No interference with text properties
- Preserves other modes' fontification
- Easy to customize appearance

## Benefits

1. Reliable editing operations:
   - Cut/paste works naturally
   - Undo/redo maintains role boundaries
   - No property stickiness issues
   
2. Better user experience:
   - Clear visual distinction of responses
   - Predictable editing behaviour
   - Compatible with standard Emacs commands
   - Non-intrusive highlighting
   
3. Technical improvements:
   - Simple to parse conversation history
   - Clean visual feedback via overlays
   - Works with all major modes
   - Separation of tracking and display
   
## Testing

To test this change:
1. Enable gptel-mode in a buffer with existing responses
2. Verify properties convert to delimiters correctly
3. Test editing operations (especially cut/paste)
4. Verify overlay highlighting
[...]

Issue #546, opened 2025-01-05
Making gptel more useful in non-chat buffers


grothesque:
Edit: Originally, the title of this issue was “Please consider not making prompt prefixes markdown/org headings by default”.  I changed it to better reflect the way into which the discussion had evolved.

-----------------

Thank you very much for this impressive tool that is helping me discover a new world of possibilities.

I have the impression that the default choice of prompt prefixes (`gptel-prompt-prefix-alist` and `gptel-response-prefix-alist`) being headings is sub-optimal both in principle and in practice:

In principle it means that the first line of each prompt is marked-up as a heading, while the following lines are regular text.  This distinction, however, is not in any way visible to the model, and has therefore no effect.  Often the first line of a multi-line prompt will not have any particular importance - why mark it up differently?

In practice, if one tries to use other headings in a `M-x gptel` buffer, the fixed "prompt headings" tend to mess up the structure.  At least with org, the possibility to structure the document as a tree is at the heart of what it is about.  Gptel's readme file suggests a solution: https://github.com/karthink/gptel?tab=readme-ov-file#use-branching-context-in-org-mode-tree-of-conversations.  How about making this the actual default?  Or something else that does not introduce an inconsistency between visuals and semantics, and that does not interfere with the document structure.

karthink:
> In principle it means that the first line of each prompt is marked-up as a heading, while the following lines are regular text. This distinction, however, is not in any way visible to the model, and has therefore no effect. Often the first line of a multi-line prompt will not have any particular importance - why mark it up differently?

I agree!  It doesn't work well, I'd like to change it.  The reasons it's still the default are:
1. It works for very simple question/answer style interactions.  This is all many users use gptel for.
2. Users have "notebooks" of stored gptel chats.  If they use the default prefixes and I change them,
   - The heading characters will be sent to the LLM in subsequent interactions.  (The choice of prefixes are not stored in the markdown files.)
   - The continuation of the chat will be formatted differently from the existing exchanges.
3. Any change is going to confuse a lot users.  You'd be surprised at how many gptel users believe the current heading/body prefixes are part of gptel's "syntax" for differentiating between prompts and responses.  Most users, even Emacs users, don't look at customization options or read the README beyond the bare minimum needed to install the package[1].
4. I don't know what good defaults should be.  If change it to "@user: " and "@assistant: " or something like that, this will make problem 3 worse, as this looks even more like special, required syntax.

[...]

PR #573, opened 2025-01-17
Dedicated buffer like gptel-inspect-query, but just to edit the response at point with the same major mode

axelknock:
I quite like the ability to edit LLM responses since it's a powerful method for guidance. Currently it's a bit finnicky to edit responses, as you have to pay attention to the boundaries of the response, and then use a self-defined function to mark the region of your edits as part of the response with the text property.

This interactive function is just designed to make that easier. It's also possible to define a wrapper for editing specific responses:
```emacs-lisp
(defun my/gptel-edit-last-response ()
  "Use `gptel-edit-response' to edit the last response before point."
  (interactive)
  (gptel-edit-response
   (progn (gptel-beginning-of-response nil nil current-prefix-arg)
          (forward-char (+ 1 (length (gptel-response-prefix-string)))))))
```
(inconsistent result when I don't move point forward into response by 1 char)

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -658,6 +658,12 @@ This opens up advanced options in `gptel-menu'.")
 (defvar-local gptel--bounds nil)
 (put 'gptel--bounds 'safe-local-variable #'always)
 
+(defvar-local gptel--edit-response-buffer nil
+  "Original buffer for gptel response being edited.")
+
+(defvar-local gptel--edit-response-pos nil
+  "Original position in `gptel--edit-response-buffer' where editing started.")
+
 (defvar gptel--num-messages-to-send nil)
 (put 'gptel--num-messages-to-send 'safe-local-variable #'always)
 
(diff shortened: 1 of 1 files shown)


karthink:
Have you seen #343?

axelknock:
> Have you seen #343?

Yes. I assume you are referring to it to point out that there is ongoing discussion on how best to track responses in buffers, and `gptel-edit-response` does not necessarily fit in with whatever might be used in the future. I would be happy for there to be a better in-buffer solution and have this function deprecated in the future, but right now I see it as the best solution to edit responses in a robust manner, which I find necessary.

`gptel-edit-response` is also technically independent of any further changes to the way in which responses are tracked, since it just relies on finding `gptel--get-bounds` which will still be functional even if its internals change to using overlays or something else. But of course it would be redundant if you could reliably edit the response in the current buffer.

I also noticed that `gptel--inspect-query` was changed to allow writing, which is another way to achieve an edited response, but I confess I do not understand in what case you would want to send a version of the current buffer that has been altered without committing those changes to the original buffer.

I've also seen #546 and commented on #565 which relate to this.

axelknock:
It looks like the response text property is now inherited from the surrounding region.

karthink:
> It looks like the response text property is now inherited from the surrounding region.

While that's true, it's not sufficiently robust -- you can yank text into a response region and it won't be categorized as the response.  So your PR remains one way to avoid this problem.

However, I have some more ideas to make it robust that I want to try before this idea or #343.

Issue #141, opened 2023-11-28
Request for comment: Use Org mode properties drawer to configure system prompt/temp for the whole subtree


doctorguile:
I ran into the same bug as https://github.com/karthink/gptel/issues/135

But ultimately even if I didn't run into this bug, I think the transient menu is not as convenient if you have a big list of predefined combos.

For an example of very convenient way to have access to a set of user defined system prompt/model/temperature, etc see

https://github.com/Bin-Huang/chatbox

where you can set a system prompt , temperature  and model for each chat thread, and continue on that thread without having to repeat the context (sys prompt, etc)

the ideal way to do this in emacs is in org mode, having a header and set system prompt , temperature and model as property drawer.

then all subtree of that header will inherit these properties and you can have a lot of useful prompt / temperature always ready to go

example

```
* Emacs
:PROPERTIES:
:SYSPROMPT: You are a emacs expert. You can help me by answering my questions. You can also ask me questions to clarify my intention.
:temperature: 0.1
:model: GPT4
:END:

** How to call REST api?
conversation re emacs continue in this subtree...

* Python
:PROPERTIES:
:SYSPROMPT: you are a python expert. respond to the task with detailed step by step instructions and code. list out all the files to be created and how to put them together
:temperature: 0.2
:model: GPT3.5
:END:

** How to setup a webserver with http3 support?
conversation re python continue in this subtree...
```

Does anyone think this is a good addition? This is how org is used (inheriting parent props) for a lot of other use cases

doctorguile:
Forget to mention I'm using the latest git clone Nov 20 on macOS

```
ommit de6d8089cdf68732a3942ef729f774322120175e (grafted, HEAD -> master, origin/master, origin/HEAD)
Date:   Mon Nov 20 11:21:50 2023 -0800

    gptel-transient: Fix system-message setting function
    
    gptel-transient.el (gptel--suffix-system-message): Removing the
    `(setf (buffer-local-value ...))` construct (as instructed to by
    the byte compiler) introduced a bug where custom system message
    were set from the wrong buffer.  Handle this correctly to fix #138
    and possibly #140.
    ```

karthink:
This works on a per-file basis in Org-mode right now if you save the buffer. (Except for Ollama models, where it's more annoying to capture state.)

I'm not convinced it's worth adding per-heading support, but I'll think about it.

NightMachinery:
[...]

PR #575, opened 2025-01-20
gptel-org: put hardcoded gptel-org-property-names into variable 

hrehfeld:
so other (potentially external) code can reference them.

e.g. I have org property inheritance disabled but want it enabled for gptel props:

``` elisp
  (cl-loop
   for prop in '("GPTEL_SYSTEM" "GPTEL_BACKEND" "GPTEL_MODEL"
            "GPTEL_TEMPERATURE" "GPTEL_MAX_TOKENS"
            "GPTEL_NUM_MESSAGES_TO_SEND")
   do
   (add-to-list 'org-use-property-inheritance prop))
```

becomes

``` elisp
  (cl-loop
   for prop in gptel-org-property-names
   do
   (add-to-list 'org-use-property-inheritance prop))
```

GPTEL_BOUNDS is purposely not included, because it seems like only used file globally.

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -327,6 +327,15 @@ ARGS are the original function call arguments."
 ;; (advice-add 'gptel-request :around #'gptel-org--send-with-props)
 
 
+
+(defconst gptel-org-property-names '("GPTEL_SYSTEM" "GPTEL_BACKEND" "GPTEL_MODEL"
+                                     "GPTEL_TEMPERATURE" "GPTEL_MAX_TOKENS"
+                                     "GPTEL_NUM_MESSAGES_TO_SEND" "GPTEL_TOPIC")
+  "Org property names that are relevant to gptel configuration.
+
+These could be useful e.g. to add to `org-use-property-inheritance' to enable
+inheritance of gptel configuration.")
+
 ;;; Saving and restoring state
 (defun gptel-org--entry-properties (&optional pt)
   "Find gptel configuration properties stored at PT."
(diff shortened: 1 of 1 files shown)


karthink:
@hrehfeld thanks for the PR.  While I'm okay with placing these in a variable, you can achieve the effect you want more easily via
```emacs-lisp
(setq org-use-property-inheritance "^GPTEL_")
```
Any reason you don't want to use this?

1. Could you make this `defvar` a `defconst`?  
2. Do you think the name `gptel-org-properties` might be better?  If Org mode already has a convention for one of these (`-properties` vs `-property-names` for lists of properties) we can stick to it.

As an aside, not making these properties inheritable by default is intentional, see [my comment](https://github.com/karthink/gptel/issues/141#issuecomment-1986489331) about this in #141.

hrehfeld:
Fixed `defconst`. Also improved documentation, gave my use case as an example.

hrehfeld:
> @hrehfeld thanks for the PR. While I'm okay with placing these in a variable, you can achieve the effect you want more easily via
> 
> ```emacs-lisp
> (setq org-use-property-inheritance "^GPTEL_")
> ```
> 
> Any reason you don't want to use this?
> 
>     1. Could you make this `defvar` a `defconst`?
> 
>     2. Do you think the name `gptel-org-properties` might be better?  If Org mode already has a convention for one of these (`-properties` vs `-property-names` for lists of properties) we can stick to it.
> 
> 
> As an aside, not making these properties inheritable by default is intentional, see [my comment](https://github.com/karthink/gptel/issues/141#issuecomment-1986489331) about this in #141.

Well, unless you're claiming the whole `GPTEL_` property namespace ;-) it just feels less robust. And I figure my PR wouldn't cause you more work, as you're listing the properties anyways.

hrehfeld:
But wait. The code where list was hardcoded before is actually a pcase-let. So one CAN'T easily put new property names into my constant without also changing that code. (`pcase-let doc: a mismatch may signal an error or may go
undetected, binding variables to arbitrary values, such as nil.`)

So we'd actually need another defconst just for that destructure, or hardcode it anyways  and just ADD the constant. Or just give up on this?

karthink:
@hrehfeld You're right, this function combines property names with separate logic for handling each corresponding property value.  So adding or removing topics from the defconst won't be enough, you'll also have to provide the logic.  I'm not sure it's worth writing and managing separate functions for handling each property at this point.

> Well, unless you're claiming the whole GPTEL_ property namespace

I'm not seeing the problem with this, at least as far as inheritance is concerned?

karthink:
> So we'd actually need another defconst just for that destructure, or hardcode it anyways and just ADD the constant. Or just give up on this?

I'm leaning towards closing this PR.  Before I do that, I'd like to take a step back and reconsider the need for this.  What do you want to be able to do with `gptel-org-property-names` exactly?

Issue #593, opened 2025-01-30
gptel chokes on nested brackets containing one or more characters


tillydray:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

Sending the following will cause `Lisp error: (wrong-type-argument integer-or-marker-p nil)`

```org
#+begin_src
[[ ]]
#+end_src
```

But this works just fine

```org
#+begin_src
[[]]
#+end_src
```

I suspect this impacts every backend

### Backend

None

### Steps to Reproduce

1. create a new gptel buffer
2. paste this into the buffer
```
#+begin_src
[[ ]]
#+end_src
```
3. call `gptel-send` or do a dry run
4. see `Lisp error: (wrong-type-argument integer-or-marker-p nil)` in *Messages*
5. request isn't sent

### Additional Context

- GNU Emacs 30.0.93 (build 2, aarch64-apple-darwin24.0.0, NS appkit-2566.00 Version 15.0.1 (Build 24A348)) of 2024-12-28
- macOS 15.0.1
- doom emacs

### Backtrace

```emacs-lisp
Debugger entered--Lisp error: (wrong-type-argument integer-or-marker-p nil)
  gptel-org--link-standalone-p((src-block (:standard-properties [6949 6949 nil nil 7066 1 nil nil element t (21 . 6949) nil nil nil #<buffer embeddings.org> nil nil (section (:standard-properties [6315 6315 6315 8151 8151 0 nil section element t (21 . 6314) 6315 8151 nil #<buffer embeddings.org> nil nil (headline ...)] :fragile-cache nil))] :language "ts" :switches nil :parameters ":tangle yes" :number-lines nil :preserve-indent nil :retain-labels t :use-labels t :label-fmt nil :value "function calcTermFreq(doc: string): [string, number][] {\n  return [[\"\", 0]];\n}\n")))
  #f(compiled-function (mode beg end) "Parse text and actionable links between BEG and END.\n\nReturn a list of the form\n ((:text \"some text\")\n  (:media \"/path/to/media.png\" :mime \"image/png\")\n  (:text \"More text\"))\nfor inclusion into the user prompt for the gptel request." #<bytecode 0x130e9247cdbaa5ce>)(org-mode 6946 7237)
[...]

PR #594, opened 2025-01-30
Improve robustness: `gptel-org--link-standalone-p`

tillydray:
- Added checks to ensure paragraph (`par`) is valid before accessing its
- properties.  Safeguarded against `nil` values for `:contents-begin`
- and `:contents-end` properties.  Ensured the function proceeds only if
- all necessary properties are non-nil.

These changes prevent errors when element properties are unexpectedly `nil`, like when trying to send this, with one or more characters between the brackets

\#+begin_src
[[ ]]
\#+end_src

closes https://github.com/karthink/gptel/issues/593

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -291,10 +291,10 @@ for inclusion into the user prompt for the gptel request."
   (when-let* ((par (org-element-lineage object '(paragraph))))
     (and (= (gptel-org--element-begin object)
             (save-excursion
-              (goto-char (and par (org-element-property :contents-begin par)))
+              (goto-char (org-element-property :contents-begin par))
               (skip-chars-forward "\t ")
               (point)))                 ;account for leading space before object
-         (<= (- (and par (org-element-property :contents-end par))
+         (<= (- (org-element-property :contents-end par)
                 (org-element-property :end object))
              1))))
 
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -294,7 +294,8 @@ for inclusion into the user prompt for the gptel request."
               (goto-char (and par (org-element-property :contents-begin par)))
               (skip-chars-forward "\t ")
               (point)))                 ;account for leading space
-         (<= (- (and par (org-element-property :contents-end par))(org-element-property :end object))
+         (<= (- (and par (org-element-property :contents-end par))
+                (org-element-property :end object))
              1))))
 
 (defun gptel-org--send-with-props (send-fun &rest args)
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -287,17 +287,17 @@ for inclusion into the user prompt for the gptel request."
 
 (defun gptel-org--link-standalone-p (object)
   "Check if link OBJECT is on a line by itself."
-  ;; Specify ancestor TYPES as list (#245)
-  (let ((par (org-element-lineage object '(paragraph))))
-    (and (= (gptel-org--element-begin object)
-            (save-excursion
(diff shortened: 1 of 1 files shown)


gptel-org.el
```diff
@@ -287,17 +287,17 @@ for inclusion into the user prompt for the gptel request."
 
 (defun gptel-org--link-standalone-p (object)
   "Check if link OBJECT is on a line by itself."
-  ;; Specify ancestor TYPES as list (#245)
```
karthink, line 290: Please retain this comment. (Or I will make the same mistake in the future.)

gptel-org.el
```diff
@@ -287,17 +287,17 @@ for inclusion into the user prompt for the gptel request."
...
-    (and (= (gptel-org--element-begin object)
-            (save-excursion
-              (goto-char (org-element-property :contents-begin par))
-              (skip-chars-forward "\t ")
-              (point)))                 ;account for leading space
-                                        ;before object
-         (<= (- (org-element-property :contents-end par)
-                (org-element-property :end object))
-             1))))
+  (let* ((par (org-element-lineage object '(paragraph)))
```
karthink, line 290: Can be simplified to just `(when-let* ((par (org-element-lineage object '(paragraph))))...)`, and you can remove the `(when (and begin end)...)`.  If there's a paragraph in the lineage, it is guaranteed to have a non-nil beginning and end positions.

You don't need to bind `begin` and `end` since their values will only be used once.

gptel-org.el
```diff
@@ -287,17 +287,17 @@ for inclusion into the user prompt for the gptel request."
 
 (defun gptel-org--link-standalone-p (object)
   "Check if link OBJECT is on a line by itself."
-  ;; Specify ancestor TYPES as list (#245)
-  (let ((par (org-element-lineage object '(paragraph))))
-    (and (= (gptel-org--element-begin object)
-            (save-excursion
-              (goto-char (org-element-property :contents-begin par))
-              (skip-chars-forward "\t ")
-              (point)))                 ;account for leading space
```
karthink, line 296: Please retain this comment too.  You can move it to be on one line.

karthink (changes requested):
Thanks for the PR!  The fix can be simplified, please see review.

gptel-org.el
tillydray: I only ever write elisp for my literate config so I definitely need guidance with it, thank you!

tillydray:
@karthink should be ready for re-review

gptel-org.el
```diff
@@ -288,14 +288,13 @@ for inclusion into the user prompt for the gptel request."
 (defun gptel-org--link-standalone-p (object)
   "Check if link OBJECT is on a line by itself."
   ;; Specify ancestor TYPES as list (#245)
-  (let ((par (org-element-lineage object '(paragraph))))
+  (when-let* ((par (org-element-lineage object '(paragraph))))
     (and (= (gptel-org--element-begin object)
             (save-excursion
-              (goto-char (org-element-property :contents-begin par))
+              (goto-char (and par (org-element-property :contents-begin par)))
```
karthink, line 294: This will never be reached if `par` is nil, because of the `when-let*`.  So all you really need is the original code but with the `let*` replaced by a `when-let*`.

If you're not sure how this works, try macroexpanding the `when-let*` clause.

karthink:
Please see comment.

karthink:
Thank you @tillydray!

PR #596, opened 2025-01-30
Pass nil results through to model

psionic-k:
When the tool call succeeds (as in does not error) but returns nil, we still need to show the model that the tool returned a nil result.

Otherwise the model seems to not progress in its reply.

`gptel--to-string` handles this fine and the model happily reported observing the "nil" result.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1931,10 +1931,9 @@ Run post-response hooks."
                   (arg-values)
                   (process-tool-result
                    (lambda (result)
-                     (when result
-                       (plist-put info :tool-success t)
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR!

Technically we should be passing `null` or `false` as appropriate instead of `nil`, but JSON-encoding tool results is a can of worms that I don't want to open yet.  If you have any ideas in this department please let me know in the tool-use thread #514.  (Please discuss it with me first if you want to make a PR, as there are some issues with consistently JSON-encoding tool call results.)

PR #599, opened 2025-02-01
Variable number of newlines before responses

psionic-k:
With fancy graphical headings, empty lines are unneeded and actually quite bad.

I can't imagine anyone wants values other than 1 or 2.  Maybe this should be...

`gptel-response-blank-line'?

I can rework it.  Probably the right way to decide is "what will users recognize on first glance?"

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -372,6 +372,9 @@ This is an alist mapping major modes to the reply prefix strings.  This
 is only inserted in dedicated gptel buffers before the AI's response."
   :type '(alist :key-type symbol :value-type string))
 
+(defcustom gptel-response-separator "\n\n"
+  "String inserted before responses.")
+
 (defcustom gptel-use-header-line t
   "Whether `gptel-mode' should use header-line for status information.
 
@@ -2321,7 +2324,7 @@ See `gptel--url-get-response' for details."
         (when-let* ((transformer (plist-get info :transformer)))
           (setq response (funcall transformer response)))
         (when tracking-marker           ;separate from previous response
-          (setq response (concat "\n\n" response)))
+          (setq response (concat gptel-response-separator response)))
         (save-excursion
           (add-text-properties
            0 (length response) '(gptel response front-sticky (gptel)) response)
@@ -2330,7 +2333,7 @@ See `gptel--url-get-response' for details."
             ;; (run-hooks 'gptel-pre-response-hook)
             (unless (or (bobp) (plist-get info :in-place)
                         tracking-marker)
-              (insert "\n\n")
+              (insert gptel-response-separator)
               (when gptel-mode
                 (insert (gptel-response-prefix-string)))
               (move-marker start-marker (point)))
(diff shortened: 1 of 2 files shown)


karthink:
> With fancy graphical headings, empty lines are unneeded and actually quite bad.

Can you give me an example?

psionic-k:
image

My headings have line-spacing and line-height already, so empty lines are extraneous.  I realize terminal users will prefer newlines all the time and prioritize plain-text readability, but it's Emacs and we make different choices.

karthink:
Values of 0, 1 and 2 newlines make sense here.  Maybe this can be specified more generally as a string?  Like `gptel-response-separator`.  There are times when I've wanted no response separator.

> but it's Emacs and we make different choices.

FWIW, I don't think it's a good idea to make user prompts headings, at least in Org mode.  You end up bundling unrelated semantic aspects of the document, and features like context restriction to headings and `gptel-org-branching-context` don't work.  Every LLM conversation that goes on for a while becomes a tree anyway, so I find branching context essential.

The default prompt prefix remains a heading for historical reasons.

Aside: Adjusting the spacing is also something you can easily do in `gptel-post-response-functions`.

``` emacs-lisp
(defun gptel-adjust-spacing (beg _end)
  (save-excursion
    (goto-char beg)
    (when (looking-back "\n\n" (- (point) 2))
      (delete-char -1))))

(add-hook 'gptel-post-response-functions #'gptel-adjust-spacing)
```

psionic-k:
> FWIW, I don't think it's a good idea to make user prompts headings, at least in Org mode.

To the extent that it's true, it's another example of displaying a view fighting against the view being literal context.  It's a good example case for needing a 'gptel-ignore text property.  If my request and response headings are ignored, it doesn't matter.

But I do also see the case for the branching context.  

For now I'm using 2nd and 3rd level headings for my context.  Because the model is mimicking the input, it occasionally throws out a first level heading, which would break the tree.  I have suppressed this in the system prompt, but 4o-mini just loves to ape whatever is in context.

> Maybe this can be specified more generally as a string

I agree.  I'll re-work it as such.

karthink:
> To the extent that it's true, it's another example of displaying a view fighting against the view being literal context.

Yes, this is why branching context is disabled by default -- if you turn it on, you know what you're doing.

psionic-k:
`response-separator` :heavy_check_mark:

gptel.el
```diff
@@ -372,6 +372,9 @@ This is an alist mapping major modes to the reply prefix strings.  This
 is only inserted in dedicated gptel buffers before the AI's response."
   :type '(alist :key-type symbol :value-type string))
 
+(defcustom gptel-response-separator "\n\n"
+  "String inserted before responseses.")
```
karthink, line 376: `responseses`: intentional?

psionic-k: I was kidnapped by people who live in row-houses in Brooklyn.

karthink:
Thanks for the PR!

PR #600, opened 2025-02-01
Parameters Required

psionic-k:
```
:TOOL_CALL:
(all_arg_types (:foo 42))

object: (:foo 42)
```
:boom: 

:shipit: 

Related tool example found [here](https://github.com/karthink/gptel/issues/514#issuecomment-2627491244)

I found this by inspecting the tool use and noticing a `:parameters nil` in the plist describing the tool to the backend.  The object spec wasn't being given to the backend at all.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1567,8 +1567,10 @@ implementation, used by OpenAI-compatible APIs and Ollama."
                              ,@(if enum (list :enum (vconcat enum)))
                              ,@(cond
                                 ((equal type "object")
-                                  :additionalProperties :json-false))
                                  (list :properties (plist-get arg :properties)
+                                       :required (or (plist-get arg :required)
+                                                     (vector))
+                                       :additionalProperties :json-false))
                                 ((equal type "array")
                                  ;; TODO(tool) If the item type is an object,
                                  ;; add :additionalProperties to it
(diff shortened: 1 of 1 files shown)


gptel.el
```diff
@@ -1567,8 +1567,9 @@ implementation, used by OpenAI-compatible APIs and Ollama."
                              ,@(if enum (list :enum (vconcat enum)))
                              ,@(cond
                                 ((equal type "object")
-                                 (list :parameters (plist-get arg :parameters)
-                                  :additionalProperties :json-false))
+                                 (list :properties (plist-get arg :properties)
+                                       :required (plist-get arg :required)
```
karthink, line 1571: If `:required` is missing in the object specification this becomes `:required nil` -- it should become `:required []` instead.

karthink:
Thanks for catching this.

gptel.el
psionic-k: Done https://github.com/karthink/gptel/pull/600/files#diff-58ae8200f2acac302967edbab7a33d6a0a6832a2b6dae6868375f4e5d1968795R1572

karthink:
> I found this by inspecting the tool use and noticing a :parameters nil in the plist describing the tool to the backend.

In case you weren't aware: you can inspect the query payload from the transient menu.

image

In this case the bug showed up in the query payload.  You only need to actually send requests to check how the LLM responds or aspects of response handling.

PR #601, opened 2025-02-01
gptel-transient: allow generating output directly to the *scratch* buffer

jwr:
I use the `*scratch*` buffer extensively when working with text and code, and I find it very useful to redirect the output from gptel directly to `*scratch*`. I've been maintaining my own branch for a while, but perhaps others would find this useful, too?

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -510,6 +510,7 @@ Also format its value in the Transient menu."
      :reader
      (lambda (prompt _ _history)
        (read-buffer prompt (buffer-name (other-buffer)) nil)))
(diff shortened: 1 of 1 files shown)


karthink:
@jwr Thanks for the PR!

Since the menu already offers a way to redirect the output to any buffer (including `*scratch*`), I don't think it makes sense to add this.

Closing as not planned.

jwr:
Understood! Although I would argue that `*scratch*` is special — but perhaps most people do not use it as much as I do.

Not a problem — that's what local forks with local modifications are for!

Issue #294, opened 2024-04-25
gptel-org-branching-context requires Org 9.7 (unreleased)


localauthor:
Hi Karthik, first, thanks for gptel. It is truly wonderful.

It seems that `gptel-org-branching-context` depends on the function `org-element-lineage-map`, which is not included in the current release of Org (9.6.27), at least as far as I can tell.

It seems that the file containing that function, org-element-ast.el, is in org's main branch, but 9.6 is not built with it. I do not know why it is excluded, at least in my case.

Might the branching-context feature be adapted (without too much trouble) for those of us on just this side of org's bleeding edge? Thanks for considering it, in any case.

karthink:
Flying by the seat of my pants here -- I bundled `org-element-lineage-map` with gptel, to be evaluated at compile time if it's not natively available.  Please let me know if it works, or if there are transitive dependencies that are also missing.

localauthor:
Looks like there are quite a few other functions that would also need to be bundled:

`org-element-parent`
`org-element-begin`
`org-element-type`
`org-element-type-p`
`org-element-secondary-p`

Not sure if bundling all these is something you would want to do.

In the meantime, I've just defined all these in my init file and `gptel-org-branching-context` works great.

karthink:
Oof, my fix was shortsighted then, as I expected.

I'm not sure what I can do here.  I can throw a warning when setting `gptel-org-branching-context` if the org version is not high enough, and disable it.  But this warning will only fire if setting the variable using setopt or the customization menu, which I doubt most users are using.

localauthor:
Or maybe instead of a warning when setting the variable, a warning when calling gptel-send in a context when a user might expect it to work, but they don’t have the right org version. Or even simpler (and what I would do) just a note about the org version requirement in the doc string. Dont think it’s worth much more effort than that, honestly. 
    
  

 From: karthink ***@***.***>Sent: Sunday, April 28, 2024 06:22To: karthink/gptel ***@***.***>Cc: Grant Rosson ***@***.***>; Author ***@***.***>Subject: Re: [karthink/gptel] gptel-org-branching-context requires Org 9.7 (unreleased) (Issue #294) 
Oof, my fix was shortsighted then, as I expected.
I'm not sure what I can do here.  I can throw a warning when setting gptel-org-branching-context if the org version is not high enough, and disable it.  But this warning will only fire if setting the variable using setopt or the customization menu, which I doubt most users are using.

ultronozm:
[...]

PR #602, opened 2025-02-01
Updates required Org version notice from 9.6.7 to 9.7

grettke:
[This explains](https://github.com/karthink/gptel/issues/294) that gptel needs Org version 9.7+ so this patch updates the notice for the user.

### DIFF
--- a/README.org
+++ b/README.org
@@ -1120,7 +1120,7 @@ If this variable is non-nil, you should probably edit =gptel-prompt-prefix-alist
 
 Otherwise, the default prompt prefix will make successive prompts sibling headings, and therefore on different conversation branches, which probably isn't what you want.
 
-Note: using this option requires Org 9.6.7 or higher to be available.  The [[https://github.com/ultronozm/ai-org-chat.el][ai-org-chat]] package uses gptel to provide this branching conversation behavior for older versions of Org.
+Note: using this option requires Org 9.7 or higher to be available.  The [[https://github.com/ultronozm/ai-org-chat.el][ai-org-chat]] package uses gptel to provide this branching conversation behavior for older versions of Org.
   
 ***** Save gptel parameters to Org headings (reproducible chats)
 
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -231,7 +231,7 @@ value of `gptel-org-branching-context', which see."
                     (gptel--parse-buffer gptel-backend max-entries)))))
           (display-warning
              '(gptel org)
-             "Using `gptel-org-branching-context' requires Org version 9.6.7 or higher, it will be ignored.")
+             "Using `gptel-org-branching-context' requires Org version 9.7 or higher, it will be ignored.")
           (gptel--parse-buffer gptel-backend max-entries))
       ;; Create prompt the usual way
       (gptel--parse-buffer gptel-backend max-entries))))


karthink:
@grettke I think the exact Org commit we need is from before 9.7, but it has been a real pain to track down the nearest version.  So let's just go with 9.7.

karthink:
Thanks for the PR!

grettke:
@karthink Here was my research and thought process behind this pull request:

In the org-mode `main` branch `org-element-lineage-map` gets added in this commit `6333a20aa5b3b69dd8c883ee51b28ff24a0f06cf` on 2023-05-16.

`git show 6333a20aa5b3b69dd8c883ee51b28ff24a0f06cf`

```
commit 6333a20aa5b3b69dd8c883ee51b28ff24a0f06cf
Author: Ihor Radchenko <yantar92@posteo.net>
Date:   Tue May 16 13:32:40 2023 +0200

    org-element-ast: New function `org-element-lineage-map'

diff --git a/lisp/org-element-ast.el b/lisp/org-element-ast.el
index 6a3ae5e43..a98e879c1 100644
--- a/lisp/org-element-ast.el
+++ b/lisp/org-element-ast.el
@@ -1018,6 +1018,40 @@ when DATUM belongs to a full parse tree."
       (setq up (org-element-property :parent up)))
     (if types up (nreverse ancestors))))
 
+(defun org-element-lineage-map (datum fun &optional types with-self first-match)
+  "Map FUN across ancestors of DATUM, from closest to furthest.
+Return a list of results.  Nil values returned from FUN do not appear
+in the results.
```

Here is a list of the tags and the date ranges under our concern that I manually pruned:

`git log --tags --simplify-by-decoration --pretty="format:%ai %d"`

```
2024-06-01 10:59:57 +0200  (tag: release_9.7)
2024-05-10 10:30:48 +0300  (tag: release_9.6.30)
2023-12-31 12:58:57 +0100  (tag: release_9.6.15
```

Checking for the `org-element-lineage-map` function in 9.6.15 with rg (ripgrep)

```
cd ~/tmp
rm -rf org-mode
git clone https://git.savannah.gnu.org/git/emacs/org-mode.git
cd org-mode
git fetch && git fetch --tags
git checkout release_9.6.15 -b local_9.6.15
make autoloads
rg org-element-lineage-map
```

Produces no results: the function isn't in there.

Follow the same process for release_9.7 and it yields the results explaining, as you noted in the other ticket, that `org-element-lineage-map` is officially released.

```
etc/ORG-NEWS
416:**** New function ~org-element-lineage-map~

lisp/org-element-ast.el
964:(defun org-element-lineage-map (datum fun &optional types with-self first-match)

lisp/org.el
182:(declare-function org-element-lineage-map "org-element-ast"
13166:    (org-element-lineage-map
21307:          (org-element-lineage-map

lisp/ox-ascii.el
634:  (or (org-element-lineage-map

testing/lisp/test-org-element.el
4874:  "Test `org-element-lineage-map' specifications."
4879:	    (org-element-lineage-map (org-element-context) #'org-element-type))))
4885:	    (org-element-lineage-map (org-element-context) #'org-element-type nil t))))
4891:	    (org-element-lineage-map
4899:	    (org-element-lineage-map
```

Based on this I observed that

- All stable release Emacs v29.4-1 users are on Org v9.6.15 from 2023-12-31 because they come with Emacs
- `org-element-lineage-map` gets added on 2023-05-16 so stable Emacs v29 user will not have it
- Org v9.7 definitely has the function
- The next stable release of Emacs v30 will contain Org v9.7 and `org-element-lineage-map`
- User running Org from source (via Git or MELPA for example) are already using Org v9.7 so they have the function.

Therefore it is justified to bump up the version notice because

- Stable Emacs users, who run Org from the Emacs distribution, can never use this feature, so it is better to announce it
- Unstable Emacs users, or Stable users who run Org via Git will never get this warning, because they are on Org current
- We don't have to track down the release in which this feature was added because the two previous conclusions negate the need to do so

Consequently this pull request makes the version change required to **gracefully**  handle users on Emacs stable and built in Org Stable, Emacs Stable and Org Current, Emacs Unstable with Org-Unstable built in, Emacs Unstable with Org Unstable. 

It is graceful of course because you already handled it gracefully!  :)

grettke:
@karthink Here is how I found the commit.

Approach 1 searches the commit message which is helpful

`git log --oneline --regexp-ignore-case --grep='org-element-lineage-map' `

produces

```
69383dfc2 org-ascii--current-justification: Use `org-element-lineage-map'
6333a20aa org-element-ast: New function `org-element-lineage-map'
```

Approach 2 is guaranteed however

`git log --pickaxe-all --pickaxe-regex -S'org-element-lineage-map'`

It has a lot of results so some digging may be required.

grettke:
@karthink Since you threw down the gauntlet for me to find the release 🤣 it looks like it was in v9.7

```
grant@donutron:~/ghb/org-mode> git --no-pager diff release_9.6.16..release_9.6.30 | grep 'org-element-lineage-map'
OK...

grant@donutron:~/ghb/org-mode> git --no-pager diff release_9.6.16..release_9.7 | grep 'org-element-lineage-map'
OK...

+**** New function ~org-element-lineage-map~
+(defun org-element-lineage-map (datum fun &optional types with-self first-match)
+(declare-function org-element-lineage-map "org-element-ast"
+    (org-element-lineage-map
+          (org-element-lineage-map
+  (or (org-element-lineage-map
+  "Test `org-element-lineage-map' specifications."
+	    (org-element-lineage-map (org-element-context) #'org-element-type))))
+	    (org-element-lineage-map (org-element-context) #'org-element-type nil t))))
+	    (org-element-lineage-map
+	    (org-element-lineage-map
```

PR #603, opened 2025-02-01
Update details for openai o models

tillydray:
From the docs https://platform.openai.com/docs/models#o3-mini

> o3-mini is our most recent small reasoning model, providing high
> intelligence **at the same cost and latency targets of o1-mini.** o3-mini
> also supports key developer features, like Structured Outputs, function
> calling, Batch API, and more. Like other models in the o-series, it is
> designed to excel at science, math, and coding tasks.

> The o1 reasoning model is designed to solve hard problems across
> domains. o1-mini is a faster and more affordable reasoning model, but we
> **recommend using the newer o3-mini model** that features higher
> intelligence at the **same latency and price as o1-mini.**

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -586,7 +586,7 @@ To set the temperature for a chat session interactively call
      :input-cost 3
      :output-cost 12
      :cutoff-date "2023-10"
-     :capabilities (nosystem reasoning)
+     :capabilities (structured-outputs tool-use json url function-calling developer-message)
      :request-params (:developer :reasoning_effort))
     ;; limited information available
     (gpt-4-32k
--- a/gptel.el
+++ b/gptel.el
@@ -574,7 +574,7 @@ To set the temperature for a chat session interactively call
      :cutoff-date "2023-10"
      :capabilities (nosystem reasoning))
     (o1-mini
-     :description "Faster and cheaper reasoning model good at coding, math, and science"
+     :description "DEPRECATED: PLEASE USE o3-mini"
      :context-window 128
      :input-cost 3
      :output-cost 12
--- a/gptel.el
+++ b/gptel.el
@@ -563,8 +563,7 @@ To set the temperature for a chat session interactively call
      :context-window 200
      :input-cost 15
      :output-cost 60
-     :cutoff-date "2023-10"
-     :request-params (:stream :json-false))
+     :cutoff-date "2023-10")
     (o1-preview
      :description "DEPRECATED: PLEASE USE o1"
      :capabilities (nosystem media)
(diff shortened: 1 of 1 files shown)


karthink:
o3-mini was added to the most of default OpenAI models yesterday.  Is there
more information in this PR than the current implementation? (On mobile
right now)

gptel.el
```diff
@@ -573,24 +572,22 @@ To set the temperature for a chat session interactively call
...
+     :capabilities (nosystem reasoning))
     (o3-mini
      :description "High intelligence at the same cost and latency targets of o1-mini"
      :context-window 200
      :input-cost 3
      :output-cost 12
      :cutoff-date "2023-10"
-     :capabilities (nosystem reasoning)
-     :request-params (:stream :json-false))
+     :capabilities (nosystem structured-outputs json url function-calling developer-message)
```
tillydray, line 589: I'm not sure what `json` and `url` mean here so I might be wrong, could use some guidance on what these mean

tillydray:
@karthink there is a small amount of detail changes in this PR that separates it from your commit from yesterday

karthink:
@tillydray I don't understand the purpose of most of these changes.  Could you explain
1. why you're removing `:request-params` for many models, and
2. changing the `:capabilities` to add symbols that are not recognized, like `developer-message` or `function-calling`?
3. What the `:developer :reasoning_effort` does in o3-mini's `:capabilities`?

gptel.el
karthink: - `json` is currently meaningless but reserved to indicate that a model can generate JSON output
- The `url` capability is for models that can fetch and include assets from urls in the context (on the server, not via tool calling in Emacs.)  The OpenAI models can do this and it is used by gptel's media support.

pabl0:
BTW, is there a reason why the models listed in definition of `gpt-model` are not in sync with `gptel--openai-models`? And why they are hard-coded and not generated from that list?

https://github.com/karthink/gptel/blob/6090687150e4781e93c1d736e2b8d1e6be60054f/gptel.el#L464

karthink:
> And why they are hard-coded and not generated from that list?

Oversight.  Would you like to make this change?

pabl0:
> > And why they are hard-coded and not generated from that list?
> 
> Oversight. Would you like to make this change?

Sure, gave it a shot at #606

tillydray:
This PR is mostly misunderstandings on my part

> 1. why you're removing `:request-params` for many models

I misunderstood what `:request-params` was doing, I'll undo these changes

> 2. changing the `:capabilities` to add symbols that are not recognized, like `developer-message` or `function-calling`?

`developer-message` has replaced `system-message` per https://platform.openai.com/docs/guides/text-generation#messages-and-roles, but when I wrote this PR I thought it was an additional capability.

> 3. What the `:developer :reasoning_effort` does in o3-mini's `:capabilities`?

per https://platform.openai.com/docs/api-reference/chat/create#chat-create-reasoning_effort

> Constrains effort on reasoning for [reasoning models](https://platform.openai.com/docs/guides/reasoning). Currently supported values are low, medium, and high. Reducing reasoning effort can result in faster responses and fewer tokens used on reasoning in a response.

karthink:
> This PR is mostly misunderstandings on my part

No problem!

> per https://platform.openai.com/docs/api-reference/chat/create#chat-create-reasoning_effort

I'm aware of this parameter -- my question was what the purpose of placing it in `:capabilities` was.  `:capabilities` is used by gptel to check if certain gptel features should be (dynamically) enabled/disabled in the payload construction and in the transient menus. They don't correspond to values we can set, like setting `reasoning_effort` to `"low"`, etc.  That happens in `:request-params`.

We need a general `reasoning` `:capability` because the reasoning-type models have extra behavior we want to be able to toggle, and I've added it for the OpenAI models.  But it does nothing yet, as gptel has no code for reasoning-model-specific behavior.

Issue #542, opened 2025-01-02
gptel agents or configuration presets


karthink:
> [!NOTE]
> This feature has been implemented.
>
> You can now define "presets", which are a bundle of gptel options, such as the backend, model, system message, included tools, temperature and so on.  This set of options can be applied together, making it easy to switch between different tasks using gptel.  From gptel's transient menu, you can save the current configuration as a preset or apply another one.  Presets can be applied globally, buffer-locally or for the next request only.  To persist presets across Emacs sessions, define presets in your configuration using `gptel-make-preset`.

Original feature request follows:

-------

> A suggestion when you do get to this, I find the system message one prefers often depends on the model that is selected. I find when having a dialogue, in which I switch models mid-chat, I also sometimes need to switch the system prompt to better match the new model.
 
This appears to be a need that many users have.  There are many discussions threads where people have asked how to implement a "presets" feature.  I think it might make sense to add a presets feature to the package which is a bundle of configuration you can switch to all at once.

``` emacs-lisp
(gptel-make-preset
 "coding-preset"
 :system "system message for coding here..."    ; can be a function, see gptel-directives
 :backend gptel--anthropic-backend              ; or name of backend, like "Claude"
 :model   'claude-3-sonnet-20240229
 :context 'gptel-context-lsp
 :tools  (list gptel-tool-1 gptel-tool-2 ...)   ; list of gptel tools to supply
 :callback nil)
```

You can then select a preset from the transient menu, which sets all of these options at once.  Via the "scope" switch in the menu, this preset can be set globally, in one buffer or just for the next request.

Some of these options/sources (like `gptel-context-lsp`) don't exist yet. `:callback` is basically a custom action you can specify instead of inserting the response, with nil being the default callback gptel uses.

In other LLM clients this is called an "agent", but really they're just a bundle of prompts+configuration.

_Originally posted by @karthink in https://github.com/karthink/gptel/issues/416#issuecomment-2567232641_

ovistoica:
Gptel presets would be very useful! 

I created a per-project side panel chat workflow inspired by Cursor, and changing the context for each project is not ideal. I would much appreciate a list of presets I could set (maybe with `.dir-locals`) for each project. 


References: 
- [ai-project-agent.el](https://github.com/ovistoica/emacs/blob/main/plugins/ai-project-agent.el)

Functionality 
```elisp
(use-package ai-project-agent
  :after (gptel flycheck)
[...]

PR #606, opened 2025-02-02
gptel: Make gptel-model variable definition dynamic

pabl0:
* gptel.el (gptel-model): Ensure the docstring documentation and customization options are current (in sync with
`gptel--openai-models') by dynamically generating them instead of hard-coding.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -461,38 +461,6 @@ call `gptel-send' with a prefix argument."
   :type '(choice (natnum :tag "Specify Token count")
                  (const :tag "Default" nil)))
 
-(defcustom gptel-model 'gpt-4o-mini
-  "GPT Model for chat.
-
-The name of the model, as a symbol.  This is the name as expected
-by the LLM provider's API.
-
-The current options for ChatGPT are
-- `gpt-3.5-turbo'
-- `gpt-3.5-turbo-16k'
-- `gpt-4o-mini'
-- `gpt-4'
-- `gpt-4o'
-- `gpt-4-turbo'
-- `gpt-4-turbo-preview'
-- `gpt-4-32k'
-- `gpt-4-1106-preview'
-
-To set the model for a chat session interactively call
-`gptel-send' with a prefix argument."
-  :safe #'always
-  :type '(choice
-          (symbol :tag "Specify model name")
-          (const :tag "GPT 4 omni mini" gpt-4o-mini)
-          (const :tag "GPT 3.5 turbo" gpt-3.5-turbo)
-          (const :tag "GPT 3.5 turbo 16k" gpt-3.5-turbo-16k)
-          (const :tag "GPT 4" gpt-4)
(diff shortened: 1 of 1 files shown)


karthink:
Thanks!  Now that I look at the code, this list is ChatGPT specific and alternatives like Claude are very popular now (as opposed to an year ago), so I'm thinking we don't even need to "hard-code" the options in the docstring.  We can just say that `gptel-model` should be the name of a model as a symbol, and continue to provide the models for the OpenAI backend as available options since it's the default.

What do you think?

psionic-k:
This looks like an improvement.  I would question having a hard-coded default value.  Can't we just decide the default based on the first one a configuration is found for?  That might need to happen downstream.  This variable would be nil in most cases and could be set / bound only when there was more than one option and the default wasn't already first.

The way things are going, we're rapidly going to wind up with several models available being the norm.  Some for completions, some for translation, some for specific languages, some local, some remote.  I don't expect this variable to have meaning in that direction.

karthink:
> Can't we just decide the default based on the first one a configuration is found for? That might need to happen downstream. This variable would be nil in most cases and could be set / bound only when there was more than one option and the default wasn't already first.

The user is expected to set `gptel-model` in their configuration.  If they don't, it defaults to `gpt-4o-mini`.  How is the current behavior different from what you describe?

> The way things are going, we're rapidly going to wind up with several models available being the norm. Some for completions, some for translation, some for specific languages, some local, some remote. I don't expect this variable to have meaning in that direction.

- There's value in simplicity.  Users like gptel because it's simple to use.
- Providing a `gptel-translation-model`, `gptel-completion-model` and so on doesn't scale and imposes constraints from the top.
- It won't stop there, because a translation model will also need a translation system message, and so on.
- Users can use something like the presets system (#542) to define everything needed for an LLM task in one place, and switch to this preset buffer-locally or for the next request.

pabl0:
I agree that listing the models in the docstring is unnecessary, so I'll drop that part. However, we might consider including the available models from the default OpenAI backend or the currently active backend. I'm uncertain option would be the best approach for listing them, and how to get the list of models of the current backend.

karthink:
This looks good to me.

> However, we might consider including the available models from the default OpenAI backend or the currently active backend.

While you can do something like this to use the currently active backend:

``` emacs-lisp
:type `(choice
	(symbol :tag "Specify model name")
	,@(mapcar (lambda (model)
		    (list 'const :tag (symbol-name (car model))
		     (car model)))
```

it won't actually work because the `defcustom` is not dynamically evaluated.  So changing the active backend will not cause the list of choices to be updated, and it will always be what it was set to at load-time, which is just `gptel--openai-models`.

karthink:
Can I go ahead and merge this now?

pabl0:
Yes I think it's good enough, thanks.

karthink:
@pabl0 Thanks for the PR!

PR #608, opened 2025-02-03
gptel: Use json-mode to inspect query if available

pabl0:
* gptel.el (gptel--inspect-query): Use json-mode when inspecting the query (dry run), if json-mode is available.


karthink:
@pabl0 Thanks for the PR.  I use `fundamental-mode` intentionally because both `json-mode` and the built-in `js-json-mode`  suck when long lines are involved, slowing Emacs down to a crawl.  And because long chunks of the buffer end up as a single line in the encoded JSON, this is _always_ the case.  (`so-long` mode does nothing to help here.)

In contrast, `lisp-data-mode`, which is used by the lisp-style query inspector, is much better at handling long lines.

If you wish to use `json-mode` you can advise the function in your configuration.

pabl0:
Thanks for feedback, @karthink. You're right. My previous tests were limited to small contexts, handwritten prompts, and brief LLM responses, so I didn't notice any performance issues.

However, when I added the entire gptel directory to the context, including byte-compiled files, it significantly slowed down, particularly in Emacs 28. With version 29, it feels much faster even on a small tablet—almost on par with fundamental mode (both json-mode and js-json-mode).

I wasn't aware of js-json-mode (introduced in Emacs 29). Perhaps we should prioritize that if available, unless we decide to abandon the idea entirely.

karthink:
You don't need an entire directory -- you can add a single 1000+ line file and observe this problem.

> With version 29, it feels much faster even on a small tablet—almost on par with fundamental mode (both json-mode and js-json-mode).

Displaying long lines is intrinsically slow so it depends on your hardware.  gptel supports Emacs 27+ and I test it on a low-end 2012 Thinkpad, and turning on `json-mode` or `js-json-mode` is a non-starter.

Another reason the lisp-mode preview works better is because it represents newlines as actual newlines, not `\n`.  So you never actually run into the long-lines problem unless the original file(s) themselves have long lines.

Closing as not planned.

PR #615, opened 2025-02-06
gptel-transient: Improve parsing of crowdsourced prompts CSV file

### LINKED ISSUE #614: Parsing crowdsourced prompts leads to completion errors

### Bug Description

gptel parses csv naively using `read` assuming that each field is always surround by `"`:

https://github.com/karthink/gptel/blob/aa649b013322a62d29e7507e22234f8000c68c48/gptel-transient.el#L260

But in the latest prompts file, there are lines which go against this assumption, for example, this line:

https://github.com/f/awesome-chatgpt-prompts/blob/3a1d66c8ca083ca340df11ed4c46e765181c50ae/prompts.csv?plain=1#L118

So that leads to reading in symbol `I`, instead of the full prompt, which results in completion error:
```
(wrong-type-argument stringp I)
```

### Backend

None

### Steps to Reproduce

1. Delete the local cache file at `~/.cache/gptel-crowdsourced-prompts.csv`
2. Reset the variable `gptel--crowdsourced-prompts` to empty hashtable.
3. In gptel menu, `set system message` `Pick crowdsourced prompt`

### Additional Context

Irrelevant

### Backtrace

```emacs-lisp
I didn't get the backtrace maybe because the error is in a hook?
```

### Log Information

```json

```

pabl0:
* gptel-transient.el (gptel--crowdsourced-prompts): Improve parsing by supporting CSV fields that are not quoted.

This should fix #614

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -215,6 +215,22 @@ OBJ is a tool-infix of type `gptel--switch'."
   (make-hash-table :test #'equal)
   "Crowdsourced LLM system prompts.")
 
+(defun gptel--read-csv-column ()
+  "Read next CSV column in the current buffer.
+
+Supports both quoted and non-quoted columns (RFC 4180)."
+  (let ((start (point)))
+    (unless (eolp)
+      (let ((column
+	     (if (eq (char-after) ?\")
+		 (when (re-search-forward "\",\\|\"$" nil t)
+		   (let ((end (match-beginning 0)))
+		     (buffer-substring-no-properties (+ start 1) (if (eolp) (- end 1) end))))
+	       (when (search-forward "," (line-end-position) t)
+		 (let ((end (match-beginning 0)))
+		   (buffer-substring-no-properties start end))))))
+	(string-replace "\"\"" "\"" column)))))
+
 (defun gptel--crowdsourced-prompts ()
   "Acquire and read crowdsourced LLM system prompts.
 
@@ -252,14 +268,10 @@ which see."
           (goto-char (point-min))
           (forward-line 1)
           (while (not (eobp))
-            (when-let ((act (read (current-buffer))))
-              (forward-char)
-              (save-excursion
-                (while (re-search-forward "\"\"" (line-end-position) t)
-                  (replace-match "\\\\\"")))
-              (when-let ((prompt (read (current-buffer))))
-                (puthash act prompt gptel--crowdsourced-prompts)))
-            (forward-line 1)))))
+	    (when-let* ((act (gptel--read-csv-column))
+			(prompt (gptel--read-csv-column)))
+		(puthash act prompt gptel--crowdsourced-prompts))
+	      (forward-line 1)))))
     gptel--crowdsourced-prompts))
 
 


gptel-transient.el
```diff
@@ -247,18 +247,23 @@ which see."
...
-                (while (re-search-forward "\"\"" (line-end-position) t)
-                  (replace-match "\\\\\"")))
-              (when-let ((prompt (read (current-buffer))))
-                (puthash act prompt gptel--crowdsourced-prompts)))
+	(with-temp-buffer
+	  (insert-file-contents gptel-crowdsourced-prompts-file)
+	  (goto-char (point-min))
+	  (forward-line 1)
+	  (while (not (eobp))
+            (let ((line (thing-at-point 'line t)))
```
karthink, line 255: Avoid `thing-at-point`, we don't want to load `thingatpt.el` unless it's required.  You can get the line via

```elisp
(buffer-substring-no-properties (line-beginning-position) (line-end-position))
```

gptel-transient.el
```diff
@@ -247,18 +247,23 @@ which see."
...
+              (when line
+		;; Use a regex to split the line while respecting quoted strings
+		(let ((fields (split-string line ",\"?")))
```
karthink, line 258: what happens if there's `,"` inside the prompt?

karthink: You can do this by parsing the buffer instead of splitting the line into a string.  `search-forward` to the next unescaped `,` and collect the text from the start of search until here.

pabl0:
OK, another attempt. This should work support ," and \," use cases, as well as old csv files that did not have the last `for_devs` column, and mixed quoted / non-quoted columns. But not sure about the corner case where a non-quoted column would have "" in it, is that supported by RFC4180?

gptel-transient.el
```diff
@@ -252,14 +269,10 @@ which see."
...
           (while (not (eobp))
-            (when-let ((act (read (current-buffer))))
-              (forward-char)
-              (save-excursion
-                (while (re-search-forward "\"\"" (line-end-position) t)
-                  (replace-match "\\\\\"")))
-              (when-let ((prompt (read (current-buffer))))
-                (puthash act prompt gptel--crowdsourced-prompts)))
-            (forward-line 1)))))
+	    (when-let ((act (gptel--read-csv-column)))
```
karthink, line 272: Use a single `when-let*` or `and-let*` instead of two `when-let` clauses.  (Also, `when-let` is deprecated in favor of `when-let*` in upcoming Emacs versions)

gptel-transient.el
```diff
@@ -215,6 +215,23 @@ OBJ is a tool-infix of type `gptel--switch'."
...
+(defun gptel--read-csv-column ()
+  "Read next CSV column in the current buffer.
+
+Supports both quoted and non-quoted columns, but not columns
+spanning multiple lines."
+  (let ((start (point)))
+    (unless (eolp)
+      (let ((column
+	     (if (eq (char-after) ?\")
+		 (when (re-search-forward "\",\\|\"$" nil t)
```
karthink, line 227: Regexp search is unnecessary, I think?  You can run `(search-forward "\"," (line-end-position) t)` and `(search-forward "," (line-end-position) t)` instead.  

Do you also need to check that the comma itself is not escaped, like here?
```csv
column 1 is here, column 2\, which includes a comma, column 3 here
```

pabl0: > Regexp search is unnecessary, I think? You can run `(search-forward "\"," (line-end-position) t)` and `(search-forward "," (line-end-position) t)` instead.

OK with the non-quoted case, I think the first one needs to be regexp since we want to match "$.
 
> Do you also need to check that the comma itself is not escaped, like here?
> 
> ```
> column 1 is here, column 2\, which includes a comma, column 3 here
> ```

The awesome prompts repo does not specify which CSV standard they support, but they do run commits against csvclean to check the format and I don't think this type of quoting is supported there. RFC4180 is the only formal CSV standard I know of and it does not support \ quoting AFAIK. I don't think we need to worry about this corner case too much.

pabl0:
BTW this is the parser that the awesome-chatgpt-prompts project use to parse csv:

https://github.com/f/awesome-chatgpt-prompts/blob/3a1d66c8ca083ca340df11ed4c46e765181c50ae/script.js#L166

The [web page prompts.chat](https::/prompts.chat) includes metadata like the name/GH handle of the the contributor of the quote. That data is not in the csv file. Turns out that when contributing a PR to add a new prompt, you're supposed to copy the same data to both README.md and prompts.csv in slightly different format, and the web page is somehow generated from the markdown file (but for some reason the csv is also parsed? :confused:). 

Also when looking into this I also noticed that there are a couple of duplicate entries in the CSV file, and we're using a hash table expecting the prompt names to be unique. The actual prompts are of different. I'm guessing it's just a mistake since the web page finds both entries using the search, but can actually show just the other.

Duplicate prompts are currently:

- Chess Player
- Life Coach
- Prompt Generator
- Python Interpreter

Anyway, this PR works fine for me. Do you think we could go ahead and merge soon even it is not perfect, since the selecting is currently broken for many users and as time passes it will affect more folks since after two weeks a new version is downloaded into the cache. (Planning implementing a better logic with If-Modified-Since if you approve the idea?)

karthink:
Thanks @pabl0!

Do you use the crowdsourced prompts, BTW?  It seemed like a good idea at the time, but I found them pretty much useless afterward a few weeks and never looked at them again.  In general my system messages tend to be very specific to tasks.

pabl0:
> Do you use the crowdsourced prompts, BTW? It seemed like a good idea at the time, but I found them pretty much useless afterward a few weeks and never looked at them again. In general my system messages tend to be very specific to tasks.

Well.. Not much. I don't think they are of very high quality in general, but maybe they are useful as inspiration and something to get started with. (I often find it surprisingly difficult to come up with anything...) Also the user interface does not really scale, especially when working on a tablet computer with very limited screen real estate. I was thinking of submitting PR where the selection would only show the names, not the actual prompts (or perhaps show them with some key press?) Tested the idea a bit and it's much more readable that way, but obviously what you get is bit of a surprise, What do you think?

I'm also wondering if these kind of prompt dataset will get more popular in the future and there would be more sources available. And most likely there will be no agreement on the format, so we'll see JSON and whatever happens to be the cool format of the day (since the kids don't know about s-expressions :smile:).

karthink:
> Also the user interface does not really scale, especially when working on a tablet computer with very limited screen real estate. I was thinking of submitting PR where the selection would only show the names, not the actual prompts (or perhaps show them with some key press?) Tested the idea a bit and it's much more readable that way, but obviously what you get is bit of a surprise, What do you think?

Do you mean the user interface for picking a crowd-sourced prompt?  That's just completing-read, so it should scale to thousands of entries?

If you mean the user interface for picking a system message from gptel-directives, then yes, it doesn't scale past ~25 or so.  So far no one's complained about it, so I'm not inclined to mess with it yet.

> I'm also wondering if these kind of prompt dataset will get more popular in the future and there would be more sources available. And most likely there will be no agreement on the format, so we'll see JSON and whatever happens to be the cool format of the day (since the kids don't know about s-expressions :smile:).

I think what we'll see is a templating system for prompts, where you can combine various predefined elements (tone, verbosity, text format, do's/don'ts) to create a system message draft on demand.  Something like this must probably exist already.

pabl0:
> Do you mean the user interface for picking a crowd-sourced prompt? That's just completing-read, so it should scale to thousands of entries?

I mean scaling in the sense that when you use the tab completion, the screen gets super busy (the whole prompt is shown wrapped into multiple lines), and since the whole prompt is displayed, you cannot fit that many in the screen. Hope this makes sense (sorry English is my third language).

> I think what we'll see is a templating system for prompts, where you can combine various predefined elements (tone, verbosity, text format, do's/don'ts) to create a system message draft on demand. Something like this must probably exist already.

How cool would that be, I want that in gptel :100:

PR #616, opened 2025-02-06
gptel-transient: Improve directive/system message editing

pabl0:
I find no value in activating the directive in the edit buffer as region, and it can be quite annoying.

Additionally, since the directives and system messages are primarily regular text, enabling visual-line-mode (word wrapping) seems logical.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -1364,7 +1364,8 @@ it is run after exiting the edit."
           ;; If it's a list, insert only the system message part
           (insert (car-safe (gptel--parse-directive directive 'raw)))
           (push-mark nil 'nomsg))
-        (activate-mark))
+        (activate-mark)
+	(visual-line-mode 1))
       (display-buffer (current-buffer)
                       `((display-buffer-below-selected)
                         (body-function . ,#'select-window)


karthink:
> I find no value in activating the directive in the edit buffer as region, and it can be quite annoying.

I find it quite useful.  If you want to edit a prompt instead, all you need to do is press `C-g`.  In contrast, if you want to set a custom directive and we turn off activating the region, you have to select the current prompt in this buffer, delete it and start typing, which is much more work.  The only way to set a custom directive interactively in gptel is to use this edit buffer.  Note that you can't do `C-x h` to mark the whole buffer (and delete text) as this selects the read-only text at the top too.

So on balance, activating the region works better.

> enabling visual-line-mode (word wrapping) seems logical.

Yes, this is a good idea.

pabl0:
> I find it quite useful. If you want to edit the existing prompt, all you need to do is press `C-g`. In contrast, if you want to write a custom directive and we turn off activating the region, you have to select the current prompt in this buffer first and delete it before you start typing, which is much more work. The only way to set a custom directive interactively in gptel is to use this edit buffer. Note that you can't do `C-x h` to mark the whole buffer (and delete text) as this selects the read-only text at the top too.

Yes I think you're right, and perhaps this is the most common use case. On the other hand, when picking a crowdsourced prompt, I don't think the user would want to delete everything to edit the prompt. Perhaps we could make setting the region optional depending on a parameter,  and call it differently from `gptel--read-crowdsourced-prompt`?

karthink:
> On the other hand, when picking a crowdsourced prompt, I don't think the user would want to delete everything to edit the prompt. Perhaps we could make setting the region optional depending on a parameter,  and call it differently from `gptel--read-crowdsourced-prompt`?

Done.

I can merge the visual-line-mode part of this PR if you undo the changes to the mark status.

karthink:
Cheers @pabl0!

PR #618, opened 2025-02-07
gptel-transient: (really) unmark before editing crowdsourced prompt

pabl0:
* gptel-transient.el (gptel--read-crowdsourced-prompt, gptel--edit-directive, gptel--suffix-system-message): Deactivate the mark after entering the prompt-edit buffer for a crowdsourced prompt (#616). This behaviour was introduced in commit cc53a81c, but it did not actually work since `deactivate-mark' was called after returning from `gptel--read-crowdsourced-prompt', when it was too late.

(Also a slight indentation adjusment, @karthink are you editing elisp with some non-standard indentation mode?)

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -587,7 +587,7 @@ generated from functions."
                 "Rewrite directive is dynamically generated: Edit its current value instead?")))))
   (if cancel (progn (message "Edit canceled")
                     (call-interactively #'gptel-rewrite))
-    (gptel--edit-directive 'gptel--rewrite-directive #'gptel-rewrite)))
+    (gptel--edit-directive 'gptel--rewrite-directive #'gptel-rewrite t)))
 
 (transient-define-suffix gptel--suffix-rewrite (&optional rewrite-message dry-run)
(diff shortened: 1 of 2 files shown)


karthink:
There is one more use of `gptel--edit-directive` in `gptel-rewrite.el`.  Please modify that call too.

pabl0:
> There is one more use of `gptel--edit-directive` in `gptel-rewrite.el`. Please modify that call too.

Good catch, will fix, thx.

karthink:
Thanks @pabl0!  Also thanks for adhering to gptel's never-formally-specified commit message format.  Did you write it that way for gptel specifically, or do you use a similar format elsewhere?

pabl0:
> Thanks @pabl0! Also thanks for adhering to gptel's never-formally-specified commit message format. Did you write it that way for gptel specifically, or do you use a similar format elsewhere?

😎 Yes, generally it's best to follow the commit message style used by a project's main authors. This format appears to be similar to that used by the GNU Emacs developers. (Which I believe is still used to generate the old-school ChangeLog when a new version is released, omitting some trivial changes.)

PR #620, opened 2025-02-08
gptel-transient: ensure point is after prompt

LemonBreezes:
Basically, this code changes this:
image
to:
image

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -172,7 +172,8 @@ will toggle its visibility state."
           (_ (overlay-put ov 'gptel 'partial)
              (funcall update ov (truncate-string-to-width
                                  prefix max-width nil nil
-                                 'ellipsis))))))))
+                                 'ellipsis)))))
+      (goto-char (point-max)))))
 
 (defun gptel--transient-read-variable (prompt initial-input history)
   "Read value from minibuffer and interpret the result as a Lisp object.


karthink:
The reason I don't do this by default is because I want the user to (i) realize that the `Rewrite: ` text is not required, i.e. not part of some pre-designed prompt and they can type anything, and (ii) it's easier to `C-k` or `C-e` and start typing than having to `C-0 k` from the end.  (Most users also don't know about `C-0 k` or how to kill to the beginning of a line, and don't use evil-mode in the minibuffer either.)

LemonBreezes:
> The reason I don't do this by default is because
> 1. I want the user to realize that the `Rewrite: ` text is not required, i.e. not part of some pre-designed prompt they need to respect, so they can type anything, and
> 2. It's easier to `C-k` or `C-e` and start typing than having to `C-0 k` from the end.  (Most users also don't know about `C-0 k` or how to kill to the beginning of a line, and don't use evil-mode in the minibuffer either.)

It seems kind of ridiculous to me that we start the point on top of a prompt. In that case, maybe we should change the prompt to just "> " like in Comint  and move the point to the end.

LemonBreezes:
Okay, I have changed my mind about this PR. I think we should remove the overlay altogether. It is extremely jarring to have an overlay in the minibuffer regardless of where the point is.

PR #621, opened 2025-02-08
feat: add gptel buffer management package

bhutley:
Provides a tabulated buffer list for managing gptel buffers with ability to visit, mark for deletion, and kill gptel buffers interactively.

### DIFF
--- a/gptel-list-buffers.el
+++ b/gptel-list-buffers.el
@@ -1,0 +1,165 @@
+;;; gptel-list-buffers.el --- Provide a buffer list of gptel buffers
+
+;; Copyright (C) 2023  Karthik Chikmagalur
+
+;; Author: Brett Hutley <brett@hutley.net>
+;; Keywords: buffer-list
+
+;; SPDX-License-Identifier: GPL-3.0-or-later
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; This creates a buffer containing a list of gptel buffers, with the
+;; ability to kill or go to a buffer via keystrokes.
+;;
+;; Kinda based on the functionality of the inbuilt `list-buffers`
+;; function
+
+;; Keystrokes when buffer is active:
+;;
+;;  d - Mark buffer for deletion
+;;  f - Visit the selected buffer
+;;  x - Delete all marked buffers
+;;  q - Quit the window
+
+;;; Code:
+
+;;; -*- lexical-binding: t -*-
+(require 'gptel)
+
+(defvar gptel-buffer-list--marked-buffers
+  (make-hash-table :test 'equal)
+  "Hash table of buffers (by name) currently marked for deletion.")
+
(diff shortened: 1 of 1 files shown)


bhutley:
This was just to scratch my own itch. I end up having a lot of gptel buffers around, some of them quite old, so I wanted an easy way to list and delete them.

gptel-list-buffers.el
```diff
@@ -0,0 +1,165 @@
+;;; gptel-list-buffers.el --- Provide a buffer list of gptel buffers
+
+;; Copyright (C) 2023  Karthik Chikmagalur
+
+;; Author: Brett Hutley <brett@hutley.net>
+;; Keywords: buffer-list
```
pabl0, line 6: elisp info documentation says:
 The ‘Keywords’ header should contain at least one standard keyword from the ‘finder-known-keywords’ list.

I don't think `buffer-list`is a standard keyword.

gptel-list-buffers.el
```diff
@@ -0,0 +1,165 @@
+;;; gptel-list-buffers.el --- Provide a buffer list of gptel buffers
+
+;; Copyright (C) 2023  Karthik Chikmagalur
```
pabl0, line 3: 2025 for a new file?

karthink:
@bhutley Re: #624 

Yes, I think this is mostly orthogonal to gptel and can be installed as an add-on package by those who want this feature.

But taking a step back, I also think this kind of listing is very easy to create with ibuffer and doesn't require any extra features, package or code:

https://github.com/user-attachments/assets/0bbf689d-a661-472b-a12b-fb2b5113a678

1. `M-x ibuffer`
2. `M-x ibuffer-filter-by-predicate`, `gptel-mode`
That's it.  By default `ibuffer-filter-by-predicate` is bound to `/ e` in `ibuffer`.  (`/` for filters, `e` for "elisp")

Now you have the full power of ibuffer to do whatever you want with these buffers.  But I understand that users might not be aware that ibuffer can do this, so you probably should publish this as a package.

gptel-list-buffers.el
```diff
@@ -0,0 +1,165 @@
...
+  (define-key (current-local-map) (kbd "g") #'gptel-buffer-list-refresh)
+  (define-key (current-local-map) (kbd "q") #'quit-window)
+
+  (tabulated-list-init-header))
+
+(defun gptel-list-buffers ()
+  "Display a list of gptel buffers."
+  (interactive)
+  (let ((gptel-buffers
+         (seq-filter (lambda (buf)
```
karthink, line 85: You can use `match-buffers` to do this more simply.

gptel-list-buffers.el
```diff
@@ -0,0 +1,165 @@
...
+;;  x - Delete all marked buffers
+;;  q - Quit the window
+
+;;; Code:
+
+;;; -*- lexical-binding: t -*-
+(require 'gptel)
+
+(defvar gptel-buffer-list--marked-buffers
+  (make-hash-table :test 'equal)
```
karthink, line 44: A hash table is overkill (and inefficient) for < 64 entries, a simple list or alist works better.  Probably too small to matter either way though.

bhutley:
@karthink and @pabl0 Thanks for the code review! As per [@karthink's comment](https://github.com/karthink/gptel/pull/621#issuecomment-2645989237) , I'll close this PR and publish this as a separate package for people to load if they want.

karthink:
Here's all the code you need to get for ibuffer:

```elisp
(define-ibuffer-column backend
  (:name "Backend" :inline t)
  (and-let* ((backend (buffer-local-value 'gptel-backend buffer)))
    (gptel-backend-name backend)))

(define-ibuffer-column model
  (:name "Model" :inline t)
  (and-let* ((model (buffer-local-value 'gptel-model buffer)))
    (gptel--model-name model)))

(add-to-list
 'ibuffer-formats
 '(mark modified read-only locked
   " " (name 18 18 :left :elide)
   " " (backend 12 20 :left :elide)
   " " (model 30 30 :left :elide)
   " " (mode 16 16 :left :elide)
   " " filename-and-process))
```

Result:
screenshot_20250209T002848

bhutley:
Nice! Thanks @karthink

PR #622, opened 2025-02-08
Fix read-only behaviour in gptel--edit-directive

pabl0:
* gptel-transient.el (gptel--edit-directive): `make-separator-line' is a function introduced in Emacs 28.1, so it was disabled with a comment.  However, we require compat-29.1.4.1 making the function available in Emacs ~~28~~ EDIT: 27, and it is already used in other places, so enable it here as well.  Drop the extra newline from the examples above.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -1361,10 +1361,8 @@ setting up the buffer."
               "# Example: You are a poet. Reply only in verse."))
          "\n\n")
         (add-text-properties
-         (point-min) (1- (point))
-         (list 'read-only t 'face 'font-lock-comment-face))
-        ;; TODO: make-separator-line requires Emacs 28.1+.
-        ;; (insert (propertize (make-separator-line) 'rear-nonsticky t))
+         (point-min) (point)
+         (list 'read-only t 'face 'font-lock-comment-face 'front-sticky t 'rear-nonsticky t))
         (set-marker msg-start (point))
         (save-excursion
           ;; If it's a list, insert only the system message part


karthink:
@pabl0 Thanks, I think this hasn't been updated since I added compat as a dependency.

1. You need to adjust the `read-only` text-property to cover the separator line.
2. There are some issues with how `make-separator-line`  and newlines interact.  Different UI toolkits (lucid, gtk, NS, NT) handle it differently, leading to issues with its appearance.  You'll have to account for this.

This is how it appears in the lucid toolkit:
screenshot_20250208T173333

Observe that the separator line is under a blank line under the `Example: ...` text.  If you use gtk Emacs, please compare with how it appears for you.

Ideally we want the separator line right under the `Example: ...` lines, followed by a newline and then the text to be edited, which is the other way around.

-----

Sorry, nothing is ever a one-line change in gptel.  Too many subtle issues.

pabl0:
> @pabl0 Thanks, I think this hasn't been updated since I added compat as a dependency.
> 
>     1. You need to adjust the `read-only` text-property to cover the separator line.

Ah makes sense, I'll have a look at that.

>     2. There are some issues with how `make-separator-line`  and newlines interact.  Different UI toolkits (lucid, gtk, NS, NT) handle it differently, leading to issues with its appearance.  You'll have to account for this.

> Observe that the separator line is under a blank line under the `Example: ...` text. If you use gtk Emacs, please compare with how it appears for you.
> 
> Ideally we want the separator line right under the `Example: ...` lines, followed by a newline and then the text to be edited, which is the other way around. And we want this uniformly in all UI toolkits.
> 
> Sorry, nothing is ever a one-line change in gptel. Too many subtle issues.

Oh boy! I am indeed using gtk and it looks perfectly fine:

gptel-separator

I wonder about non-unix platforms like Windows?

pabl0:
Holy smokes, this separator line thing turned out to be a real rabbit hole.

The function itself is very simple – it basically returns just a newline (\n) with the :extend property and a face that defines that it's height is 0.1 and whatever color. But it acts rather inconsistently when it's near other newlines. I really can't figure out why.

I think I'm giving up – I don't think there is much added value in such line.

However I'd like to fix the read-onlynes of the comment to behave correctly with the following, what do you think? Now you can go to (point-min) and insert text before the comment (really a minor detail) and you can delete the empty line between the comment and the text to be edited. Again, not important, but attention to detail matters.

```diff
@@ -1356,8 +1356,8 @@ activate the mark."
          "# Example: Reply only with shell commands and no prose.\n"
          "# Example: You are a poet. Reply only in verse.\n\n")
         (add-text-properties
-         (point-min) (1- (point))
-         (list 'read-only t 'face 'font-lock-comment-face))
+         (point-min) (point)
+         (list 'front-sticky t 'rear-nonsticky t 'read-only t 'face 'font-lock-comment-face))
         ;; TODO: make-separator-line requires Emacs 28.1+.
         ;; (insert (propertize (make-separator-line) 'rear-nonsticky t))
         (set-marker msg-start (point))
```

karthink:
This change looks good to me.  I changed the function a bit as it's going to be used in more places soon, so you'll have to resolve a merge conflict.  (Since yours is a one-line change it should be easy.)

karthink:
> But it acts rather inconsistently when it's near other newlines.  I really can't figure out why.

Since the result is toolkit-dependent, the answer lies somewhere deep inside the display engine, and I think it's a waste of our time to go digging without knowledge of the internals.  I'm content to only use `make-separator-line` where getting the spacing wrong isn't going to be an eyesore.

karthink:
Thanks @pabl0!

PR #631, opened 2025-02-10
gptel-transient: Improve "Scope" infix appearence

pabl0:
* gptel-transient (gptel--infix-variable-scope): Drop extra space from :format and `transient-inactive-argument' face.

(Unless there is some reason to have the word "Scope" greyed out... but how is it "inactive"?)

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -761,9 +761,9 @@ value of `gptel-use-context', set from here."
   :argument "scope"
   :variable 'gptel--set-buffer-locally
   :class 'gptel--scope
-  :format "  %k %d %v"
+  :format " %k %d %v"
   :key "="
-  :description (propertize "Scope" 'face 'transient-inactive-argument))
+  :description "Scope")
 
 (transient-define-infix gptel--infix-num-messages-to-send ()
   "Number of recent messages to send with each exchange.


karthink:
Ah, removing `:pad-keys` from the `Request Parameters` column definition might fix the alignment issue.

pabl0:
> Is it too confusing if I make the scope part of the heading?

I like the look, but it is definitely confusing. Especially since the key is `=` and the other columns have "<" and ">" for decoration, I think many users would not realize that you're supposed to hit the `=`key. 

I should've guessed that you've already been thinking about this without coming up with a superior approach. :smile_cat: 
Perhaps it is good as-is...

karthink:
> Especially since the key is = and the other columns have "<" and ">" for decoration, I think many users would not realize that you're supposed to hit the =key.

screenshot_20250210T202435

Any better now?

> I should've guessed that you've already been thinking about this without coming up with a superior approach. 😸
> Perhaps it is good as-is...

As mentioned in another PR, no fix to gptel is as obvious as it appears.  Emacs bugs, Transient's limitations, UI edge cases, the list of complications is long...

But I think I can fix the alignment issue here, if nothing else.

pabl0:
How about putting it above the heading? Not sure if the "<Prompt from" and ">Response to" headings should align vertically with "Request Parameters". Visually looks better, logically it could be confusing since the scope doesn't apply (?).

gptel-scope

I know this is a very minor detail, but if there's a screen in some interface you use many times each day, it doesn't hurt if it looks neat, or what?

pabl0:
(BTW, while you're at it, did you consider making `-i`and `-t`work in similar fashion as `-v` and `=`? And if so, why did abandon the idea, strings too long with at least `-i` but might work for "use tools"?)

karthink:
> (BTW, while you're at it, did you consider making -iand -twork in similar fashion as -v and =? And if so, why did abandon the idea, strings too long with at least -i but might work for "use tools"?)

It's a lot more work than it appears to be, at minimum it requires a new EIEIO class.  So I judged the menu to be adequate and moved on.

If you try to implement this keep in mind that, technical issues aside, splatting all the options onto the menu will make the menu busier and harder to parse.  That said I didn't try so I don't know for sure that if will make the visuals worse.

pabl0:
> If you try to implement this keep in mind that, technical issues aside, splatting all the choices for `-t` or `-i` onto the menu will make the menu busier and harder to parse. That said I didn't try so I don't know for sure that if will make the visuals worse.

Yes I was thinking exactly the same. A tradeoff between consistent and visually pleasing UX. Definitely not high priority,.

pabl0:
I think I'm happy with how this looks now after edf834a, closing.

PR #635, opened 2025-02-12
gptel: Prefer LLM terminology over GPT

pabl0:
GPT, or Generative Pre-trained Transformer, is a designation originally used by OpenAI to describe their language models. Other companies or researchers might develop models with similar architectures, but "GPT" primarily associates with OpenAI's series.

* gptel.el (gptel-model, gptel-mode): Use "LLM model" instead of "GPT model".
* README.md: Ditto.
* gptel-transient.el (gptel--infix-provider): Just "Model" instead of "GPT Model".  The meaning should be clear in this context, and leaves bit more space for potentially long model names.

### DIFF
--- a/README.org
+++ b/README.org
@@ -860,7 +860,7 @@ gptel provides a few powerful, general purpose and flexible commands.  You can d
 2. If a region is selected, the conversation will be limited to its contents.
 
 3. Call =M-x gptel-send= with a prefix argument (~C-u~)
-   - to set chat parameters (GPT model, backend, system message etc) for this buffer,
+   - to set chat parameters (model, backend, system message etc) for this buffer,
    - include quick instructions for the next request only,
    - to add additional context -- regions, buffers or files -- to gptel,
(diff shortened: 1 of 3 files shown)


metachip:
My 2c... agree with the sentiment, GPT should apply to models from OpenAI only. Also term "LLM Model" is redundant. "M" means Model. I'd just replace "GPT Model" with "Model" or perhaps "LLM". I think "Model" is better that's what they all are and not all of them are particularly large.

pabl0:
> My 2c... agree with the sentiment, GPT should apply to models from OpenAI only. Also term "LLM Model" is redundant. "M" means Model. I'd just replace "GPT Model" with "Model" or perhaps "LLM". I think "Model" is better that's what they all are and not all of them are particularly large.

Good point, it's "CD disc" again. Thanks.

karthink:
Yes, just "model" instead of "GPT model" everywhere should be fine.

karthink:
Thank you @pabl0

Issue #346, opened 2024-07-25
Enhance mode name handling in directives


nfedyashev:
This would fix the existing issue:
`You are an enh-ruby programmer. Refactor the following code.`

This is a bit similar to existing `gptel--strip-mode-suffix`(used in `gptel--rewrite-message`) but for.. prefixes as well?

Related links on enh-ruby major mode:
https://github.com/zenspider/enhanced-ruby-mode
https://stackoverflow.com/questions/30318515/what-are-advantages-of-enh-ruby-mode-vs-ruby-mode

karthink:
Well there's no easy heuristic way to solve this.  I can provide an option where you can populate major-mode -> mode string associations (e.g. `((enh-ruby-mode . "ruby"))`) for use by gptel, but perhaps there's a way to do it automatically by consulting one of the several major-mode-related settings in Emacs?  `auto-mode-alist`, `major-mode-remap-alist` etc.

nfedyashev:
Unfortunately, I'm not familiar with the modules that you mentioned so I'm not sure if there is a better way

From what I experienced, it is not a major issue in this particular case(at least not for gpt4) and LLMs can resolve this prompt ambiguity

nfedyashev:
[Related stackoverflow topic](https://emacs.stackexchange.com/questions/81832/how-to-get-human-readable-name-for-the-current-major-mode/81834#81834)

nfedyashev:
same issue with [js2](https://github.com/mooz/js2-mode) Improved JavaScript editing mode for Emacs

I don't mind too much but I think for a typical user it might be rather annoying to edit this prompt every time. Takes about 4 seconds(depends on keyboard settings) just to get to that chunk of text using backspace/arrows.

nfedyashev:
As a workaround, I'm using this patched version of the related function.

```
(defun gptel--strip-mode-suffix (mode-sym)
  "Remove the -mode suffix from MODE-NAME.

MODE-NAME is typically a major-mode symbol."
  (cond
   ((eq mode-sym 'enh-ruby-mode) "ruby")
   ((eq mode-sym 'js2-mode) "js")
   (t
    (let ((mode-name (thread-last
                       (symbol-name mode-sym)
                       (string-remove-suffix "-mode")
                       (string-remove-suffix "-ts"))))
      (if (provided-mode-derived-p
           mode-sym 'prog-mode 'text-mode 'tex-mode)
          mode-name "")))))
```

karthink:
[...]

PR #636, opened 2025-02-12
gptel: Improve major mode name handling

pabl0:
* gptel.el: (gptel--mode-description-alist, gptel--pretty-mode-name-alist): Rename alist.  Add emacs-lisp-mode ("ELisp" is not as clear).  Add cperl-mode.  Drop Yaml-mode. (gptel--strip-mode-suffix, gptel--pretty-mode-name): Replace function with new implementation with more meaningful name.  It uses the pretty name already provided by the major-mode, which typically starts with capital letter.  Using mode-name alone is not robust since it's not always a string.

* gptel-rewrite.el (gptel--rewrite-directive-default): Call the new function.  Don't downcase it.

* gptel-context.el (gptel-context--insert-buffer-string): Call the new function.

### DIFF
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -361,7 +361,7 @@ START and END signify the region delimiters."
     (let ((is-top-snippet t)
           (previous-line 1))
       (insert (format "In buffer `%s`:" (buffer-name buffer))
-              "\n\n```" (gptel--strip-mode-suffix (buffer-local-value
+              "\n\n```" (gptel--pretty-mode-name (buffer-local-value
                                                    'major-mode buffer))
               "\n")
       (dolist (context contexts)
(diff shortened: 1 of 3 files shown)


pabl0:
Oh and if this idea is approved, should I make `gptel-pretty-name-alist` a defcustom? (#346)

karthink:
@pabl0 I don't understand the purpose of this PR or how it's an improvement.  Could you give me some more details?

pabl0:
I agree that polishing may seem somewhat pointless. It often feels
that the name of a major mode is more appropriately labeled than the
`xx-mode` symbol with the `-mode` stripped away. However, in most
cases, the only improvement is the correct use of uppercase and
lowercase letters. Additionally, we still need a table to translate
names like "CPerl" to "Perl" or "ELisp" to "Emacs Lisp." Therefore, I
don't believe this is a particularly good idea after all.

PR #637, opened 2025-02-12
gptel-transient: Ensure user enters a number when prompted

pabl0:
* gptel-transient.el (gptel--transient-read-variable, gptel--transient-read-number): New function that uses standard `read-number' to read a numeric value, returning nil for default.  Drop the read-variable function since we were only using it for prompting numbers.

* gptel-transdient.el (gtpel--infix-num-messages-to-send, gptel--infix-max-tokens, gptel--infix-temperature): Use the new function as :reader.

This also fixes handling of history, which is broken with at least transient 0.8.4.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -174,13 +174,19 @@ will toggle its visibility state."
                                  prefix max-width nil nil
                                  'ellipsis))))))))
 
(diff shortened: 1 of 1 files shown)


karthink:
Thank you @pabl0.  I didn't realize the history feature was broken.

PR #638, opened 2025-02-12
gptel-transient: Fix face related bug

pabl0:
* gptel-transient (gptel-menu): Fix a slight bug introduced in commit edf834a ("Provide description of send action") where an extra quote (') caused error "Invalid face reference: quote" to appear in *Messages* and the "Context" heading to have incorrect face (color).

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -550,7 +550,7 @@ Also format its value in the Transient menu."
     (gptel--infix-add-directive)]
    [:pad-keys t ""
     (:info #'gptel--describe-infix-context
-     :face 'transient-heading :format "%d")
+     :face transient-heading :format "%d")
     (gptel--infix-context-add-region)
     (gptel--infix-context-add-buffer)
     (gptel--infix-context-add-file)


gptel-transient.el
```diff
@@ -550,7 +550,7 @@ Also format its value in the Transient menu."
     (gptel--infix-add-directive)]
    [:pad-keys t ""
     (:info #'gptel--describe-infix-context
```
karthink, line 552: I think the `#'` in `#'gptel--describe-infix-context` is not required either, as everything is quoted by default here.

pabl0: > I think the `#'` in `#'gptel--describe-infix-context` is not required either, as everything is quoted by default here.

Noted, seems to work and byte-compilation is is happy.

karthink: I meant that it should be unquoted, so just `gptel--describe-infix-context`, without the `'`.

pabl0: > I meant that it should be unquoted, so just `gptel--describe-infix-context`, without the `'`.

Nope: 

`list: Symbol’s value as variable is void: gptel--describe-infix-context`

I think the # should be there as well for proper elisp convention, right?

karthink: Sorry, my bad. In that case the `#` should also be present, please add it back in.

I'm confused about the quoting convention inside Transient group definitions now, though.

karthink:
Thank you @pabl0!

Issue #634, opened 2025-02-12
New Gemini models


artem110101:
There has been new changes to Gemini models

https://blog.google/technology/google-deepmind/gemini-model-updates-february-2025/

Would be great to get those into **gptel**.

My understanding of the changes needed to satisfy this:

1. [gemini-2.0-pro-exp-02-05](https://ai.google.dev/gemini-api/docs/models/experimental-models#available-models)
2. [gemini-2.0-flash-lite-preview-02-05](https://ai.google.dev/gemini-api/docs/models/gemini#model-variations)
3. [gemini-2.0-flash](https://ai.google.dev/gemini-api/docs/models/gemini#model-variations)

I wanted to create PR first but wanted to make sure I'm not missing anything.

metachip:
This works for me:
```elisp
(gptel-make-gemini "Gemini"
  :protocol "https"
  :host "generativelanguage.googleapis.com"
  :key (lambda () (auth-source-pick-first-password
                    :host "generativelanguage.googleapis.com"
                    :user "gemini-api-key"))
  :stream t
  :models '((gemini-2.0-flash
              :description "Next gen, high speed, multimodal for a diverse variety of tasks"
              :capabilities (tool-use json media)
              :mime-types ("image/png" "image/jpeg" "image/webp" "image/heic" "image/heif"
                            "application/pdf" "text/plain" "text/csv" "text/html")
              :context-window 1000
              :input-cost 0.10
              :output-cost 0.40
              :cutoff-date "2024-08")
             (gemini-2.0-flash-lite-preview-02-05
               :description "Gemini 2.0 Flash model optimized for cost efficiency and low latency"
               :capabilities (json)
               :context-window 1000
               :input-cost 0.075
               :output-cost 0.30
               :cutoff-date "2024-08")
             (gemini-2.0-pro-exp-02-05
               :description "Next gen, high speed, multimodal for a diverse variety of tasks"
               :capabilities (json)
               :context-window 2000
               :input-cost 0.00
               :output-cost 0.00
               :cutoff-date "2024-08")
             (gemini-2.0-flash-thinking-exp-01-21
               :description "Next gen, high speed, multimodal for a diverse variety of tasks"
               :capabilities (json)
               :input-cost 0.00
               :output-cost 0.00
               :cutoff-date "2024-08")
             (gemini-2.0-flash-exp
               :description "Multi-modal, streaming, tool use 2000 RPM"
               :capabilities (tool-use json media)
[...]

PR #639, opened 2025-02-13
Update gptel-gemini.el models

braun-steven:
See #634, suggestion by @metachip.

Adds
- `gemini-2.0-flash`
- `gemini-2.0-flash-exp`
- `gemini-2.0-flash-lite-preview-02-05`
- `gemini-2.0-flash-thinking-exp-01-21`
- `gemini-2.0-pro-exp-02-05`

Removes
- `gemini-2.0-flash-thinking-exp` (superseded by `gemini-2.0-flash-thinking-exp-01-21`)

### DIFF
--- a/gptel-gemini.el
+++ b/gptel-gemini.el
@@ -345,13 +345,6 @@ files in the context."
      :input-cost 0.075
      :output-cost 0.30
      :cutoff-date "2024-10")
-    (gemini-2.0-flash-thinking-exp
-     :description "Stronger reasoning capabilities."
-     :capabilities (tool-use media)
-     :context-window 32
-     :mime-types ("image/png" "image/jpeg" "image/webp" "image/heic" "image/heif"
-                  "text/plain" "text/csv" "text/html")
-     :cutoff-date "2024-08")
     (gemini-exp-1206
      :description "Improved coding, reasoning and vision capabilities"
      :capabilities (tool-use json media)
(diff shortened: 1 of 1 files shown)


karthink:
Thanks @braun-steven!

Instead of removing `gemini-2.0-flash-thinking-exp`, please "deprecate" it by changing its `:description` to "DEPRECATED: Please use <replacement>", and move it to the end of the model list so it shows up last.  We will remove it in the release after the next.

braun-steven:
Thanks for the quick review! I've applied the changes you suggested.

benthamite:
Hi @braun-steven, any reason

benthamite:
Hi @braun-steven, I noticed you set the input and output costs of some Gemini models to $0. It seems like, for Gemini, we are quoting the prices listed for the “paid tier” [here](https://ai.google.dev/gemini-api/docs/pricing), in which case the input and output cost should be $0.10 and $0.40, respectively. Or am I missing something? Thanks!

metachip:
My understanding is that some models in experimental phase are not charged until the production versions are released.

braun-steven:
According to the model selection in https://aistudio.google.com/prompts/new_chat, the experimental models are all free:

SCR-20250224-hhnd-2
SCR-20250224-hhcx-2
SCR-20250224-hguu-2

PR #640, opened 2025-02-13
gptel-transient: Use proper ellipsis for consistent UI

pabl0:
* gptel-transient.el (gptel-option-overlaid): Use a proper ellipsis (…) instead of three ASCII dots (...) when truncating the user instruction, provided Unicode is displayable and `truncate-string-ellipsis' is nil.  This change allows for two additional characters and enhances visual appeal and consistency.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -512,7 +512,7 @@ This is used only for setting this variable via `gptel-menu'.")
   "Transient options for overlays displayed in the working buffer.")
 
 (cl-defmethod transient-format-value ((obj gptel-option-overlaid))
-  "set up the in-buffer overlay for additional directive, a string.
+  "Set up the in-buffer overlay for additional directive, a string.
 
 Also format its value in the Transient menu."
   (let ((value (oref obj value))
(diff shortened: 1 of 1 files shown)


gptel-transient.el
```diff
@@ -532,7 +532,7 @@ Also format its value in the Transient menu."
         (add-hook 'transient-exit-hook ov-clear-hook)))
     ;; Updating transient menu display
     (if value
-        (propertize (concat argument (truncate-string-to-width value 25 nil nil "..."))
+        (propertize (concat argument (truncate-string-to-width value 39 nil nil 'ellipsis))
```
karthink, line 535: Any reason you changed the width to 39?

Also please just use `t` instead of `'ellipsis`.  I know gptel uses both `t` and `'ellipsis` across its code for this (mostly `t`). I plan to switch to just `t`.

pabl0: > Any reason you changed the width to 39?

It was explained in commit msg of second version.  Basically it uses all available space without making the default transient more wide:

ellipsis39

But of course the width depends on other settings as well so 39 feels way too specific.  Changed it to 35, or do you think it should be as short as possible? Of course the full directive is visible in the actual buffer as well.

ellipsis35

> Also please just use `t` instead of `'ellipsis`. I know gptel uses both `t` and `'ellipsis` across its code for this (mostly `t`). I plan to switch to just `t`.

Yes I agree it's better, done.

karthink: 35 is fine, I'll merge this.  But I wonder if it's possible to show the additional instruction on the following line, where it can be wider.  I remember trying and failing to do that.

PR #642, opened 2025-02-13
gptel-transient: Report what is being sent from kill-ring

pabl0:
* gptel-transient.el (gptel--describe-suffix-send): Enhance reporting to include the line count of the most recent kill from the kill ring, or indicate "empty" if the kill ring is devoid of entries.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -311,8 +311,7 @@ which see."
 (defun gptel--describe-suffix-send ()
   "Describe the action of `gptel--suffix-send'."
   (cl-flet ((ptv (s) (propertize s 'face 'warning))
-            (pth (s) (propertize s 'face 'transient-heading))
-	    (pte (s) (propertize s 'face 'error)))
+            (pth (s) (propertize s 'face 'transient-heading)))
     (let* ((args (or (and transient-current-command
                           (transient-args transient-current-command))
 	             ;; Not yet exported, simulate.  HACK: We are accessing
@@ -352,18 +351,17 @@ which see."
 			     (if-let* ((val (car-safe kill-ring))
 				       (val (substring-no-properties val))
 				       (len (length val)))
-				 (ptv (concat "\""
-					      (string-replace
-					       "\n" "⮐"
-					       (truncate-string-to-width
-						val 20 nil nil t))
-					      "\""
-					      (when (> len 20)
-						(concat
-						 ", "
-						 (file-size-human-readable len 'si " ")
-						 " chars"))))
-			     (pte "empty!"))
+				 (ptv (concat
+                                       "\"" (string-replace
+					     "\n" "⮐"
+					     (truncate-string-to-width
+					      val 20 nil nil t))
+				       "\"" (when (> len 20)
+					      (concat
+					       ", "
+					       (file-size-human-readable len 'si " ")
+					       " chars"))))
+			     (propertize "empty" 'face 'error))
 			   (ptv ")"))
                      context
                      (if dest (concat (pth ", response to ") dest)
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -311,7 +311,8 @@ which see."
 (defun gptel--describe-suffix-send ()
   "Describe the action of `gptel--suffix-send'."
   (cl-flet ((ptv (s) (propertize s 'face 'warning))
-            (pth (s) (propertize s 'face 'transient-heading)))
+            (pth (s) (propertize s 'face 'transient-heading))
+	    (pte (s) (propertize s 'face 'error)))
     (let* ((args (or (and transient-current-command
                           (transient-args transient-current-command))
 	             ;; Not yet exported, simulate.  HACK: We are accessing
@@ -346,7 +347,24 @@ which see."
                      (if dest (concat (pth ", response to ") dest)
                        (concat (pth ", insert response at point")))))
             ((member "y" args)
-             (concat (pth "Send prompt from ") (ptv "kill-ring")
+             (concat (pth "Send prompt from ")
+		     (concat (ptv "kill-ring (")
+			     (if-let* ((val (car-safe kill-ring))
+				       (val (substring-no-properties val))
+				       (len (length val)))
+				 (ptv (concat "\""
+					      (string-replace
+					       "\n" "⮐"
+					       (truncate-string-to-width
+						val 20 nil nil t))
+					      "\""
+					      (when (> len 20)
+						(concat
+						 ", "
+						 (file-size-human-readable len 'si " ")
+						 " chars"))))
+			     (pte "empty!"))
+			   (ptv ")"))
                      context
                      (if dest (concat (pth ", response to ") dest)
                        (concat (pth ", insert response at point")))))


gptel-transient.el
```diff
@@ -346,7 +346,13 @@ which see."
                      (if dest (concat (pth ", response to ") dest)
                        (concat (pth ", insert response at point")))))
             ((member "y" args)
-             (concat (pth "Send prompt from ") (ptv "kill-ring")
+             (concat (pth "Send prompt from ")
+		     (ptv (concat "kill-ring ("
+				  (if-let* ((val (car-safe kill-ring))
+					    (len (length (split-string val "\n" t))))
```
psionic-k, line 352: Regex scan this rather than split.  Lower garbage.

pabl0: > Regex scan this rather than split. Lower garbage.

Good point, thanks @psionic-k . Do you happen to know what would be the best implementation? Tried benchmarking a few simple ways of doing this and surprisingly this performed best (faster than the simplest possible loop of `string-match`), even if might not be very elegant:

```elisp
    (benchmark-run 1000
      (with-temp-buffer
	(insert my-large-string)
	(count-lines (point-min) (point-max)))
      (garbage-collect))
```

This took ~110 seconds with `my-large-string`having 10,000 lines and 1108890 characters. (Tested on rather sluggish 1.1 GHz tablet PC) 

How likely it is that a user would use the kill ring to send tens of thousands of lines to an LLM, so how far do we need to optimize? Of course efficient code is always preferred,

BTW my first thought was simply reporting the number of characters, but somehow it feels more consistent to use lines since that's what the other modes use (well, range of line numbers).

karthink:
I think mentioning the _number_ of lines in the current kill is not useful information.  In the other cases, line numbers are indicated because that gives the user an idea of which _part_ of the buffer is going to be sent, acting as a quick check.  So the important bit here is the _positions_, not the number of lines.

When sending the current kill, the uncertainty is about what is going to be sent, not how many lines it is.  ("Is the top of the kill-ring the string that I think it is, or did I kill something else afterwards without realizing it?  Did the system clipboard change and mess up my kill-ring?")

So if we want to improve this we should provide a preview of what will be sent.  Something as basic as the first few characters of the current-kill should be enough.

```
RET Send prompt from kill-ring ("Some text..."), insert response at point.
```

More extensive solutions are possible, like popping up a buffer showing the full string that will be sent, or allowing the user to select the kill-ring item to be sent.  But I don't want to get carried away, something simple should be fine.

pabl0:
> So if we want to improve this we should provide a preview of what will be sent. Something as basic as the first few characters of the current-kill should be enough.
> 
> ```
> RET Send prompt from kill-ring ("Some text..."), insert response at point.
> ```

Thanks for your insightful feedback.

Yes that might work actually. My main motivation was just to give some indication of how much data I am sending to some computer owned by some AI company somewhere and give we warning if I had screwed up and it's way more than I expected. It's true that seeing part of the actual text is better in most use cases, but maybe not always. What if, for instance, the beginning of the text is what I'm planning to send, but instead of one paragraph I managed to kill a whole 100K confidential file? How about having both the snippet and number of lines/characters as well?
 
This is actually my main concern with gptel, it's way too easy to send data by mistake that I'd rather keep private, so I'm thinking of what kind of safeguards we could have to protect against accidental leaks.

karthink:
Yeah you could show the length.  The number of lines is not very reliable as you can have any amount of data on one line, like if you copy some json or shell output.

```
RET Send prompt from kill-ring ("{"foo": "bar", ...", 120k chars), insert response at point
```

pabl0:
> Yeah you could show the length. The number of lines is not very reliable as you can have any amount of data on one line, like when you copy some json or shell output.
> 
> ```
> RET Send prompt from kill-ring ('{"foo": "bar", ...', 120k chars), insert response at point
> ```

Very much true, and also there is always the question should empty lines count. The number of lines actually really bad metric for anything practical. I'll test this idea.

pabl0:
Please check this latest iteration. I am pretty happy about how it looks and I think it's useful. Thanks.

gptel-transient.el
```diff
@@ -346,7 +346,22 @@ which see."
...
-             (concat (pth "Send prompt from ") (ptv "kill-ring")
+             (concat (pth "Send prompt from kill-ring (")
+		     (ptv (concat
+			   (if-let* ((val (car-safe kill-ring))
+				     (val (substring-no-properties val))
+				     (len (length val)))
+			       (concat "\""
+				       (string-replace "\n" "⮐"
+						      (truncate-string-to-width
+						       val 20 nil nil 'ellipsis))
```
psionic-k, line 357: I like this better.  The reason we don't want transient UI things to be too heavy is that they are updated in order to display the transient menu.  This solution is a better approach than the first.

gptel-transient.el
karthink: Here's my version of counting newlines:

``` emacs-lisp
(cl-loop for c across (car-safe kill-ring)
         when (= c 10) count it)
```

gptel-transient.el
```diff
@@ -346,7 +346,22 @@ which see."
                      (if dest (concat (pth ", response to ") dest)
                        (concat (pth ", insert response at point")))))
             ((member "y" args)
-             (concat (pth "Send prompt from ") (ptv "kill-ring")
+             (concat (pth "Send prompt from kill-ring (")
```
karthink, line 349: The string `kill-ring ("some text...", 20K chars)` should all be `ptv`'d.

This is consistent with how the other options are shown, every piece of text describing the source is passed to `ptv`.

gptel-transient.el
```diff
@@ -346,7 +346,22 @@ which see."
...
+			       (concat "\""
+				       (string-replace "\n" "⮐"
+						      (truncate-string-to-width
+						       val 20 nil nil 'ellipsis))
+				       "\""
+				       (when (> len 20)
+					 (concat ", "
+						 (file-size-human-readable len 'si " ")
+						 " chars")))
+			     "empty!")))
```
karthink, line 363: Show this as `(propertize "empty" 'face 'error)` to highlight that the kill-ring is empty.

gptel-transient.el
```diff
@@ -346,7 +346,22 @@ which see."
                      (if dest (concat (pth ", response to ") dest)
                        (concat (pth ", insert response at point")))))
             ((member "y" args)
-             (concat (pth "Send prompt from ") (ptv "kill-ring")
+             (concat (pth "Send prompt from kill-ring (")
+		     (ptv (concat
```
karthink, line 350: The `concat` on this line isn't doing anything, unless I misread the parens.

gptel-transient.el
karthink, line 357: Please change `'ellipsis` to just `t`.  (See my comment in your other PR about this for why.)

gptel-transient.el
pabl0: > The string `kill-ring ("some text...", 20K chars)` should all be `ptv`'d.
> 
> This is consistent with how the other options are shown, every part of the text describing the source is passed to `ptv`.

That was my first thinking, but this just looked so much more visually pleasing that after comparing them for quite some time I decided to go with this. But I'll change it if you think consistency is more important.

gptel-transient.el
pabl0: > Here's my version of counting newlines:
> 
> ```emacs-lisp
> (cl-loop for c across (car-safe kill-ring)
>          when (= c 10) count it)
> ```

That is so neat. Motivates me to teach myself CL loops properly!

karthink:
Thank you @pabl0!

PR #645, opened 2025-02-15
transient-el: Separate key and buffer name of response buffers

pabl0:
* transient.el (gptel-menu, gptel--suffix-send): Separate `b' (other buffer) and `g' (gptel session) argument name and the buffer name with an equal sign `='.  This aligns with the formatting of transient options in other packages like Magit, enhancing readability, particularly when buffer names start with the letter `b' or `g'.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -332,8 +332,8 @@ which see."
                   ((member "k" args) (ptv "kill-ring"))
                   ((cl-some (lambda (s)
                               (and (stringp s) (memq (aref s 0) '(?g ?b))
-                                   (not (equal (substring s 1) (buffer-name)))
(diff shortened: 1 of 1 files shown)


karthink:
The prefix and the buffer are colored differently.  Do you still find it confusing?

pabl0:
Actually, I am not so sure if we want this after all.

If you look closely, in Magit only the GNU style long options (`--author=`) have `=`, but not single character options (`-G`), which make sense since that's the convention with shell commands.

magit-options

Of course we're not dealing with external commands, but maybe it is more consistent to leave it as-is or perhaps use just a space.

karthink:
I'm content to leave it as it is for now.  If we get any complaints that it's not readable enough we can consider adding a space.

PR #652, opened 2025-02-18
gptel-transient: Polish gptel--read-with-prefix user experience

pabl0:
* gptel-transient.el (gptel--read-with-prefix-help): Turn into a function to show one function.
(gptel--read-with-prefix): Enable resizing minibuffer (to make cycling the ov meaningful). Drop `make-separator-line' which is buggy. Move the prompt to the last line, which looks better and avoids cursor jumping. Use `window-max-chars-per-line' instead of `window-width' (which is incorrect in terminal). Replace newlines with return symbol. Calculate the multi-line layoyt correctly to use all available space on the last line. Cycle between just two modes: one line and multiple lines (hiding one line does not add much value). Don't show the TAB help if the whole prefix fits into one line. (gptel--infix-add-directive): Use simple prompt.

* gptel-rewrite.el (gptel--infix-rewrite-extra): Use simple prompt.

Replace RETURN LEFT (U+2C90) with smaller RETURN SYMBOL (U+23CE), which looks more readable (with many typefaces) between two pieces of text.


pabl0:
I made slight refactoring changes (use cl-flet) and fixed a bug related to calling `gptel--read-with-prefix` with a `nil` value.

pabl0:
Have you had the opportunity to review this? I find the current implementation a bit clumsy, and this took unexpectedly significant effort to achieve. :-)

karthink:
Not yet, as it looks pretty involved.  I will get to this, I'm busy writing the gptel manual at the moment.

PR #654, opened 2025-02-18
gptel: Support prefixing dynamic system messages with lambda

pabl0:
* gptel.el (gptel--parse-directive): Add optional `prefix-lambda` argument to prepend a lambda symbol (λ) and the first line of the documentation string, if present, to the system message.

* gptel-prefix.el (gptel--infix-add-directive): Call `gptel--parse-directive' with the new `prefix-lambda' arg.

---
This is of course much more fun if #652 gets merged.


pabl0:
(And pushed a brown bag bugfix, never use setcar when tired :)

PR #659, opened 2025-02-22
gptel-org: Improve error handling in response boundary saving

benthamite:
This pull request addresses a `wrong-type-argument number-or-marker-p nil` error that occurs in `gptel-org--save-state` when saving an Org buffer with `gptel-mode` enabled before any AI responses have been received.

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -416,14 +416,14 @@ non-nil (default), display a message afterwards."
    ;; Save response boundaries
    (letrec ((write-bounds
              (lambda (attempts)
-               (let* ((bounds (gptel--get-buffer-bounds))
-                      (offset (caar bounds))
-                      (offset-marker (set-marker (make-marker) offset)))
-                 (org-entry-put (point-min) "GPTEL_BOUNDS"
-                                (prin1-to-string (gptel--get-buffer-bounds)))
-                 (when (and (not (= (marker-position offset-marker) offset))
-                            (> attempts 0))
-                   (funcall write-bounds (1- attempts)))))))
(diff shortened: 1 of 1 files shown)


pabl0:
Looks good, but unnecessary whitespace changes (since the project mostly uses spaces and no tabs for indentation which is not the Emacs default).

psionic-k:
Bounds checking is getting overhauled.  I confirmed the bug exists in my branch as well.

benthamite:
If I understand the new changes correctly, my fix is no longer needed. So I’m closing this pull request.

karthink:
> If I understand the new changes correctly, my fix is no longer needed. So I’m closing this pull request.

I actually haven't tested the new save-state function for the case which your PR is addressing (a buffer with no assistant messages).  I'll test it when I can, or you can let me know if the bug is gone.

benthamite:
Upon further investigation, I believe I had misdiagnosed the source of the error. The issue arises when the gptel buffer is saved AND that buffer is not the current buffer (for example, if saved by calling `save-some-buffers` from a buffer that is not the gptel buffer). In such a case, `gptel--get-buffer-bounds` in `gptel-org--save-state` will return `nil` and the function will eventually throw an error.

It looks like this happens because `gptel-org--save-state` assumes that it is being called from a gptel buffer. So one option is to make this assumption explicit and wrap the function in a check that ensures that `gptel-mode` is non-nil. One could also, as I did in my original PR, use `when-let` for the let-bound variables, though the intent of this fix is less clear, and also allows all the code up until the `when-let` macro to be needlessly evaluated.

PR #661, opened 2025-02-24
Add a --noproxy option to gptel-curl--get-args

ElleNajt:
This adds a --noproxy option to the call to curl in gptel-curl--get-args, along with a variable to configure that option.

In an development environment I use, I need to set the noproxy environment variable in emacs to get various features to work, but to communicate with an LLM endpoint I need to call curl with a conflicting noproxy variable. This provides a convenient way to manage this.


karthink:
Added `gptel-curl-extra-args` where you can specify whatever extra command line arguments are required.  (if you want to set Curl arguments for a specific backend, you can use the gptel-backend's `:curl-args` slot instead of this.)

Issue #513, opened 2024-12-21
Adding context files recursively from a directory


hartikainen:
Thanks for such a great and pleasant-to-use package!

I've had several occasions recently where I would've wanted to add the contents of a whole directory to the context of my model. I can do that manually but with more than a couple of files that becomes quite clumsy. It would be great if there was a function like `gptel-add-directory` or similar, which would allow files in the directory to be added, ideally recursively.

metachip:
Hi @hartikainen 

This is a great idea.

It is related to a wider discussion about how to manage context beyond the "labour-intensive mechanism" that it offers now. Especially if one is used to using markdown or org mode as the source syntax for dialogue with the models. My view is that the management of context within gptel is an area that can be improved significantly with what are, conceptually at least, relatively minor changes.

The ability to add directories of files could be difficult to "keep simple". All nested subdirectories? Only files with specific extensions? Which ones? How should they be filtered? I'm sure there are ways to simplify this using metaphors like "dired", which can already be used in a way similar to this, But it is not automated in the way I think you are suggesting.

Have a look at the summary below and the related issues on this topic.

See what you think :) 

## 1. Context Buffer Navigation (#459)

**Suggestion**:

* Make context previews collapsible in the context buffer
* Hide previews by default with TAB to reveal (similar to Magit)
* Current preview chunks can be unwieldy when large

**Status**: Under consideration, but no specific implementation committed yet.

## 2. Multiple Context Sets (#475, #459)

**Suggestion**:

* Support multiple named context buffers
* Allow association between context sets and chat buffers
* Use prefix arguments (C-u) to select context sets
* Add widget to chat buffers showing active context

**Karthink's Response** (#475):

* Concerned about user confusion and complexity
* Prefers one-shot interactions without additional decisions
* Raises questions about ambiguity in context assignment
* Notes problems with maintaining internal registry
* Suggests project-local contexts might be more intuitive

## 3. Links as Context (#481)

**Suggestion**:

* Allow links in non-chat buffers
* Support text file content via links
* Remove standalone link requirement
* Use `:CONTEXT:` drawer in org-mode to isolate LLM links
* Support region targeting via link labels

**Karthink's Response**:

* Supportive of using links for per-buffer context
* Concerned about accidental inclusion of links
* Prefers standalone links for clear intent
* Willing to add text file support
* Cautious about org-specific link types

## 4. Quick Context Management (#486)

**Suggestion**:

* Add command to quickly delete all context
[...]

PR #665, opened 2025-02-26
gptel-context: Allow exclusion of gitignored files

benthamite:
Continuing the discussion [here](https://github.com/karthink/gptel/issues/513), this PR allows the user to exclude, by setting the user option `gptel-context-exclude-gitignored` to `t`, gitignored files or directories from the context.

Tagging participants in previous discussion: @hartikainen, @collinarnett, @metachip, @sandinmyjoints.

### DIFF
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -206,10 +206,19 @@ Return PATH if added, nil if ignored."
 (defun gptel-context--add-directory (path action)
   "Process all files in directory at PATH according to ACTION.
 ACTION should be either `add' or `remove'."
-  (let ((files (directory-files-recursively path "." t)))
+  (let ((files (directory-files-recursively path "." t))
+        (git-root (and gptel-context-exclude-gitignored
+                       (executable-find "git")
+                       (locate-dominating-file path ".git")))
+        unignored-files)
+    ;; Get the list of unignored files once if needed
+    (when git-root
+      (setq unignored-files (gptel-context--git-files git-root)))
     (mapc (lambda (file)
             (unless (or (file-directory-p file)
-			(gptel-context--skip-file-p file))
+                        (and git-root
+                             (not (member (file-relative-name file git-root)
+                                          unignored-files))))
               (pcase-exhaustive action
                 ('add
                  (if (gptel--file-binary-p file)
@@ -228,37 +237,16 @@ If PATH is a directory, recursively add all files in it.
 PATH should be readable as text."
   (interactive "fChoose file to add to context: ")
   (cond ((file-directory-p path)
-	 (gptel-context--add-directory path 'add))
-	((gptel-context--skip-file-p path)
-	 (message "Skipping git-ignored file: %s" path))
-	((gptel--file-binary-p path)
+         (gptel-context--add-directory path 'add))
+        ((and gptel-context-exclude-gitignored
+              (gptel-context--skip-file-p path))
+         (message "Skipping git-ignored file: %s" path))
+        ((gptel--file-binary-p path)
          (gptel-context--add-binary-file path))
-	(t (gptel-context--add-text-file path))))
+        (t (gptel-context--add-text-file path))))
 
 ;;;###autoload (autoload 'gptel-add-file "gptel-context" "Add files to gptel's context." t)
 (defalias 'gptel-add-file #'gptel-context-add-file)
-
-(defvar gptel-context--git-cache (make-hash-table :test 'equal)
-  "Cache of `.gitignore' hashes and unignored files per Git root.
-Keys are directory paths.  Each value is a plist of the form:
-  (:files <list-of-unignored> :hash <.gitignore-md5-or-nil>)
-
-If the .gitignore changes on disk, we automatically refresh
-the unignored file list and hash.  If you modify the ignores in
-other ways (e.g. .git/info/exclude), you may need a manual reset:
-  (setq gptel-context--git-cache (make-hash-table :test 'equal))")
-
-(defun gptel-context--gitignore-hash (git-root)
-  "Return an MD5 hash of the top-level `.gitignore' file in GIT-ROOT, else nil.
-Ignores other exclude files (like .git/info/exclude). We only track the
-top-level `.gitignore' file for a quick on-disk change check."
-  (let ((ignore-path (expand-file-name ".gitignore" git-root)))
-    (when (file-exists-p ignore-path)
-      (condition-case _
-          (with-temp-buffer
-            (insert-file-contents ignore-path)
-            (secure-hash 'md5 (current-buffer)))
-        (error nil)))))
 
 (defun gptel-context--git-files (dir)
   "Return a list of unignored files in the Git repo at DIR."
@@ -270,33 +258,16 @@ top-level `.gitignore' file for a quick on-disk change check."
        nil))))
 
 (defun gptel-context--skip-file-p (file)
-  "Return non-nil if FILE should be skipped.
-If `gptel-context-exclude-gitignored' is non-nil and Git is found, treat any
-FILE that does not appear in `gptel-context--git-files' as ignored.
-
-Caches unignored files and the MD5 of .gitignore in `gptel-context--git-cache'.
-If the `.gitignore' file changes, automatically re-scan the repo’s unignored
-files. If the Git executable is missing or we’re not inside a Git repo, return
-nil (not skipped)."
+  "Return non-nil if FILE should be skipped due to gitignore rules.
+This function assumes it will be called with a list of unignored files
+available or will make a direct Git call for a single file check."
   (when (and gptel-context-exclude-gitignored
              (executable-find "git"))
     (when-let* ((git-root (locate-dominating-file file ".git"))
                 (rel-path (file-relative-name file git-root)))
-      ;; Look up or populate the cache for this Git root.
-      (let* ((entry (gethash git-root gptel-context--git-cache))
-             (cached-files (plist-get entry :files))
-             (cached-hash  (plist-get entry :hash))
-             (current-hash (gptel-context--gitignore-hash git-root)))
-        (cond
-         ;; If we already have a cached list of files and the .gitignore is unchanged:
-         ((and cached-files (string= cached-hash current-hash))
-          (not (member rel-path cached-files)))
-         ;; Otherwise refresh:
-         (t
-          (let ((unignored (gptel-context--git-files git-root)))
-            (puthash git-root `(:files ,unignored :hash ,current-hash)
-                     gptel-context--git-cache)
-            (not (member rel-path unignored)))))))))
+      ;; Get unignored files directly for this single file
+      (let ((unignored (gptel-context--git-files git-root)))
+        (not (member rel-path unignored))))))
 
 (defun gptel-context-remove (&optional context)
   "Remove the CONTEXT overlay from the contexts list.
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -84,6 +84,16 @@ context chunk.  This is accessible as, for example:
   :group 'gptel
   :type 'function)
 
+(defcustom gptel-context-exclude-gitignored t
+  "When non-nil, exclude files that are ignored by Git from gptel context.
+
+This is checked by calling \"git check-ignore -q FILE\" for each file
+when adding a directory.  If the command succeeds, gptel will not add
+that file to its context.  Requires Git installed and accessible from
+Emacs."
+  :group 'gptel
+  :type 'boolean)
+
 (defun gptel-context-add (&optional arg confirm)
   "Add context to gptel in a DWIM fashion.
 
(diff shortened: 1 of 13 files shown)


benthamite:
(I just noticed that I forgot to set the default value of `gptel-context-exclude-gitignored` to `nil`, but now I wonder if we should leave it as `t`. I don’t have a clear view on this.)

collinarnett:
I would set it to `t` by default. I tested out the new add directory functionality on a repository with `node_modules` folder and it froze my emacs adding thousands of files to the context window. Other utilities like [RepoMix](https://github.com/yamadashy/repomix) enable gitignore exclusion by default as well.

gptel-context.el
```diff
@@ -219,14 +237,38 @@ If PATH is a directory, recursively add all files in it.
 PATH should be readable as text."
   (interactive "fChoose file to add to context: ")
   (cond ((file-directory-p path)
-	 (gptel-context--add-directory path 'add))
-	((gptel--file-binary-p path)
+         (gptel-context--add-directory path 'add))
+        ((and gptel-context-exclude-gitignored
+              (gptel-context--skip-file-p path))
+         (message "Skipping git-ignored file: %s" path))
```
karthink, line 243: Add to this message something like: `Unset gptel-context-exclude-gitignored to include this file`

gptel-context.el
```diff
@@ -84,6 +84,13 @@ context chunk.  This is accessible as, for example:
   :group 'gptel
   :type 'function)
 
+(defcustom gptel-context-exclude-gitignored t
```
karthink, line 87: `-exclude-gitignored` was confusing to me at first because it reads like there's a double negative (even though there isn't).

How about `gptel-context-respect-gitignore`?  Is that also confusing?

gptel-context.el
```diff
@@ -219,14 +237,38 @@ If PATH is a directory, recursively add all files in it.
 PATH should be readable as text."
   (interactive "fChoose file to add to context: ")
   (cond ((file-directory-p path)
-	 (gptel-context--add-directory path 'add))
-	((gptel--file-binary-p path)
+         (gptel-context--add-directory path 'add))
+        ((and gptel-context-exclude-gitignored
+              (gptel-context--skip-file-p path))
```
karthink, line 242: I see the following problem here: Imagine you mark 30 files in Dired (but not a directory) and run `gptel-add`.  Then `(gptel-context--git-files)` will run 30 times, which means `git ls-files` will be called 30 times -- this is quite bad, especially on Windows.

I think this requires either the results of `gptel-context--git-files` to be cached, or the control flow to be changed somehow so that it only runs once when calling `gptel-add` on a bunch of files at once.

gptel-context.el
```diff
@@ -219,14 +237,38 @@ If PATH is a directory, recursively add all files in it.
...
 (defalias 'gptel-add-file #'gptel-context-add-file)
 
+(defun gptel-context--git-files (dir)
+  "Return a list of unignored files in the Git repo at DIR."
+  (let ((default-directory dir))
+    (condition-case err
+        (process-lines "git" "ls-files" "--cached" "--others" "--exclude-standard")
+      (error
+       (message "Error running git ls-files in %s: %S" dir err)
+       nil))))
```
karthink, line 258: If git fails for some reason, this means there's no way to add the file to gptel's context.  I think we should fail in the other direction -- message that git couldn't run so we don't know if the file(s) should be ignored, but we add it anyway.

karthink (changes requested):
Looks good.  I've got one performance concern and one behavior suggestion.

gptel-context.el
collinarnett: How about just extending it a bit `gptel-context-exclude-gitignore-content`?

benthamite: Sounds good, though I lean towards adding a "d" at the end of "gitignore": `gptel-context-exclude-gitignored-content`. It could also be `gptel-context-exclude-gitignored-files`, though I guess there’s the annoying ambiguity of "file", which may refer to either files in the narrow sense that excludes directories and to files in the broad sense that includes both directories and files in the narrow sense.

Separtely, we should probably use a consistent spelling: `git-ignored` or `gitignored`?

karthink: It was just my reaction to seeing the name -- we don't need to get hung up on this, it can be changed later if required.  I think even `gptel-context-gitignore` is fine, as I expect most users will immediately recognize the meaning.  The more important thing is messaging the user about why a file(s) was not added to the context and how to change it.

gptel-context.el
benthamite: I made some changes to address this.

gptel-context.el
benthamite: Addressed.

karthink:
Could you move all the git-related functions `gptel-context--git-files`, `gptel-context--skip-file-p`, `gptel-context--message-git-skipped` and `gptel-context--build-git-cache` into one place in the file?  You can add a `;;; git tracking functions` header over them for navigation.

Additionally, anywhere you run a file-name function like `file-relative-name` on a file, please check that the file argument can't be nil.  I haven't analyzed all possible cases.

gptel-context.el
```diff
@@ -119,16 +126,25 @@ context chunk.  This is accessible as, for example:
...
+			 (lambda (f) (gptel-context-add-file f git-cache)))
+		       file))
+	  (mapc action-fn files)))))
    ;; If in an image buffer
    ((and (derived-mode-p 'image-mode)
-         (gptel--model-capable-p 'media)
-         (buffer-file-name))
+	 (gptel--model-capable-p 'media)
+	 (not (gptel-context--skip-file-p (buffer-file-name)))
+	 (buffer-file-name))
```
karthink, line 147: Check for `(buffer-file-name)` first.  It can be `nil` and `(gptel-context--skip-file-p (buffer-file-name))` will throw an error.

karthink:
`locate-dominating-file` can be _very slow_ on remote files, to the point of locking up Emacs for a while.

It might be worth using `file-remote-p` and skipping the gitignore check if it's a remote file.

vermiculus:
Late to the party here (and I wouldn't want to derail this feature; I was recently bitten by it), but was the use of `project.el` (via `project-files`) considered now that it's included in emacs? That would get you cross-VCS support for free.

tlonteam:
Hi @vermiculus, I didn’t consider it because I don’t use `project.el` and am not familiar with that package. Let me take a look...

benthamite:
Hi @vermiculus. I didn’t consider using `project.el` because I am not familiar with that package. Let me take a look and I’ll see if I can adapt it.

benthamite:
@vermiculus: I revised the PR to make it use `project.el`. The code is simpler now.

It would be great if folks could test this before it is merged: I’m not a programmer and my eyes may fail to catch problems.

gptel-context.el
```diff
@@ -195,38 +211,85 @@ Return PATH if added, nil if ignored."
...
+	     (if (gptel--file-binary-p file)
+		 (gptel-context--add-binary-file file)
+	       (gptel-context--add-text-file file)))
+	    ('remove
+	     (setf (alist-get file gptel-context--alist nil 'remove #'equal) nil))))))
     (when (eq action 'remove)
       (message "Directory \"%s\" removed from context." path))))
 
-(defun gptel-context-add-file (path)
+(defun gptel-context-add-file (path &optional project-files)
```
vermiculus, line 240: It seems pretty odd (to me) to expose this in public API. I would expect this function (just going off its name) to add the named file.

If this is intended to opt into / out of automatically excluding ignored files, I would kinda expect that to be controlled by let-binding the option. This differs from codebase to codebase, though, so I might be off there.

gptel-context.el
```diff
@@ -195,38 +211,85 @@ Return PATH if added, nil if ignored."
...
+
+(defun gptel-context--message-skipped (file &optional project-files)
+  "Message that FILE is skipped because it is not a project file.
+PROJECT-FILES is a list of all files in the current project."
+  (let* ((type (if (file-directory-p file) "directory" "file"))
+	 (var (symbol-name 'gptel-context-restrict-to-project-files))
+	 (reminder (format "To include this file, unset `%s'." var)))
+    (if-let* ((project (project-current nil (car-safe project-files)))
+	      (root (project-root project))
+	      (rel-file (file-relative-name file root)))
```
vermiculus, line 289: This kind does not appear to have appropriate indentation; I would consider using aggressive-indent-mode when writing lisp.

gptel-context.el
```diff
@@ -195,38 +211,85 @@ Return PATH if added, nil if ignored."
...
+  "Return a list of files in the project DIR, or nil if no project is found."
+  (when-let ((project (project-current nil dir)))
+    (project-files project)))
+
+(defun gptel-context--skip-p (file)
+  "Return non-nil if FILE should not be added to the context."
+  (when (and gptel-context-restrict-to-project-files
+	     (not (file-remote-p file)))
+    (when-let* ((project (project-current nil file)))
+      (not (member file (gptel-context--get-project-files (project-root project)))))))
```
vermiculus, line 279: You're calling project-current twice here, I think. Is this value cached by project.el?

gptel-context.el
```diff
@@ -195,38 +211,85 @@ Return PATH if added, nil if ignored."
...
+
+(defun gptel-context--message-skipped (file &optional project-files)
+  "Message that FILE is skipped because it is not a project file.
+PROJECT-FILES is a list of all files in the current project."
+  (let* ((type (if (file-directory-p file) "directory" "file"))
+	 (var (symbol-name 'gptel-context-restrict-to-project-files))
+	 (reminder (format "To include this file, unset `%s'." var)))
```
vermiculus, line 286: You can use %S instead of %s to simplify here.

vermiculus (changes requested):
Took a pass here, though I'm just on my phone so sorry for formatting.

How does this perform functionally for you? Any noticeable delay with larger/smaller projects with more/fewer ignored build artifacts?

pabl0:
I see some whitespace changes that apparently just change spaces to TABs, most likely since emacs has `indent-tabs-mode` on by default, and especially if you have `aggressive-indent-mode` on, it can cause unecessary changes in places you did not even meant to touch.  (This is why I submitted #712.) In case you remove those changes, consider doing some squashing as well, this PR has currently 34 commits and I think some of them could be squashed together to make the git history cleaner.

gptel-context.el
benthamite: Apologies, I’m not sure I understand. Can you elaborate?

gptel-context.el
benthamite: Thanks, done.

gptel-context.el
benthamite: The indentation is correct, but you may be referring to the phenomenon @pabl0 [mentioned](https://github.com/karthink/gptel/pull/665#issuecomment-2757487447), caused by having both `aggressive-indent-mode` and `indent-tabs-mode` on, which I’ll fix.

benthamite:
I think I'm basically done with the changes here. My only concern is that `(member file gptel-context--project-files)` may be a bit slow for projects with lots of files. I believe this issue arose after we started using `project.el`, and happens because `(member file gptel-context--project-files)` has to linearly search through the entire list of project files to check if a file is present. Claude suggests using a hash table, but I don’t know enough Elisp to implement this suggestion.

If and when @karthink agrees to merge this PR, I'm happy to do some squashing to clean up the commit history.

benthamite:
(If the performance issue is not address, I am now inclined to set the default value of `gptel-context-restrict-to-project-files` to `nil`.)

komikat:
Hi @karthink, and great work @benthamite on this PR!

This feature is excellent! For me, it addresses a pressing need, as it was the first thing I looked for when I started adding projects to the context.

@karthink, is there an estimated timeline for merging this? I'd be happy to contribute by implementing the hash-table optimisation that was discussed, if that would help move this PR forward towards main.

karthink:
> @karthink, is there an estimated timeline for merging this? I'd be happy to contribute by implementing the hash-table optimisation that was discussed, if that would help move this PR forward towards main.

I've just been busy with other features (and the new release) is all.  I had some performance concerns about this PR, but they might be fixed now.  I need to check.

Is adding entire projects to the context even viable?  It seems like a lot, not to mention inefficient.  Maybe I'm just behind in my thinking, context windows have been increasing recently.

benthamite:
FWIW, I do agree it’s generally inefficient. My workflow for integrating AI with projects has changed since submitting this PR, and now I almost always add selected project files rather than all files in a project. Still, I guess that *if* one wants to do the latter, one typically wants to exclude gitignored files, so I think this is probably still a useful feature to have (modulo the performance concerns, which I agree are valid).

peterbecich:
> Still, I guess that if one wants to do the latter, one typically wants to exclude gitignored files

Agreed, excluding the `.gitignore`d files would be a very useful feature in my opinion

ChetanKoneru:
Waiting on this feature. Thank you.

benthamite:
@karthink, shall we make a decision regarding this? If you don’t want to merge it, that’s totally fine by me, but I would prefer to close it so that people are not kept waiting.

karthink:
> @karthink, shall we make a decision regarding this? If you don’t want to merge it, that’s totally fine by me, but I would prefer to close it so that people are not kept waiting.

I wanted to merge this today, but the cache check is too fragile.  Here is an example of the kind of bug you can produce:

1. Run `M-x gptel-context-add-file` and navigate to a file from _another_ project.  Add it.
2. Now add a gitignored file from the current project.  It will be added too.

karthink:
Changed the caching logic, fixed some bugs and merged.  Thanks for the PR @benthamite.

Issue #503, opened 2024-12-11
Remove unnecessary org mode headers and properties in prompts sent


aGolduck:
Org mode headers and properties are sent in promts, they seemed unnecessary, for example, this is a request body

```
{
  "model": "gpt-4",
  "messages": [
    {
      "role": "system",
      "content": "You are a large language model living in Emacs and a helpful assistant."
    },
    {
      "role": "user",
      "content": "* another test\n:PROPERTIES:\n:GPTEL_TOPIC: another-test\n:END:\n\n** this is a test"
    }
  ],
  "stream": true,
  "temperature": 1.0
}
```

karthink:
Thanks for the suggestion -- this is a frequent one!  See #141 and many discussions in the discussions page.

This is a low-medium priority issue for me right now as this is basically harmless, except for the negligible token cost of few dozen extra tokens.   There are several semi-broken things I need to fix first -- I'll get to this eventually.

hrehfeld:
not filtering properties gets more entertaining if you ever decide to use parts of a response as your system prompt, because you have '(gptel response) text-properties in the property:

``` llm.org
# -*- eval: (gptel-mode); -*-
:PROPERTIES:
:GPTEL_MODEL: gpt-4o
:GPTEL_BACKEND: ChatGPT
:GPTEL_SYSTEM: file level system prompt
:END:
* technical questions
:PROPERTIES:
:CREATED:  [2025-01-15 00:53:18]
:GPTEL_SYSTEM: technical questions system prompt
:END:
** a technical question
:PROPERTIES:
:CREATED:  [2025-01-15 00:53:40]
:_GPTEL_TEXT_PROPS: this should be assistance voice
:END:

```

``` json
{
  "gptel": "request body",
  "timestamp": "2025-01-15 00:58:56"
}
{
  "model": "gpt-4o",
  "messages": [
    {
      "role": "system",
      "content": "technical questions system prompt"
    },
    {
      "role": "user",
[...]

Issue #328, opened 2024-06-18
Evaluate elisp/shell in prompts


osheari1:
Currently prompts can only be static. Adding syntax that would evaluate an elisp statement / shell commands would allow dynamic prompting.

#### Data extraction from local machine

Gives the ability to load data from a shell command, local file, ect, into the prompt without copy paste.

```
# prompt
Format the contents of a directory as a table of contents in markdown: (shell-command-to-string "ls") 
```
or
```
# prompt
Format the contents of a directory as a table of contents in markdown: $(ls)
```

#### Prompt organization
(technically the same thing as the previous) Would allow organization of long complex prompts

```
# Directory structure
# TODO-1.org
# TODO-2.org
# TODO-3.org

# Prompt
You are a careful programmer. Complete all TODO instructions. First $(cat TODO-1.org), then $(cat TODO-2.org), then $(cat TODO-3.org)
```

### Example from Alternative
[shell_gpt](https://github.com/TheR1D/shell_gpt)

```bash
>>> sgpt "format the contents of a directory as a markdown table of contents: $(ls)"
- [README.md](README.md)
- [some-file.md](some-file.md)
- [documents](documents)
```

karthink:
Thanks, this is a neat idea.  I'll look into adding something like this.

nfedyashev:
Super excited about this
Another probable use case:
elisp function that reads database schema and the injects it into the prompt, "given the following DB schema \n<schema>...</schema> write SQL migration that does X"

karthink:
This is still on the docket -- I just need to fix a few other things in gptel before I can get to this.

karthink:
> I just need to fix a few other things in gptel before I can get to this.

"just", "few other things"...

Well this is now implemented via `gptel-prompt-filter-hook`.  You can use it to transform the buffer text however you'd like.  To evaluate shell commands like `$(ls)`, for example, you can do this:

``` emacs-lisp
(defun gptel-interpolate-sh ()
  "Replace $(cmd) with the output of shell command cmd."
  (while (re-search-backward "\\$(\\([^)(]+?\\))" nil t)
    (let ((output (shell-command-to-string
                   (match-string-no-properties 1))))
      (delete-region (match-beginning 0) (match-end 0))
      (insert output))))

(add-hook 'gptel-prompt-filter-hook #'gptel-interpolate-sh)
```

You can modify the function to handle errors better, also evaluate Elisp etc.

PR #675, opened 2025-03-01
gptel-org-hide #503

hrehfeld:
re: #503 I've been using this for a few weeks with great success (minus todays rebase onto latest master and the if-let change, haven't tested that yet.)

Please check the FIXME question for variable naming.

0. `gptel-org--parse-buffer` wraps gptel--parse-buffer
1. there's a new hook `gptel-org-before-parse-buffer-hook`
2. within that hook, there are currently two functions:
   - `gptel-org--delete-file-local-variable-prop-line` deletes the file local prop manually/specially handled.
   - `gptel-org--delete-regex-alist` replaces matches of regexes with the given replacement (or "") in the new variable `gptel-org--delete-regex-alist` that currently just contains `org-property-drawer-re . nil`
 
So the user can either modify the hook to disable removal (or we can change the default to empty) or add their own functions. Or, for simplicity, just add regexes to the variable.

Now that I think about it: does it make sense to have this within non-org buffers as well? Start with an empty hook, and just add functions to it buffer-locally in e.g. org-mode.

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -167,6 +167,41 @@ current heading and the cursor position."
                                  50))))))
   (when (stringp topic) (org-set-property "GPTEL_TOPIC" topic)))
 
+(defcustom gptel-org-hide t
+  "Hide property drawers when creating a conversation prompt.
+
+When non-nil, property drawers are removed from the text sent to the LLM."
+  :type 'boolean
+  :group 'gptel)
+
+(defcustom gptel-org-hide-re-list '(org-property-drawer-re)
+  "List of regular expressions to hide when creating a conversation prompt.
+
+When `gptel-org-hide' is non-nil, these are
+applied the prompt text before sending it to the LLM."
+  :type '(repeat regexp)
+  :group 'gptel)
+
+
+(defun gptel-org--parse-buffer (backend max-entries)
+  "Exactly like `gptel--parse-buffer`, but apply some filters on the org buffers first"
+  (if gptel-org-hide
+      (let ((org-buffer (current-buffer)))
+        (with-temp-buffer
+         ;; save-current-buffer
+         ;;  (set-buffer (get-buffer-create " *gptel-org-temp*" t))
+          ;; (erase-buffer)
+          (insert-buffer-substring org-buffer)
+          (cl-loop
+            for re in gptel-org-hide-re-list
+            do (save-excursion
+            (goto-char (point-min))
+            (let ((case-fold-search nil))
+              (while (re-search-forward org-property-drawer-re nil t)
+                (replace-match "")))))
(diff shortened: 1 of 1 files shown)


hrehfeld:
(note: I haven't tested actually replacing with replacement)

psionic-k:
If this is an option, I would bet every use of the option is unconditional.  Keywords and drawers are both primarily noise in chat, never content.  If it's not conditional, we don't even need a variable.  `gptel-org` is just incomplete and needs to be finished.

A single hook could exist and the user can decide based on `major-mode` what to do, but are these not still just frontend behaviors that `gptel-org` and `gpte-md` can do unconditionally?

On the surface, it seems nice to respect each other's use cases, but I think this case is just "we forgot" and not "the user might not want it."

karthink:
I've added `gptel-org-ignore-elements` to remove unwanted Org elements from Org buffers before sending the prompt.

Currently `property-drawer` elements (`:PROPERTIES:...:END:`) are removed by default, but I might turn this off depending on how much confusion it causes.  Adding more unwanted elements to this list (like `comment` to remove comments) will work but can slow down `gptel-send`.

@hrehfeld the general idea of this PR is fine, but I think it's too structured and limiting.  Instead of specifying regexps, I would prefer a general `gptel-prompt-filter-hook` where the user can make whatever modifications are required in the (temp) buffer containing the prompt.  A typical example is #328.

hrehfeld:
> @hrehfeld the general idea of this PR is fine, but I think it's too structured and limiting. Instead of specifying regexps, I would prefer a general `gptel-prompt-filter-hook` where the user can make whatever modifications are required in the (temp) buffer containing the prompt. A typical example is #328.

If I understand you correctly, that's exactly what `gptel-org-before-parse-buffer-hook` is.

karthink:
Yes, it looks like they do similar things.  I've since added `gptel-prompt-filter-hook`.  The only functional differences between the current state of gptel and this PR are (i) the regexps in `gptel-org--delete-regex-alist` and (ii) the assertion for `(= (marker-position assert-marker) (point))`.  

Instead of regexps I'm using org-element and `gptel-org-ignore-elements`, and for anything more bespoke there's `gptel-prompt-filter-hook`.

I don't understand the need for the marker and the assertion, it seems like you can just wrap the `run-hooks` in `save-excursion`?

I think the two special functions (for the regexp-alist and the prop-line deletion) are handy and can be added to the [wiki](https://github.com/karthink/gptel/wiki), along with a page on prompt filtering. This PR can be closed.

karthink:
Closing as this functionality is covered by `gptel-org-ignore-elements` and `gptel-prompt-filter-hook`.

PR #677, opened 2025-03-02
Store some strings in variables in gptel-rewrite.el

rkallio:
Hi, thanks for creating this great package.

I've been exploring `gptel-rewrite`, and I primarily use language models in Finnish. I noticed that each rewrite request includes the prompt string "What is the required change?". This may cause issues, particularly with Google's Gemini, which sometimes identifies itself as an English model and declines to process Finnish text. To address this, I've modified the code to make this string configurable. Additionally, I've made the "Rewrite: " string configurable as well, as I find it useful to customize. I'm willing to contribute these changes, so please let me know what adjustments are needed for them to be accepted.  I feel the change is necessary to support non-english use cases.

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -29,6 +29,10 @@
 (defvar eldoc-documentation-functions)
 (defvar diff-entire-buffers)
 
+(defvar gptel-rewrite--assistant-instruction-question
+  "What is the required change?")
+(defvar gptel-rewrite--default-rewrite-prefix "Rewrite: ")
+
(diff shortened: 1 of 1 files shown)


pabl0:
This could be a good idea. However, if these are user-customizable variables, it may be more appropriate to define them with `defcustom` instead of `defvar`. Additionally, a two-dash prefix typically indicates internal variables rather than user-facing options.

rkallio:
@pabl0 can you take another look?

PR #685, opened 2025-03-06
Implement a `gptel-define-tool` macro

swflint:
Based on discussion in #514 (https://github.com/karthink/gptel/issues/514#issuecomment-2702241665 through https://github.com/karthink/gptel/issues/514#issuecomment-2702399246), this PR adds a macro to define tools in one single form, simplifying their creation.

As of now, this PR is very much a draft, as noted, I will need to get debug and edebug working, but I also will need to finish the docstring.  That said, I would appreciate other feedback on the macro and its developer experience.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1602,6 +1602,59 @@ implementation, used by OpenAI-compatible APIs and Ollama."
                     :additionalProperties :json-false))))))
     (ensure-list tools))))
 
+(cl-defmacro gptel-define-tool (function-name
+                                (&rest args-list)
+                                (&key name
+                                      description
+                                      async
+                                      category
+                                      confirm
+                                      include
+                                      &allow-other-keys)
+                                &body body
+                                &aux
+                                (docstring (when (stringp (car body))
+                                             (car body))))
+  "Define an LLM-callable tool, FUNCTION-NAME.
+
+TODO: Write docstring
+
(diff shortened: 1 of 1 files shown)


karthink:
@swflint I'm busy with other parts of gptel presently but I popped in to say that this is a great idea.  I second @psionic-k's suggestion in [this comment](https://github.com/karthink/gptel/issues/514#issuecomment-2702352927) about the macro accepting keywords directly.

I also think the naming is going to be confusing -- there will be a `gptel-make-tool` and a `gptel-define-tool` and users will have to check both out to figure out which one to use.  I don't yet know of a good solution to this.

As for this comment:

> This could make it easier to write without repeating too much, and avoid the use anonymous functions, making debugging easier.

You don't have to use anonymous functions?  You can provide a regular named elisp function to the tool.  This design was intentional, so you can reuse regular elisp functions as LLM tools if required.  See for instance the yt-dlp tool in my [opening comment](https://github.com/karthink/gptel/issues/514#issue-2754538957).

That said a macro to define an elisp function and a gptel tool together like this is welcome.

psionic-k:
@swflint  if you PR the [experimental branch of my fork](https://github.com/positron-solutions/gptel/tree/experimental), we can develop this independently.

Just one macro should be able to handle the existing keyword-only implementation while recognizing several "shorthand" forms by their unique content.  That is how transient's definers work.

PR #695, opened 2025-03-12
gptel-transient: Fix two bugs in model selection (`-m')

### LINKED ISSUE #693: gptel seems to write to user's history.el and breaks transient for other packages (magit, rg, etc)

### Bug Description

I have been experiencing this issue from time to time and I may be wrong about whether it is an issue or not.  

I have two models registered in my config like below
```elisp
(gptel-make-openai "Deepseek"
  :host "api.deepseek.com"
  :endpoint "/chat/completions"
  :stream t
  :key (lambda () (password-store-get "Dev/deepseek"))
  :models '(deepseek-chat deepseek-reasoner))

(setq gptel-default-mode 'markdown-mode
      gptel-prompt-prefix-alist
      '((markdown-mode . "## ") (org-mode . "** ") (text-mode . ">> "))
      gptel-backend (gptel-make-anthropic "Claude"
                      :stream t
                      :models '(claude-3-7-sonnet-20250219
                                claude-3-5-sonnet-20241022)
                      :key (lambda () (password-store-get "Dev/claude-key1"))))
```

The default model is Claude and I can switch to DeepSeek using gptel-menu.  The problem is, switching models in this sway seems to cause gptel to write something like below into my `~/.emacs.d/transient/history.el`:

```elisp
((gptel--infix-provider #s(gptel-anthropic "Claude" "api.anthropic.com" #<subr 616d626732d616e6f6e796d6f6c1_anonymous_lambda_23> "https" t "/v1/messages"
                                           #[nil
                                             ((password-store-get "Dev/claude-key1"))
                                             (t)]
                                           (claude-3-7-sonnet-20250219 claude-3-5-sonnet-20241022)
                                           "https://api.anthropic.com/v1/messages" nil nil))  
 (gptel-menu nil)
 (gptel-system-prompt nil)
 (magit-branch nil)
 (magit-commit nil)
 (magit-dispatch nil)
 (magit-log
  ("-n256" "--graph" "--decorate"))
 (magit-merge
  ("--ff-only"))
 (magit-pull nil
             ("--rebase"))
 (magit-push nil
             ("--force-with-lease"))
 (magit-rebase nil)
 (magit-remote
  ("-f"))
 (magit-reset nil)
 (magit-stash nil))

```
And this breaks transient for many packages that rely on it. A common situation I often run into is, after switching a model, I cannot use magit and get the following error from Emacs:
```
Debugger entered--Lisp error: (invalid-read-syntax "#<" 2 74)
  read(#<buffer  *temp*>)
  transient--read-file-contents("~/.emacs.d/transient/history.el")
  (defvar transient-history (transient--read-file-contents transient-history-file) "History of transient commands and infix arguments.\nThe value of this variable persists between Emacs sessions\n(unless `transient-save-history' is nil) and you usually\nshould not change it manually.")
  require(transient)
  byte-code("\300\301!\210\300\302!\210\300\303!\210\300\304!\210\300\305!\210\300\306!\210\300\307!\210\310\311\312\313\314DD\315\316\317\320\321\322\323\324\325\326D&\13\210\310\327\312\313\330DD\331\316\332\320\321\322\323\324\333\334D&\13\210\310\335\312\313\336DD\337\316\340\320\341\322\323\324\333C&\13\210\310\342\312\313\343DD\344\316\345\320\341\322\323\324\346&\13\210\310\347\312\313\350DD\351\316\352\320\353\322\354\355\356D\355\357D\355\360D\355\361D\355\362D\355\363D\364BBBBBBB&\11\210\310\365\312\313\366DD\367\316\370\320\353\322\323\371\372\324\373C&\15\210\310\374\312\313\375DD\376\316\377\320\353\322\323\371\372\324\201@\0C&\15\210\310\201A\0\312\313\201B\0DD\201C\0\316\201D\0\320\353\322\354\355\201E\0D\201F\0BB&\11\210\310\201G\0\312\313\201H\0DD\201I\0\316\201J\0\320\353\322\201K\0&\11\210\310\201L\0\312\313\201M\0DD\201N\0\316\201O\0\320\353\322\201P\0&\11\210\310\201Q\0\312\313\201R\0DD\201S\0\316\201T\0\320\353\322\354\355\201U\0D\355\201V\0D\355\201W\0D\201X\0BBBB&\11\210\310\201Y\0\312\313\201Z\0DD\201[\0\316\201\\\0\320\353\320\201]\0\320\201^\0\320\201_\0\322\201`\0&\17\210\310\201a\0\312\313\201b\0DD\201c\0\316\201d\0\320\353\320\201]\0\320\201^\0\320\201_\0\322\201e\0&\17\207" [require magit-base magit-git benchmark browse-url format-spec help-mode transient custom-declare-variable magit-mode-hook funcall function #f(compiled-function () #<bytecode 0x18c0f2ca42814>) "Hook run when entering a mode derived from Magit mode." :package-version (magit . "3.0.0") :group magit-modes :type hook :options magit-load-config-extensions bug-reference-mode magit-setup-buffer-hook #f(compiled-function () #<bytecode 0x186c47fcd2628>) "Hook run by `magit-setup-buffer'.\n\nThis is run right after displaying the buffer and right before\ngenerating or updating its content.  `magit-mode-hook' and other,\nmore specific, `magit-mode-*-hook's on the other hand are run\nright before displaying the buffer.  Usually one of these hooks\nshould be used instead of this one." (magit . "2.3.0") magit-maybe-save-repository-buffers magit-set-buffer-margin magit-pre-refresh-hook #f(compiled-function () #<bytecode 0x18c0f2ca42714>) "Hook run before refreshing in `magit-refresh'.\n\nThis hook, or `magit-post-refresh-hook', should be used\nfor functions that are not tied to a particular buffer.\n\nTo run a function with a particular buffer current, use\n`magit-refresh-buffer-hook' and use `derived-mode-p'\ninside your function." (magit . "2.4.0") magit-refresh magit-post-refresh-hook #f(compiled-function () #<bytecode 0x2eac6c7f89614>) "Hook run after refreshing in `magit-refresh'.\n\nThis hook, or `magit-pre-refresh-hook', should be used\nfor functions that are not tied to a particular buffer.\n\nTo run a function with a particular buffer current, use\n`magit-refresh-buffer-hook' and use `derived-mode-p'\ninside your function." (magit . "2.4.0") (magit-auto-revert-buffers magit-run-post-commit-hook magit-run-post-stage-hook magit-run-post-unstage-hook) magit-display-buffer-function #f(compiled-function () #<bytecode 0x18c0a51d70b14>) "The function used to display a Magit buffer.\n\nAll Magit buffers (buffers whose major-modes derive from\n`magit-mode') are displayed using `magit-display-buffer',\nwhich in turn uses the function specified here." (magit . "2.3.0") magit-buffers radio function-item magit-display-buffer-traditional magit-display-buffer-same-window-except-diff-v1 magit-display-buffer-fullframe-status-v1 magit-display-buffer-fullframe-status-topleft-v1 ...] 17)
  require(magit-mode)
  byte-code("\300\301!\210\300\302!\210\300\303!\210\300\304!\210\300\305!\210\300\306!\210\300\307!\210\310\311\312\313\314\315\314\316\314\317\320\321&\13\210\310\322\312\323\314\324\320\311&\7\210\310\325\312\326\320\311%\210\310\327\312\330\320\311%\210\310\331\312\332\320\311%\210\310\333\312\334\314\335\320\311&\7\210\310\336\312\337\314\340\320\311\320\333&\11\210\310\341\312\342\320\311\320\343&\7\210\344\341\345\346#\210\344\341\347\346#\210\310\350\312\351\320\311%\210\344\331\352\353#\210\344\341\354\353#\210\344\331\355\353#\210\344\341\356\353#\210\344\311\357\353#\210\344\341\360\353#\210\344\305\361\353#\210\310\362\312\363\314\364\314\365\314\366\314\367\320\311\320\350\320\322&\21\210\344\362\370\371#\210\344\372\370\371#\210\344\362\373\371#\210\344\362\374\371#\210\344\362\375\371#\210\376\377!\207" [require magit-base magit-git magit-mode magit-margin magit-process magit-transient magit-autorevert custom-declare-group magit nil "Controlling Git from Emacs." :link (url-link "https://magit.vc") (info-link "(magit)FAQ") (info-link "(magit)") :group tools magit-essentials "Options that every Magit user should briefly think about.\n\nEach of these options falls into one or more of these categories:\n\n* Options that affect Magit's behavior in fundamental ways.\n* Options that affect safety.\n* Options that affect performance.\n* Options that are of a personal nature." (info-link "(magit)Essential Settings") magit-miscellaneous "Miscellaneous Magit options." magit-commands "Options controlling behavior of certain commands." magit-modes "Modes used or provided by Magit." magit-buffers "Options concerning Magit buffers." (info-link "(magit)Modes and Buffers") magit-refresh "Options controlling how Magit buffers are refreshed." (info-link "(magit)Automatic Refreshing of Magit Buffers") magit-faces "Faces used by Magit." faces custom-add-to-group diff-refine-added custom-face diff-refine-removed magit-extensions "Extensions to Magit." git-commit custom-group git-commit-faces git-rebase git-rebase-faces magit-section magit-section-faces with-editor ...] 18)
  require(magit-core)
  byte-code("\300\301!\210\300\302!\210\300\303!\210\300\304!\210\300\305!\210\300\306!\210\300\307!\210\300\310!\210\300\311\312\313#\210\300\314!\210\315\316\317\320\321DD\322\323\324\325\326\327\330&\11\207" [require magit-core magit-diff magit-log magit-wip magit-apply magit-repos git-commit format-spec package nil t with-editor custom-declare-variable magit-openpgp-default-signing-key funcall function #f(compiled-function () #<bytecode 0x18c000ae69414>) "Fingerprint of your default Openpgp key used for signing.\nIf the specified primary key has signing capacity then it is used\nas the value of the `--gpg-sign' argument without prompting, even\nwhen other such keys exist.  To be able to select another key you\nmust then use a prefix argument." :package-version (magit . "4.0.0") :group magit-commands :type string] 10)
  interactive-form(magit-status)
  meow--keypad-execute(magit-status)
  meow--keypad-try-execute()
  meow--keypad-handle-input-event(117)
  meow--keypad-handle-input-with-keymap(117)
  meow-keypad-start-with(nil)
  meow-keypad()
  funcall-interactively(meow-keypad)
  command-execute(meow-keypad)
```

After removing the following part from history.el file
```
(gptel--infix-provider #s(gptel-anthropic "Claude" "api.anthropic.com" #<subr 616d626732d616e6f6e796d6f6c1_anonymous_lambda_23> "https" t "/v1/messages"
                                           #[nil
                                             ((password-store-get "Dev/claude-key1"))
                                             (t)]
                                           (claude-3-7-sonnet-20250219 claude-3-5-sonnet-20241022)
                                           "https://api.anthropic.com/v1/messages" nil nil))
```
The transient comes back to normal. I'm not sure if this is a bug or I'm simply using gptel in a wrong way. If I stick to the default model and never uses gptel-menu, then this problem will not occur. 

### Backend

None

### Steps to Reproduce

1. Config two models in emacs config file
2. Use gptel-menu to switch models
3. After a while, you may notice transient not working normally and `transient/history.el` file gets polluted

### Additional Context

Emacs version: 30.1
OS: Arch Linux

I have been experiencing this issue since Emacs 29.4. I'm currently on the following commit
```
e9d7903 master origin/master gptel-perplexity: reasoning and deep-research models (#664)
```

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

pabl0:
Attempt to fix #693.

Furthermore, add a default value to the reader to prevent `gptel-backend` from being `nil` when the user presses RET on an empty prompt.


karthink:
Modifying the history like this seems like a cursed solution, guaranteed to bite us in the back in the future when Transient updates.  I'd like to make the value of the provider infix be a string (the backend name or backend-name:model-name), then the problem should disappear for good.  Any idea if we can do that?  I haven't looked at this code in a while.

pabl0:
I agree it feels like a hack, OTOH not sure how it could break even more, since I don't think the actual semantics `completing-read` would change in Emacs. Perhaps it would be good idea to store the provider as a string but it's bit complicated right now.

Transient closed https://github.com/magit/transient/issues/172 without fixing transient.el.

Do you agree with the second fix? Now if the user does `C-u M-x gptel-send RET -m RET`, it causes the error "transient-setup: Wrong type argument: gptel-backend, nil" and there is no second chance to change the model using the transient menu. Not very likely scenario, but perhaps the user looks what available by hitting TAB and then decides that she is okay with the currently selected model after all, and assumes hitting RET to an empty prompt will just get her back to the menu without changing anything. That actually happened to me once, that's how I noticed this problem.

karthink:
Yeah the fix for the second issue looks fine at first glance, but I need to
think about it a bit.

For the main issue I have an idea for fixing it at the root, in the custom
transient class used by the infix.  Both the backend and model need to be
set by side effect (instead of just the model right now) and the main value
of the infix needs to be a string.  If that makes sense to you you can try
to fix it, or I can eventually.

We're close to releasing v0.9.8, I just want to make sure there are no
outstanding bugs.

karthink:
I fixed the main issue at the source just now.  Please test.  The commit message explains the fix.

Let's leave this PR open to fix the completing-read default value problem.

pabl0:
Okay this kinda works, but if you don't pass the history argument to `completing-read`, hitting <down> does not really go through the actual history elements but apparently all available completions? And if you do pass it, transient will store the history as strings instead of symbols. Somehow I get the feeling that transient don't like anything else than strings in the history.

karthink:
> but if you don't pass the history argument to completing-read, hitting the arrow down or M-n does not really go through the actual history elements but apparently all available completions?

Even if we did pass a history variable, I don't think it would work, would it?  What's being stored is in the history file is not what we read from `completing-read`.  To make it compatible the `gptel-provider-variable` class needs to be modified some more.  Its `value` should be exactly the string that's read by `completing-read`, and it should have two additional slots `:backend` and `:backend-value` for managing the actual backend structure.

I can merge this for now and add a TODO, but can you verify that the correct buffer-local value of the backend and model are being set if you press `RET` without a selection?  I remember having some trouble, with the values in the _transient buffer_ being chosen instead of the values in the main buffer, i.e. `transient--original-buffer`.

pabl0:
> Even if we did pass a history variable, I don't think it would work, would it? What's being stored is in the history file is not what we read from `completing-read`. To make it compatible the `gptel-provider-variable` class needs to be modified some more. Its `value` should be exactly the string that's read by `completing-read`, and it should have ~two~ four additional slots `:backend`, `:backend-value`, `model` and `model-value` for managing the actual backend structure and model.

Yeah, perhaps not worth the effort, at least as a v0.9.8 goal Though I can imagine history being somewhat useful when switching between couple of models.
 
> I can merge this for now and add a TODO, but first could you verify that the correct buffer-local value of the backend and model are being set if you press `RET` without a selection? I remember having some trouble, with the values in the _transient buffer_ being chosen instead of the values in the main buffer, i.e. `transient--original-buffer`.

Yes, if I understand this right, it restores the same model that is displayed in in the transient as the current model, even if I switch between buffers with different buffer-scope models. 

I was thinking if it should reset it to the standard or customized value of `gptel-model` instead, but somehow it does not feel quite right -- at least it should be somehow indicated in the prompt I think to avoid surprises.

Anyway, I doubt this is a very common use case, it is just a way of making sure the user won't end up in a state where the transient no longer can be accessed without manually changing the variable or restarting Emacs.

PR #696, opened 2025-03-12
NEWS: Use org comment

pabl0:
Use `#' instead of Lisp `;;;' for comment.  This way Github hides it, and it's more correct.

(_Bikeshedding:_ Or should we use `NEWS.org` instead, to be consistent with `README.org`? Also many projects like `transient.el` use `CHANGELOG`.)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,104 +1,203 @@
-;;; -*- mode: org; -*-
+# -*- mode: org; -*-
 
 * 0.9.8 2025-03-11
 
-Version 0.9.8 adds support for new Gemini, Anthropic, OpenAI, Perplexity, and Deepseek models, introduces LLM tool use/function calling, a redesign of ~gptel-menu~, includes new customization hooks, dry-run options and refined settings, improvements to the rewrite feature and control of LLM "reasoning" content.
+Version 0.9.8 adds support for new Gemini, Anthropic, OpenAI,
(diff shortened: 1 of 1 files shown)


pabl0:
Another potential bikeshedding idea: should the `NEWS` file use filled paragraphs? Not everyone enables `visual-line-mode` by default in Org mode, and shorter lines often improve readability. (The same applies to `README.org`.)

In general, it seems that filled paragraphs are more common in Org culture compared to Markdown.

karthink:
> (_Bikeshedding:_ Or should we use `NEWS.org` instead, to be consistent with `README.org`? Also many projects like `transient.el` use `CHANGELOG`.)

Emacs uses NEWS, Org mode uses ORG-NEWS, so I went with NEWS.  FWIW ELPA/package.el, Straight and Elpaca understand NEWS, NEWS.org, NEWS.md and CHANGELOG.

> Another potential bikeshedding idea: should the `NEWS` file use filled paragraphs? Not everyone enables `visual-line-mode` by default in Org mode, and shorter lines often improve readability. (The same applies to `README.org`.)

I don't want to touch README.org right now, but NEWS is a new file so let's do it.

Could you fill them a 70 chars (the Emacs default)?  You could also fix a typo near the top where I say Perplexity instead of Deepseek.

karthink:
Thank you @pabl0!

PR #698, opened 2025-03-12
* gptel-curl: Require gptel-org when needed for org transform (#694)

### LINKED ISSUE #694: Latest commit d9edbbc seems broken "gptel--stream-convert-markdown->org is void"
### Bug Description

While running commit d9edbbc i get the following message in the messages bufffer
```
gptel-curl-get-response: Symbol’s function definition is void: gptel--stream-convert-markdown->org
```
the last commit i was on 8889e27, works fine

### Steps to Reproduce

execute gptel in an orgmode buffer

pabl0:
FIxes #694 by requiring gptel-org when needed for  gptel--stream-convert-markdown->org'.

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -137,6 +137,7 @@ the response is inserted into the current buffer after point."
                            (when (with-current-buffer (plist-get info :buffer)
                                    (and (derived-mode-p 'org-mode)
                                         gptel-org-convert-response))
+			     (require 'gptel-org)
                              (gptel--stream-convert-markdown->org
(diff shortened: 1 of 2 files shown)


karthink:
There are several ways to solve this, and I think it's better to do it upstream of `gptel-curl-get-response`.

I've posted a question I could use some feedback on in #694.

karthink:
Fixed by delegating prompt creation for the selected region to gptel-org.

[v0.9.8 is up](https://github.com/karthink/gptel/releases/tag/v0.9.8).

PR #699, opened 2025-03-12
elpaignore: Add .github

pabl0:
Just some minor finishing touch before 0.9.8 release to elpa.nongu.org. 

MELPA does not include documentation (`README.org`, `NEWS`, `LICENSE`) by default, unlike (non-)GNUELPA.  But IMHO they are useful (and not litter, like `.github/ISSUE_TEMPLATE`)  and consume tiny amount of disk space, so I don't see a reason to exclude them.

### DIFF
--- a/.elpaignore
+++ b/.elpaignore
@@ -1,1 +1,3 @@
 test
+.github
+.elpaigore


karthink:
- I'm confused by what placing .elpaignore in .elpaignore does.
- I didn't understand how this PR relates to what MELPA includes or doesn't include.

pabl0:
> * I'm confused by what placing .elpaignore in .elpaignore does.

ELPA does not exclude the .`elpaignore` file itself from the package tarballs. Many projects (but definedly not all) seem to include it in `.elpaignore`. I just verified that installing v0.9.7 from elpa.nongnu.org does indeed leave the file under `.emacs.d/elpa/gptel-0.9.7`, so it is not striped during installing.

I'm not sure whether the ELPA team sees this as an issue or an intended part of the design. Should we discuss it on emacs-devel?

>     * I didn't understand how this PR relates to what MELPA includes or doesn't include.

It does not directly. But since ELPA and MELPA have the opposite logic, right now gptel installed from ELPA includes documentation files, while MELPA only contain `.el` files.

If we aim for identicality, we'd either have to drop the documentation with in `.elpaignore` or adjust the MELPA recipe to include them. However, MELPA advises against including additional files, a "cultural difference" we likely want to observe. In ELPA, it's common to include, however `LICENSE` and `COPYING` (etc) are mostly listed as `:ignored-files`. 

(I am probably just dumb, but I don't know where to find the `elpa-packages` file for nongnu, is it on a separate git repo or some branch?)

karthink:
> ELPA does not exclude the .`elpaignore` file itself from the package tarballs. Many projects (but definedly not all) seem to include it in `.elpaignore`. I just verified that installing v0.9.7 from elpa.nongnu.org does indeed leave the file under `.emacs.d/elpa/gptel-0.9.7`, so it is not striped during installing.

Ah, okay.  Thanks for the clarification.

> I'm not sure whether the ELPA team sees this as an issue or an intended part of the design. Should we discuss it on emacs-devel?

Maybe we can seek some quick verification from Philip Kaludercic that this is okay to add .elpaignore to .elpaignore?  Unless you are sure that it's okay, then I'll take your word for it.

>>     * I didn't understand how this PR relates to what MELPA includes or doesn't include.
>
> It does not directly. But since ELPA and MELPA have the opposite logic, right now gptel installed from ELPA includes documentation files, while MELPA only contain `.el` files.
>
> If we aim for identicality, we'd either have to drop the documentation with in `.elpaignore` or adjust the MELPA recipe to include them. However, MELPA advises against including additional files, a "cultural difference" we likely want to observe. In ELPA, it's common to include, however `LICENSE` and `COPYING` (etc) are mostly listed as `:ignored-files`. 

I don't think we need to aim for identical installs. 

> (I am probably just dumb, but I don't know where to find the `elpa-packages` file for nongnu, is it on a separate git repo or some branch?)

I don't know what you mean by the "elpa-packages file".  My only point of connection with nongnu-elpa is this page: https://elpa.nongnu.org/nongnu/gptel.html

pabl0:
> (I am probably just dumb, but I don't know where to find the `elpa-packages` file for nongnu, is it on a separate git repo or some branch?)

To be honest, I never even imagined I could be anything but dumb. The repo is of course in [git.savannah](https://git.savannah.gnu.org/cgit/emacs/nongnu.git) and here is the relevant part for gptel in `elpa-packages` (in the branch called `main`):

```elisp
 (gptel                 :url "https://github.com/karthink/gptel"
  :readme ignore)
```
I agree that ignoring the README file is a good idea, since the commentary section in `gptel.el` is more consise for this purpose. Of course the two are somewhat redundant.

pabl0:
> I agree that ignoring the README file is a good idea, since the commentary section in `gptel.el` is more consise for this purpose. Of course the two are somewhat redundant.

And to clarify, `:readme ignore` means the elpa.nongnu.org web site or the `describe-package` functions shows the commentary section from `gtpel.el`, but it still shipped under the `.emacs.d/elpa` directory for the user's local viewing pleasure.

pabl0:
> Maybe we can seek some quick verification from Philip Kaludercic that this is okay to add .elpaignore to .elpaignore? Unless you are sure that it's okay, then I'll take your word for it.

Well at least it should be safe and not break anything, since e.g. transient is doing it:

https://github.com/magit/transient/blob/main/.elpaignore

It seems that including `.gitignore` is unecessary, it get's excluded by default.

karthink:
> Well at least it should be safe and not break anything, since e.g. transient is doing it:

Works for me!

PR #702, opened 2025-03-14
Enable Markdown formatting for "o3-mini" by removing `nosystem` capability

Inkbottle007:
**Description:**  
Previously, the "o3-mini" model stopped accepting system prompts, leading to the introduction of the `nosystem` capability in gptel. However, OpenAI once again allows system prompts (with function equivalently to "developer messages") allowing to restore Markdown formatting by including "Formatting re-enabled" at the start of the system prompt.

Removing the `nosystem` capability from “o3-mini” allows gptel to once again send system prompts, and in turn re-enable Markdown output for users who prepend instructions like “Formatting re-enabled” to their prompts.

**Changes**  
- Remove/comment out the `nosystem` capability for “o3-mini.”  
- Allow system prompts (e.g., “Formatting re-enabled”) to restore Markdown formatting in responses.

**Testing**  
Verified locally by sending a system prompt containing “Formatting re-enabled” to “o3-mini,” which produced Markdown-formatted responses. This aligns with the [Azure OpenAI docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/reasoning?tabs=python-secure#markdown-output).

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -585,7 +585,7 @@ the same as t."
      :cutoff-date "2023-10")
     (o1
      :description "Reasoning model designed to solve hard problems across domains"
-     :capabilities (nosystem media reasoning)
+     :capabilities (media reasoning)
      :mime-types ("image/jpeg" "image/png" "image/gif" "image/webp")
      :context-window 200
      :input-cost 15
@@ -616,8 +616,7 @@ the same as t."
      :input-cost 3
      :output-cost 12
      :cutoff-date "2023-10"
-     :capabilities ( ;; nosystem
-                    reasoning)
+     :capabilities (reasoning)
      :request-params (:stream :json-false))
     ;; limited information available
     (gpt-4-32k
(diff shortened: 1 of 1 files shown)


gptel.el
```diff
@@ -616,7 +616,8 @@ the same as t."
      :input-cost 3
      :output-cost 12
      :cutoff-date "2023-10"
-     :capabilities (nosystem reasoning)
+     :capabilities ( ;; nosystem
```
karthink, line 619: You can delete it instead of commenting it out.

Inkbottle007: I’ve removed `nosystem` entirely for `o3-mini` and `o1` as suggested, but left it for `o1-mini` due to testing. Thanks!

karthink: What do you mean by "due to testing"?  Are you not sure if it supports system messages?

Inkbottle007: Sorry for the confusion! I tested all three models. Both `o3-mini` and `o1` accept system prompts successfully, but `o1-mini` rejects them in my tests. Accordingly, I removed `nosystem` for `o3-mini` and `o1`, but left it in place for `o1-mini`. Let me know if you'd like me to check or change anything else.

karthink:
Thanks for the PR @Inkbottle007!

PR #706, opened 2025-03-15
gptel: Add describe-gptel-model feature

pabl0:
The new `describe-gptel-model` command displays model information in the `*Help*` buffer in similar fashion as `describe-package`.  This naming convention aligns with Emacs Lisp coding conventions documented in the elisp info manual (I believe).   The command is accessible via `M-x` or the transient menu (bound to `h`).

Several new keys have been added to model definitions.  A future enhancement could be to add more details, such as cache-hit pricing and web URLs.

The model selection completion for the `-m` option could benefit from removing some details to improve readability, particularly on smaller screens, if you agree with this new feature.

### DIFF
--- a/gptel-anthropic.el
+++ b/gptel-anthropic.el
@@ -468,45 +468,58 @@ files in the context."
 ;;                   (t context-string)))
 ;;         (plist-get (car (last prompts)) :content)))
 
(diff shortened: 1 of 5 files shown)


pabl0:
Hmm I think I've misunderstood the streaming bit. It should be detected per-model from `:request-params` having `:stream`?
(#710)

PR #707, opened 2025-03-15
README: Tweak list of alternatives

pabl0:
Drop leafy-mode, it's gone 404 from Github.  Add Minuet, which is a new code-completion specific LLM client.


karthink:
Thanks @pabl0

PR #708, opened 2025-03-16
elpaignore: Oopsie daisy

pabl0:
I guess there is no change too small not to introduce a typo, sorry.


karthink:
At least it was a harmless typo.

PR #711, opened 2025-03-16
gptel: Ensure restoring state does not flag the buffer modified

pabl0:
* gptel.el (gptel--restore-props): Save `buffer-modified-p` before restoring properties and revert it to its original state.  This prevents a visited file from being marked as modified until the user actually makes changes.  This is particularly useful when multiple files are accessed automatically, such as from the `org-agenda-files` location (otherwise when exiting Emacs the user is asked about each file if it needs to be saved).

* gptel-org.el (gptel-org--restore-state): Likewise for Org.

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -451,28 +451,30 @@ ARGS are the original function call arguments."
 (defun gptel-org--restore-state ()
   "Restore gptel state for Org buffers when turning on `gptel-mode'."
   (save-restriction
-    (widen)
-    (condition-case status
-        (progn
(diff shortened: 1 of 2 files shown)


karthink:
I can see this creating some edge case issues involving file-locking in the future, but we'll revisit that when it crops up.

Thanks for the PR @pabl0!

pabl0:
Of course I just learned about the macro `with-silent-modification' that is the correct way of doing this, I suppose.
It might help with potential locking issues, but have to look into it in more detail.

karthink:
Yeah that would be the better way, but we can leave it be for now

PR #712, opened 2025-03-16
dir-locals: Set some project specific local variables

pabl0:
.dir-locals.el: Set `indent-tabs-mode` to nil to avoid whitespace problems (particularly with `aggressive-indent-mode`), as the project predominantly uses spaces for indentation.  Ensure sentences end with double spaces to adhere to the house style.  Point `bug-reference-url-format` to Github issues; while there are few bug references in the codebase, this may be useful later.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,3 @@
+((nil . ((indent-tabs-mode . nil)
+	 (sentence-end-double-space . t)
+	 (bug-reference-url-format . "https://github.com/karthink/gptel/issues/%s"))))


benthamite:
Perhaps we could also set `fill-column` to its default value of `70`? Some users (myself included) have a different value set, which means that unless they always remember to call `set-fill-column` before filling a paragraph, the file will not be filled consistently.

karthink:
@pabl0 are any of these going to cause the the verification dialog to pop up?

pabl0:
All the variables are designated as safe file-local variables, which
means there is no annoying verification dialog.

Do you have opinion about adding (emacs-lisp-docsting-)-fill column?

karthink:
Thanks for the PR @pabl0!

`emacs-lisp-docstring-fill-column` and `fill-column`: We can set them to the default values for Emacs 30, so 72 and 70 respectively.

PR #715, opened 2025-03-17
README: Clarify backend configuration and registration

spudlyo:
Reading #275, I noticed that there was some confusion around how to use `gptel-api-key` and if you had to provide your own function for reading the keys out of `auth-sources`. I thought I'd take a shot at updating the documentation to make it a bit clearer on how this works.

### DIFF
--- a/README.org
+++ b/README.org
@@ -89,6 +89,7 @@ gptel uses Curl if available, but falls back to the built-in url-retrieve to wor
       - [[#spacemacs][Spacemacs]]
   - [[#setup][Setup]]
     - [[#chatgpt][ChatGPT]]
+    - [[#configuring-other-llm-backends][Configuring other LLM backends]]
     - [[#other-llm-backends][Other LLM backends]]
       - [[#azure][Azure]]
       - [[#gpt4all][GPT4All]]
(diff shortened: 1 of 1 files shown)


karthink:
To be clear, I would like this information in the README.  I just think it has to be placed in a way that tricks the user into reading it.

spudlyo:
You're probably right that most users aren't going to know or care about `auth-sources`, but if we can save one or two hapless fools from hardcoding their OpenAI api keys into their `init.el` files and then checking them into GitHub we've done the world a mini-service ;)

To that end I've tightened up the language. One thing I think adding this section does is to help people understand what it means to have multiple backends, i.e. they'll have more choices in the transient menu. We could also explain how to set the default backend in this section as well, but I thought we'd keep it as short as possible. I also changed the ChatGPT section so it mentions the relationship between `gptel-api-key` and `gptel-apikey-from-auth-source` but postpones the details of how `~/.authinfo` works until the next section. My guess is that people will begin to care more about `auth-sources` the more backends they configure.

Hopefully this tightened up language will make your eyes gloss over less ;)

karthink:
Thanks @spudlyo, this looks like something I would skip over with much less annoyance!  To my detriment, of course.

karthink:
I'm considering hiding the authinfo instructions again, see #736.  It's just too confusing for users to set up both gptel and authinfo unless they're already familiar with authinfo... for whom the authinfo tutorial is superfluous anyway.

PR #721, opened 2025-03-19
gptel-rewrite: Make rewrite work on Emacs < 29 again

pabl0:
* gptel-rewrite.el (gptel--rewrite-dispatch): The function `rmc--add-key-description` was introduced in Emacs 29, resulting in an error for `gptel-rewrite`: "mapcar: Symbol’s function definition is void: rmc--add-key-description".  For older supported Emacs versions, skip this part to restore rewrite functionality, albeit with slightly reduced user guidance.  The actual rmc menu remains accessible in the minibuffer.


karthink:
Thanks @pabl0!

PR #727, opened 2025-03-20
gptel-context: Improve context management of buffers

pabl0:
* gptel-context.el (gptel-context-add-buffer): Introduce new function for adding a buffer (object or name).  This is useful for incorporating buffers from Lisp code in addition to interactive usage. (gptel-add-buffer): Create an alias (similar to `gptel-add-file`). (gptel-context-add): Utilize the new function.  Add support for adding/removing items in the Buffer List menu.
(gptel-context-remove): Add functionality to remove (entire) buffers from context (used when removing from the Buffer List).

### DIFF
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -97,6 +97,9 @@ context chunk.  This is accessible as, for example:
   CONFIRM is non-nil. With negative prefix ARG, remove all files from
   the context instead.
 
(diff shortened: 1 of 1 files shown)


pabl0:
Okay,  @Stebalien implemented roughly the same functionality in 51fa575.  So I think this can be closed.

PR #731, opened 2025-03-22
gptel-rewrite: Diff buffers, not files

pabl0:
* gptel-rewrite.el (gptel--rewrite-diff): Show the difference between buffers rather than file names.  Otherwise, if a) the buffer has never been saved, the error "No such file or directory" will occur, or b) if there are unsaved changes in the buffer, the diff output will not reflect those changes.


karthink:
Thanks @pabl0!  I don't remember why I set it up to look for a file -- it was definitely deliberate considering the trouble I went to there instead of just calling `diff-no-select` the usual way.  If something breaks I guess we'll find out.

PR #732, opened 2025-03-22
README: Clarify installation instructions

pabl0:
With default Emacs configuration, latest stable (tagged) version of gptel is available for installation from NonGNU ELPA, so it is not the latest commit that will be installed if the user has not modified `package-archives`.

README.org: Make it more clear how to install stable or latest commit, and mention `use-package`.


karthink:
This is quite confusing to read now.  How about something like:

```
** Installation

- *Release version*: =M-x package-install= ⏎ =gptel= in Emacs.
- *Development snapshot*: Add MELPA or NonGNU-devel ELPA to your list of package sources, then install with =M-x package-install= ⏎ =gptel=.
```

The mention of use-package is misleading as use-package doesn't install packages, so I'd drop it.

pabl0:
> This is quite confusing to read now.

I agree, pushed a new version that's hopefully more clear.

> The mention of use-package is misleading as use-package doesn't install packages, so I'd drop it.

Sure, you need to add `:ensure t` in order to do that. In any case, it is somewhat unnecessary here. If you are familiar with `use-package`, you probably don't need detailed instructions in the README file.

karthink:
Thanks @pabl0, this looks good.

Issue #730, opened 2025-03-21
gptel-rewrite: Asks for confirmation with kill-buffer


pabl0:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

When using `gptel-rewrite` in `message-mode` (sending email from Gnus), and probably in some other situations as well, after the rewrite response is ready, Emacs will ask:

`Buffer *gptel-rewrite* modified: kill anyway? (yes/no/save and then kill)`

I can answer "yes" and after that it works okay, but this is confusing and annoying.

From debugging, I believe it happens here:

https://github.com/karthink/gptel/blob/af4ddb1dffe96461501c08d061d51ac2dbe99c77/gptel-rewrite.el#L417C1-L431C36

If I add `(set-buffer-modified-p nil)` before `(kill-buffer proc-buf)` it works as expected, but I am unsure if this is the correct fix. Also, I don't quite understand why this happens with Message buffers specifically.

### Backend

None

### Steps to Reproduce

1. Start sending an email (`C-x m`)
2. Type in some nonsense and select the text.
3. `M-x gptel-rewrite RET r`

### Additional Context

Emacs 30.1 on linux.

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

karthink:
Ugh, gnus can be overbearing in how it modifies Emacs' global state.

Something is setting up a global hook to query _any_ buffer via `kill-buffer-query-functions`, and this is spilling over into temp buffers like `*gptel-rewrite*`.  Anyway, I think it's handled now.  Please update and test.

pabl0:
> Something is setting up a global hook to query _any_ buffer via `kill-buffer-query-functions`, and this is spilling over into temp buffers like `*gptel-rewrite*`. Anyway, I think it's handled now. Please update and test.

Thanks` for looking into this. However, this patch does not seem to change behavior for me. 

I think it must be something else than the `kill-buffer-query-functions` hook, since if I set up a debugger break point just before `(kill-buffer)` and inspect the value, it remains as the default value, which is `process-kill-buffer-query-function`. It's a built-in from `substrl.el` that checks for running processes and asks for confirmation if necessary, 

However, it seems `kill-buffer-hook` has the value `(t mml-destroy-buffers)`, when in the unsent mail folder. However in the debugger break point it is the global value `(browse-url-delete-temp-file uniquify-kill-buffer-function
[...]

PR #733, opened 2025-03-22
gptel-rewrite: Ensure kill-buffer does not ask for confirmation

pabl0:
I think this is a reasonable workaround for #730.


karthink:
This is quite confusing to read now.  How about something like:

```
** Installation

- *Release version*: =M-x package-install= ⏎ =gptel= in Emacs.
- *Development snapshot*: Add MELPA or NonGNU-devel ELPA to your list of package sources, then install with =M-x package-install= ⏎ =gptel=.
```

The mention of use-package is misleading as use-package doesn't install packages, so I'd drop it.

karthink:
This will do for now, I guess. Thanks.

Issue #725, opened 2025-03-20
Intermittent `invalid_request_error` and "Unexpected tool_call_id" when using tools with Sonnet 3.7


kmontag:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

When using tools with Claude Sonnet 3.7 (and possibly others, I haven't tested), requests sometimes fail with an error like this in the logs:

```
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "messages.2.content.0.tool_use.args: Extra inputs are not permitted"
  }
}
```

And a corresponding error is always echoed to the messages buffer like:

```
Unexpected tool_call_id format: 017S7rrJajodRVr9sbKE9Hby
```

Frustratingly, this happens quite frequently (half the time or so) on a more complex `gptel-request` flow that I've implemented, but it's much less common during simpler requests, so it's a bit tough to isolate anything for debugging. I'm sometimes able to reproduce it with the following:

```elisp
;; -*- lexical-binding: t; -*-
;;
;; To test use `emacs -q --load [filename]` and run `my/gptel-query'.
;;
;; Messages like "Unexpected tool_call_id format: 017S7rrJajodRVr9sbKE9Hby" will sometimes be
;; printed, and corresponding 'invalid_request_id' errors will appear in the gptel log.


;; The tool_call_id errors show up in the messages buffer.
(setq inhibit-splash-screen t)
(switch-to-buffer "*Messages*")

(require 'use-package)
(use-package gptel
  :ensure t
  :demand t
  ;; For example.
  :load-path "elpaca/repos/gptel/"
  :custom
  (gptel-log-level 'info)
  (gptel-use-tools t)
  :config
  (require 'gptel-anthropic)
  (require 'gptel-curl)
  (gptel-make-anthropic "Claude" :stream t :key (getenv "ANTHROPIC_API_KEY"))

  (setq
   gptel-backend (gptel-get-backend "Claude")
   gptel-model 'claude-3-7-sonnet-20250219)

  (setq gptel-tools
        (list
         (gptel-make-tool
          :name "get_random_word"
          :function
          (lambda ()
            (let* ((words
                    '("elephant" "sunshine" "keyboard" "labyrinth" "galaxy" "velvet" "harmony"))
                   (word-count (length words))
                   (random-index (random word-count)))
              (nth random-index words)))
          :description "Get a random word from a predefined list of words.")

[...]

PR #734, opened 2025-03-23
prevent invalid duplication of tool_call messages

tcahill:
I believe this fixes https://github.com/karthink/gptel/issues/725

When a `message_delta` event is received, we append any tool calls from the response to the global vector of messages and then modify the `tool-use` lists in the FSM info to put them in the right shape for actually invoking tools. This can cause issues if an error occurs after those modifications have already been made - if the `gptel--json-read` call for the `message_delta`'s data fails because we don't yet have the full object, the error handler silently discards the error and resets the point to the start of the same event again. Next time we re-enter `gptel-curl--parse-stream`, we'll again try to process the `message_delta` event but because we modified the tool use blocks in the last invocation, we'll push duplicate messages with the incorrect format (`input` is empty and `args` has the contents that `input` is supposed to).

For example, here's what the messages vector looks like when we encounter this issue in the repro script described by @kmontag in https://github.com/karthink/gptel/issues/725:

```
        [(:role "user" :content
                "Open the file '~/example.el' and make 5 changes of your choosing. Make each change individually. Don't ask for clarification.")
         (:role "assistant" :content
                [(:type "text" :text
                        #("I'll open the file and make 5 individual changes of my choosing. Let me first check the contents of the file."
                          0 9 (gptel response front-sticky (gptel)) 9 40
                          (gptel response front-sticky (gptel)) 40 87
                          (gptel response front-sticky (gptel)) 87 109
                          (gptel response front-sticky (gptel))))
                 (:type "tool_use" :id "toolu_01WHEjX2hivzCrcufv9RaCnY" :name
                        "readFile" :input (:filename "~/example.el"))])
         (:role "assistant" :content
                [(:type "tool_use" :id "01WHEjX2hivzCrcufv9RaCnY" :name
                        "readFile" :input nil :args (:filename "~/example.el"))])]
```

### DIFF
--- a/gptel-anthropic.el
+++ b/gptel-anthropic.el
@@ -117,7 +117,8 @@ information if the stream contains it.  Not my best work, I know."
            ((looking-at "message_delta")
             ;; collect stop_reason, usage_tokens and prepare tools
             (forward-line 1) (forward-char 5)
-            (when-let* ((tool-use (plist-get info :tool-use)))
+            (when-let* ((tool-use (plist-get info :tool-use))
+                        (response (gptel--json-read)))
               (let* ((data (plist-get info :data))
                      (prompts (plist-get data :messages)))
                 (plist-put ; Append a COPY of response text + tool-use to the prompts list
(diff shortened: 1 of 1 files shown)


karthink:
@tcahill Wow, good catch!  Thank you for diagnosing it.

I'll merge this soon, I want to set up a test for tool parsing with the stream parser for Anthropic first.  Do you still have access to the log for the stream that triggered this error for you?

tcahill:
Sure! Here's the log (w/ API key redacted): [gptel.log](https://github.com/user-attachments/files/19406852/gptel.log)

karthink:
Thanks.  It looks like I won't be able to write the tests for a couple more days, so I'm going ahead and merging this now.  If this breaks something else in the Anthropic stream parser (unlikely as that is), I'll bring it up here.

PR #735, opened 2025-03-23
fix(gptel): add missing newline after end_tool marker

LuciusChen:


karthink:
Thanks @LuciusChen.  Whitespace handling around tool-result blocks is more subtle than it appears because of streaming, and this fix is required but it doesn't actually cover the edge cases.  I'll add some more fixes on top of this.

PR #738, opened 2025-03-24
Add gptel-register-tool function

ultronozm:
* gptel.el (gptel-register-tool): New function to register a tool with
gptel, replacing any existing tool with the same name.

Following
https://github.com/skissue/llm-tool-collection/pull/2#issuecomment-2746561384

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1675,6 +1675,18 @@ callback as its first argument, which it runs with the result:
            nil nil #'equal)
           tool)))
 
+(defun gptel-register-tool (spec)
+  "Register a tool defined by SPEC with gptel.
+SPEC is a plist that can be passed to `gptel-make-tool'.  Add the tool
+to `gptel-tools', replacing any tool with the same name."
+  (let ((tool (apply #'gptel-make-tool spec)))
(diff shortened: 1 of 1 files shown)


karthink:
@ultronozm I'm a little confused about this.  `gptel-make-tool` already registers the tool in `gptel--known-tools`, so the only thing this function does is to add a tool to `gptel-tools`?

ultronozm:
@karthink I had thought that `gptel-make-tool` was a dumb struct
constructor that did not register the tool, but I see now that it does.

The intent of this PR was to provide a hook via which evaluating an
`llm-tool-collection-deftool` macro would update the tool in `gptel`, to
make it easier to develop and test tools.  I can't check right now
whether `gptel-make-tool` already achieves this due to the issue noted
at https://github.com/skissue/llm-tool-collection/issues/3, but if not,
then I guess `gptel-make-tool` would be the right place to address that.

Thus, I think you can safely close this PR.  Thanks!

ultronozm:
I can confirm now that there's no use for this PR, so I'm closing it.

Issue #691, opened 2025-03-11
The highlighting of the currently-focused rewritten region makes elsip code illegible


harob:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

When I use `gptel-rewrite` to modify a region of elisp code, and then place the cursor within the region, the bright purple background makes much of the code illegible -- see screenshot below. I have tried customizing `gptel-rewrite-highlight-face`, but that only changes the background when the cursor is outside of the rewritten region, not inside.

<img width="843" alt="Image" src="https://github.com/user-attachments/assets/ebe97068-cf94-45a4-a8c5-89199ee3e897" />

This issue does not happen with Python code. I'm using the `sanityinc-tomorrow-bright` color scheme from the `color-theme-sanityinc-tomorrow` package.

Thanks!

### Backend

None

### Steps to Reproduce

1. Select a region of elisp code
2. M-x `gptel-rewrite` -> press `r`
3. Move cursor inside the rewritten region

### Additional Context

Emacs version: GNU Emacs 30.1 (build 1, aarch64-apple-darwin21.6.0, NS appkit-2113.65 Version 12.7.6 (Build 21H1320)) of 2025-02-24
Mac OS: Version 14.2.1 (23C71)

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

karthink:
This is a strange one, I don't know what's causing that highlighting.  It doesn't appear to be any of the properties I'm adding to the overlay in `gptel-rewrite`.  

It should also be highlighted as a button when you move your mouse over the overlay (so you can just click), but that color is different as well.

Maybe someone will know.

pabl0:
It is the `show-paren-match` face `show-paren-mode`, which matches the beginning `(`, but of course it highlights the whole overlay.

harob:
@pabl0 you are right! If I turn off `show-paren-mode` then the problem goes away. And conversely, if I run gptel-rewrite with `show-paren-mode` on and on a block of Python code that starts with a `(` then the issue manifests.

Any idea why `show-paren-mode` is erroneously highlighting the whole overlay, and/or how to fix the bad interaction?

pabl0:
[...]

PR #744, opened 2025-03-26
gptel-rewrite: Make other modes not clobber rewrite overlay

pabl0:
When rewriting sections of Lisp or other code where the initial character is a parenthesis, it results in `show-paren-mode`highlighting the entire overlay when the point is positioned within the overlay area.  Avoid this issue by using a higher priority for `gptel-rewrite-highlight-face`.  (#691)

* gptel-rewrite.el (gptel-rewrite-highlight-face): Inherit from the `default'`face to avoid foreground being the same as `show-paren-match`, which is usually a reverse color. (gptel--rewrite-callback): Use priority higher than `show-paren-priority`.

---

I believe this fixes issue #691.

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -87,9 +87,9 @@ overlay."
 
 (defface gptel-rewrite-highlight-face
   '((((class color) (min-colors 88) (background dark))
-     :background "#041714" :extend t)
+     :background "#041714" :extend t :inherit default)
     (((class color) (min-colors 88) (background light))
-     :background "light goldenrod yellow" :extend t)
+     :background "light goldenrod yellow" :extend t :inherit default)
     (t :inherit secondary-selection))
   "Face for highlighting regions with pending rewrites."
   :group 'gptel)
@@ -434,6 +434,7 @@ INFO is the async communication channel for the rewrite request."
               (add-hook 'eldoc-documentation-functions #'gptel--rewrite-key-help nil 'local)
               ;; (overlay-put ov 'gptel-rewrite response)
               (overlay-put ov 'face 'gptel-rewrite-highlight-face)
+	      (overlay-put ov 'priority 2000)
               (overlay-put ov 'keymap gptel-rewrite-actions-map)
               (overlay-put ov 'mouse-face 'highlight)
               (overlay-put


karthink:
Thanks for the PR.  I think we need to set the rewrite overlay, but with a better solution than `(1+ show-paren-priority)`.  There are overlays created by many minor modes that can cause the same problem.

gptel-rewrite.el
```diff
@@ -87,9 +87,9 @@ overlay."
 
 (defface gptel-rewrite-highlight-face
   '((((class color) (min-colors 88) (background dark))
-     :background "#041714" :extend t)
+     :background "#041714" :extend t :inherit default)
```
karthink, line 90: Could you explain what the purpose of `:inherit default` is here?

pabl0: > Could you explain what the purpose of `:inherit default` is here?

Without it, only the background color will be overridden by the higher priority face `gptel-rewrite-highlight-face`. And since show-paren-mode usually uses reverse colors, that makes the gptel overlay text basically unreadable.

See (info "(elisp) Overlay Properties").

karthink: Ah, I didn't realize `show-paren-mode` uses `:inverse-video`.  But then there are faces that use `:inverse-video` that we do need to respect, like `highlight` (selected region face).

pabl0: Well it doesn't literally use `:inverse-video` and the standard face in `faces.el`does not change the foreground color. But the `sanityinc-tomorrow-bright` (which is a dark theme, BTW, with a slightly confusing name) theme sets background color to `#c397d8`and foreground to `#000000`, and if we only override the background color, there isn't much contrast left.

Perhaps instead of `:inherit default` we just pick some nice foreground color for both dark and light modes?

BTW for me, the gptel overlay doesn't change colors when it's inside region.

karthink: I'll merge this now in the spirit of not chasing a perfect solution.  We can revisit it if there are issues with other faces later.

karthink:
@pabl0 Your commit messages are great, so comprehensive!

PR #745, opened 2025-03-26
gptel-rewrite: Use correct form of `with-demoted-errors`

pabl0:
Avoid these warnings:
Warning: Missing format argument in 'with-demoted-errors`

This was also suggested by Stefan Monnier in the emacs-devel mailing list.


karthink:
Thanks @pabl0.  I always get the syntax of `with-demoted-errors` and `ignore-errors` confused.  I'm demonstrating once again that no change is too small to introduce bugs.

PR #746, opened 2025-03-26
Add Gemini 2.5 Pro

benthamite:
https://blog.google/technology/google-deepmind/gemini-model-thinking-updates-march-2025/

### DIFF
--- a/gptel-gemini.el
+++ b/gptel-gemini.el
@@ -440,7 +440,7 @@ Keys:
   include when using this model.
 
 Information about the Gemini models was obtained from the following
-source:
+sources:
 
 - <https://ai.google.dev/pricing>
 - <https://ai.google.dev/gemini-api/docs/models>")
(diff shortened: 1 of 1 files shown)


benthamite:
The source for the cost ($0) comes from inspecting Going to the AI Studio (https://aistudio.google.com/prompts/new_chat), going to the Models section on the right sidebar, and hovering over `Gemini 2.5 Pro Experimental 03-25`.

pabl0:
According to (info "(elisp) Documentation Tips") hyperlinks should have the format

```
URL `https://gnu.org'

```

```

to have them appear as clickable links in `describe-variable`.

benthamite:
Thanks, @pabl0. In my system, the two formats are equally “clickable” in the `help` buffer that pops up upon running `M-x describe-variable RET gptel--gemini-models`. Maybe there’s some other package or feature that makes strings that look like URLs clickable? I ask to decide whether making the change is worth it.

benthamite:
I agree with all the points you raise.

We could create a separate issue if you think it is worth discussing this further.

karthink:
`gemini-2.5-pro-exp-03-25` is already in the models list, it was added in #742.  Could you check if the metadata for that entry needs to be edited instead?  I can sill merge this PR for the URLs.

PR #754, opened 2025-03-29
UTF-8 encoding issue in *ChatGPT* buffer

kayomarz:
Bullet points (•) from GPT responses get incorrectly printed as `â\200¢` in the `*ChatGPT*` buffer.

The problem does not seem to happen if the process coding system is explicitly set to `utf-8-unix` when reading output from subprocesses.

Platform: Linux (Debian 12.5)

(I have been using gptel since around two weeks and it is great! Thank you karthink.)

Below is a screenshot of the incorrect printing of bullets:

encoding-issue

Below is a screenshot after this code commit.

encoding-issue-fix

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -135,11 +135,10 @@ the response is inserted into the current buffer after point."
          (stream (plist-get info :stream))
          (process (apply #'start-process "gptel-curl"
                          (gptel--temp-buffer " *gptel-curl*") "curl" args)))
+    ;; Don't try to convert cr-lf to cr on Windows so that curl's "header size
+    ;; in bytes" stays correct. Explicitly set utf-8 for non-win systems too,
+    ;; for cases when buffer coding system is not set to utf-8.
     (set-process-coding-system process 'utf-8-unix 'utf-8-unix)
-    (when (memq system-type '(windows-nt ms-dos))
-      ;; Don't try to convert cr-lf to cr on Windows so that curl's "header size
-      ;; in bytes" stays correct
-      (set-process-coding-system process 'utf-8-unix 'utf-8-unix))
     (when (eq gptel-log-level 'debug)
       (gptel--log (mapconcat #'shell-quote-argument (cons "curl" args) " \\\n")
                   "request Curl command" 'no-json))
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -135,6 +135,7 @@ the response is inserted into the current buffer after point."
          (stream (plist-get info :stream))
          (process (apply #'start-process "gptel-curl"
                          (gptel--temp-buffer " *gptel-curl*") "curl" args)))
+    (set-process-coding-system process 'utf-8-unix 'utf-8-unix)
     (when (memq system-type '(windows-nt ms-dos))
       ;; Don't try to convert cr-lf to cr on Windows so that curl's "header size
       ;; in bytes" stays correct
--- a/gptel.el
+++ b/gptel.el
@@ -2763,6 +2763,8 @@ the response is inserted into the current buffer after point."
                            (if (functionp backend-url)
                                (funcall backend-url) backend-url))
                          (lambda (_)
+                           (set-buffer-multibyte t)
+                           (set-buffer-file-coding-system 'utf-8-unix)
                            (pcase-let ((`(,response ,http-status ,http-msg ,error)
                                         (gptel--url-parse-response backend info))
                                        (buf (current-buffer)))


gptel-curl.el
```diff
@@ -135,6 +135,7 @@ the response is inserted into the current buffer after point."
          (stream (plist-get info :stream))
          (process (apply #'start-process "gptel-curl"
                          (gptel--temp-buffer " *gptel-curl*") "curl" args)))
+    (set-process-coding-system process 'utf-8-unix 'utf-8-unix)
```
karthink, line 138: See the check for `windows-nt` immediately below this line.  That can be removed and we can unconditionally set the process coding system.

karthink:
Thanks for the PR @kayomarz.

We need to make the same change in `gptel--url-get-response`, when not using Curl.  Before that, can you check if the same problem occurs when you don't use Curl? `(setq gptel-use-curl nil)`.

karthink:
Also, I'm curious about what the coding system is set to by default in your Emacs, if not `utf-8-unix`.

kayomarz:
My emacs init does not set any coding system. I might add the following to my emacs init:

(set-default-coding-systems 'utf-8)
(setq locale-coding-system 'utf-8)
(set-selection-coding-system 'utf-8)

kayomarz:
btw on your system, what does `(describe-current-coding-system)` give you?

gptel-curl.el
kayomarz: Done - Coding system is set [unconditionally](https://github.com/karthink/gptel/pull/754/files?diff=unified&w=0#diff-3b65c1cc47ce5206ac9e29d41bac39285eb1fe9dc192603cb6120c752cb6b57b).

karthink:
> btw on your system, what does `(describe-current-coding-system)` give you?

<details>
<summary>M-x describe-current-coding-system</summary>

```
Coding system for saving this buffer:
  U -- utf-8-unix (alias: mule-utf-8-unix cp65001-unix)

Default coding system (for new files):
  U -- utf-8-unix (alias: mule-utf-8-unix cp65001-unix)

Coding system for keyboard input:
  U -- utf-8-unix (alias: mule-utf-8-unix cp65001-unix)

Coding system for terminal output:
  U -- utf-8-unix (alias: mule-utf-8-unix cp65001-unix)

Coding system for inter-client cut and paste:
  nil
Defaults for subprocess I/O:
  decoding: U -- utf-8-unix (alias: mule-utf-8-unix cp65001-unix)

  encoding: U -- utf-8-unix (alias: mule-utf-8-unix cp65001-unix)


Priority order for recognizing coding systems when reading files:
  1. utf-8 (alias: mule-utf-8 cp65001)
  2. iso-2022-7bit 
  3. iso-latin-1 (alias: iso-8859-1 latin-1)
  4. iso-2022-7bit-lock (alias: iso-2022-int-1)
  5. iso-2022-8bit-ss2 
  6. emacs-mule 
  7. raw-text 
  8. iso-2022-jp (alias: junet)
  9. in-is13194-devanagari (alias: devanagari)
  10. chinese-iso-8bit (alias: cn-gb-2312 euc-china euc-cn cn-gb gb2312)
  11. utf-8-auto 
  12. utf-8-with-signature 
  13. utf-16 
  14. utf-16be-with-signature (alias: utf-16-be)
  15. utf-16le-with-signature (alias: utf-16-le)
  16. utf-16be 
  17. utf-16le 
  18. japanese-shift-jis (alias: shift_jis sjis)
  19. chinese-big5 (alias: big5 cn-big5 cp950)
  20. undecided 

  Other coding systems cannot be distinguished automatically
  from these, and therefore cannot be recognized automatically
  with the present coding system priorities.

Particular coding systems specified for certain file names:

  OPERATION	TARGET PATTERN		CODING SYSTEM(s)
  ---------	--------------		----------------
  File I/O      "\\.\\(arc\\|zip\\|lzh\\|lha\\|zoo\\|[jew]ar\\|xpi\\|rar\\|7z\\|squashfs\\|ARC\\|ZIP\\|LZH\\|LHA\\|ZOO\\|[JEW]AR\\|XPI\\|RAR\\|7Z\\|SQUASHFS\\)\\'"
                                        no-conversion-multibyte
                "\\.\\(exe\\|EXE\\)\\'" no-conversion
                "\\.\\(sx[dmicw]\\|odt\\|tar\\|t[bg]z\\)\\'"
                                        no-conversion
                "\\.\\(gz\\|Z\\|bz\\|bz2\\|xz\\|gpg\\)\\'"
                                        no-conversion
                "\\.\\(jpe?g\\|png\\|gif\\|tiff?\\|p[bpgn]m\\)\\'"
                                        no-conversion
                "\\.pdf\\'"             no-conversion
                "/#[^/]+#\\'"           utf-8-emacs-unix
                "\\.tzst\\'"            (no-conversion . no-conversion)
                "\\.zst\\'"             (no-conversion . no-conversion)
                "\\.dz\\'"              (no-conversion . no-conversion)
                "\\.txz\\'"             (no-conversion . no-conversion)
                "\\.xz\\'"              (no-conversion . no-conversion)
                "\\.lzma\\'"            (no-conversion . no-conversion)
                "\\.lz\\'"              (no-conversion . no-conversion)
                "\\.g?z\\'"             (no-conversion . no-conversion)
                "\\.\\(?:tgz\\|svgz\\|sifz\\)\\'"
                                        (no-conversion . no-conversion)
                "\\.tbz2?\\'"           (no-conversion . no-conversion)
                "\\.bz2\\'"             (no-conversion . no-conversion)
                "\\.Z\\'"               (no-conversion . no-conversion)
                "\\.elc\\'"             utf-8-emacs
                "\\.el\\'"              prefer-utf-8
                "\\.utf\\(-8\\)?\\'"    utf-8
                "\\.xml\\'"             xml-find-file-coding-system
                "\\(\\`\\|/\\)loaddefs.el\\'"
                                        (raw-text . raw-text-unix)
                "\\.tar\\'"             (no-conversion . no-conversion)
                "\\.po[tx]?\\'\\|\\.po\\."
                                        po-find-file-coding-system
                "\\.\\(tex\\|ltx\\|dtx\\|drv\\)\\'"
                                        latexenc-find-file-coding-system
                ""                      (undecided)
  Process I/O	nothing specified
  Network I/O	nothing specified
```
</details>

We also need to set the process coding system to `utf-8-unix` in `gptel--url-get-response`.

kayomarz:
> We also need to set the process coding system to `utf-8-unix` in `gptel--url-get-response`.

I gave it an attempt but it wasn't as straightforward.  Will try further.

kayomarz:
> We also need to set the process coding system to `utf-8-unix` in `gptel--url-get-response`.

[Done](https://github.com/karthink/gptel/pull/754/commits/e787b313741703f53442ae7916576be5255543dd).

btw for this case, we did not set the **process** coding system as we did in the case of curl as discussed above.

Instead we used  `(set-buffer-file-coding-system ...)` after setting the buffer to multibyte.

karthink:
> btw for this case, we did not set the **process** coding system as we did in the case of curl as discussed above.
>
> Instead we used  `(set-buffer-file-coding-system ...)` after setting the buffer to multibyte.

Yes, this is the way to change url-retrieve settings, everything is done via buffer-local variables.

I think you can also do

(set-process-coding-system (get-buffer-process (current-buffer)) ...)

kayomarz:
btw, on my system `iso-latin-1` was the default for buffer-file and process coding systems. Just to experiment, I made `utf-8` the default coding system using `(set-default-coding-systems 'utf-8)`. Now gptel **using curl**  worked fine without using any of the code fixes mentioned above in this PR.

Does this mean that your init has something to make utf-8 the default? Or might this be related to my emacs build that `iso-latin-1` is my default?  Or is it more likely related to my system locale?

karthink:
> Does this mean that your init has something to make utf-8 the default? Or might this be related to my emacs build that `iso-latin-1` is my default?  Or is it more likely related to my system locale?

It's probably the system locale.  I checked, I don't set the coding system to utf-8 explicitly in my config.

kayomarz:
> I think you can also do (set-process-coding-system (get-buffer-process (current-buffer)) ...)

I tried this but it didn't work.  Assuming `set-process-coding-system` would need to be done before some **process** passes the response to emacs, I tried setting it immediately after calling `url-retrieve`, not waiting for the callback. But something goes wrong and no content appers in the gptel `*ChatGPT*` buffer.

Seemingly related, I noticed that `url-retrieve` does not allow the process coding system to a be multibyte buffer such as `utf-8` and in which case this error occurs: `This function can be called only in unibyte buffers` where "This function" refers to `zlib-decompress-region` which gets called by `url-http-generic-filter`.

gptel.el
```diff
@@ -2763,6 +2763,8 @@ the response is inserted into the current buffer after point."
                            (if (functionp backend-url)
                                (funcall backend-url) backend-url))
                          (lambda (_)
+                           (set-buffer-multibyte t)
```
karthink, line 2766: To be safe, I think this needs to be set as soon as the process is started, and not in the callback.  The buffer returned by `url-retrieve` is stored in `proc-buf` in the let-binding, so you can instead run
```elisp
(with-current-buffer proc-buf
  (set-buffer-multibyte t)
  (set-buffer-file-coding-system 'utf-8-unix))
```
at the end of the function, above the line that says `TODO: Add transformer here`.

kayomarz: ok, let me give this a try... I think I came across the situation where `url-retrive` does not allow the process buffer to be a multibyte buffer.

karthink:
> I think I came across the situation where `url-retrive` does not allow the process buffer to be a multibyte buffer.

Cool, if that happens we can live with setting the coding-system in the callback.  But it would be safer to do it right away.

gptel.el
kayomarz: Tried it... this error appears in the message buffer and gptel remains in the "Waiting..." state.

    error in process filter: This function can be called only in unibyte buffers


I wonder if this is a limitation with function [url-retrieve](https://github.com/karthink/gptel/blob/975c3e64eb834b939e0d61dfc39fed8395afcc45/gptel.el#L2747). I do not know much but it looks like the process buffer cannot be made multibyte when using `url-retrieve`.

The below code demonstrates the error.

    (with-current-buffer (url-retrieve "https://www.wikipedia.org/"
                                              (lambda (_)
                                                (switch-to-buffer (current-buffer))))
             (set-buffer-multibyte t) )

[url-retrieve-internal](https://github.com/emacs-mirror/emacs/blob/6a299b3caceb2c73b932ba73849738faa8c5d975/lisp/url/url.el#L169) selects the loader [url-http](https://github.com/emacs-mirror/emacs/blob/6a299b3caceb2c73b932ba73849738faa8c5d975/lisp/url/url.el#L206) which sets filter function [url-http-generic-filter](https://github.com/emacs-mirror/emacs/blob/6a299b3caceb2c73b932ba73849738faa8c5d975/lisp/url/url-http.el#L1389) which eventually calls [zlib-decompress-region](https://github.com/emacs-mirror/emacs/blob/6a299b3caceb2c73b932ba73849738faa8c5d975/src/decompress.c#L247) which has a condition to decline multibyte buffers, printing the error: `This function can be called only in unibyte buffers`

Stack trace:

>   zlib-decompress-region(871 22994 t)
>   url-handle-content-transfer-encoding()
>   url-http-parse-headers()
>   url-http-content-length-after-change-function(19018 23017 3999)
>   url-http-generic-filter(#<process www.wikipedia.org> "...")

karthink: I wasn't aware of this, thanks for digging through the call stack to figure out the problem.

Let's leave it as it is then, with the buffer coding system set in the callback. :+1:

karthink:
Thank you for the PR and for the careful testing, @kayomarz!

Issue #716, opened 2025-03-18
test submodule marks repo as dirty in straight.el


martenlienen:
Since the tests were moved into a submodule, [straight.el](https://github.com/radian-software/straight.el) thinks that the local gptel repo is dirty and asks me interactively what to do when I `straight-pull-package` or `straight-pull-all` to update. Is there a specific reason why the tests are in separate repository?

karthink:
> Since the tests were moved into a submodule, [straight.el](https://github.com/radian-software/straight.el) thinks that the local gptel repo is dirty and asks me interactively what to do when I straight-pull-package or straight-pull-all to update.

Hmm, I guess it doesn't update the submodule record automatically.  You could try asking at the Straight repo if there's a setting for this.  I don't use Straight.

> Is there a specific reason why the tests are in separate repository?

Yes, there are now 47 test data files (org, markdown and lisp data) used for testing, and there will eventually be 200+.  I don't want all users using Elpaca/Straight/package-vc to be downloading so many unnecessary files.

gregoryg:
Thank you for this simple solution, @martenlienen !  This has been a thorn in my side for several months.

I believe this problem affects everyone using straight.el.  The Doom Emacs version of this fix (from #758)  is already in the documentation.  

This recipe should replace what's currently in the README for straight.el users!

karthink:
@gregoryg or @martenlienen , feel free to submit a PR to the readme.  Patch also works.

gregoryg:
#1046

PR #758, opened 2025-04-01
README: Install without submodule in Doom

real-or-random:
This avoids the `test` submodule in Doom Emacs, which uses Straight, see #716.


karthink:
@real-or-random Thank you!

Issue #502, opened 2024-12-10
What token to use for Github 


kidmose:
**Briefly describe what you are trying to do**

Set up gptel to use Github Copilot

1. Installed `gptel` with `package-install`
2. Created token from https://github.com/settings/tokens (Tried both "classic" and "fine-grained") 
3. Set up gptel in emacs, `init.el`:

        (use-package gptel
          :ensure t
          :custom
          (gptel-model 'gpt-4o)
          (gptel-log-level 'debug)
          (gptel-backend
           (gptel-make-openai "Github Models"
                              :host "models.inference.ai.azure.com"
                              :endpoint "/chat/completions"
                              :stream t
                              :key "<REDACTED>"
                              :models '(gpt-4o))))

4. Started gptel buffer: `M-x gptel`
5. Selected default (I.e. "Github Models" from above)
6. Entered prompt and press `C-c RET`

Expected: Some response from the LLM, entered into the buffer.

Received:
- Nothing in gptel buffer (`*Github Models*`)
- Error in `*gptel-log*` (full content below): 401 Your account type is not currently supported

Is there anyone able to provide some guidance on what I am missing here? (Apparently some permission/role from my GH org. but which?)
Would be greatly appreciated :)

**Additional context**
Emacs version: GNU Emacs 28.2
Operating system: Debian on 12 on WSL

** `gptel.log` **

    {
      "gptel": "request headers",
      "timestamp": "2024-12-10 15:56:10"
    }
    {
      "Content-Type": "application/json",
      "Authorization": "Bearer <REDACTED>"
    }
    {
      "gptel": "request body",
      "timestamp": "2024-12-10 15:56:10"
    }
    {
      "model": "gpt-4o",
      "messages": [
        {
          "role": "system",
          "content": "You are a large language model living in Emacs and a helpful assistant. Respond concisely."
        },
        {
          "role": "user",
          "content": "hi"
        }
      ],
      "stream": true,
      "temperature": 1.0
    }
    {"gptel": "request Curl command", "timestamp": "2024-12-10 15:56:10"}
    curl \
    --disable \
    --location \
    --silent \
    --compressed \
    -XPOST \
    -y300 \
    -Y1 \
    -D- \
    -w\(<REDACTED>\ .\ \%\{size_header\}\) \
[...]

PR #670, opened 2025-02-28
AWS Bedrock Support

felipeochoa:
Closes #379 

Use it with this:

```el
(gptel-make-bedrock "bedrock" :region "us-east-1")
```

I've manually tested with streaming on and off. Tool use, media, and context all work

PR #767, opened 2025-04-06
gptel-openai-extras: Add support for Github Copilot Chat

kiennq:
Solve the issue mentioned in #502 
This adds support for the more premium Github copilot chat models in https://github.com/copilot.

Note: this implementation is based on https://github.com/chep/copilot-chat.el


karthink:
@kiennq thanks for the PR!

This looks good to me, with one issue that I need to resolve:  The model metadata are stored as symbol properties, so a model (== symbol) like `claude-3.7-sonnet` can only have one set of metadata.  If someone defines both an anthropic and github-copilot backend, the model metadata for `claude-3.7-sonnet` will be set to whichever definition ran second.

This is not a deal-breaker because the model capabilities will be the same and the rest of the metadata only shows up as completion annotations, but I still need to fix it.

kiennq:
Yeah, since the model's name is used in the message sending to server as well, I have to fix it like that. It would be better to have other optional field (`id` maybe?) so that once it's set, it's used as the model's name in the request. Then we can twitch it slightly to have different model names for different providers.

gptel.el
```diff
@@ -912,6 +912,60 @@ Later plists in the sequence take precedence over earlier ones."
...
+    (dotimes (_ length)
+      (setq hex (nconc hex (list (aref hex-chars (random 16))))))
+    (apply #'string hex)))
+
+(cl-defun gptel--url-retrieve (url &key method data headers)
+  "Retrieve URL synchronously with METHOD, DATA and HEADERS."
+  (let ((url-request-method (if (eq method'post) "POST" "GET"))
+        (url-request-data (encode-coding-string (gptel--json-encode data) 'utf-8))
+        (url-mime-accept-string "application/json")
+        (url-request-extra-headers
```
karthink, line 939: Since you are setting copilot-chat-specific headers here, this is not a general-purpose function and can be moved to `gptel-gh.el`.

gptel.el
```diff
@@ -912,6 +912,60 @@ Later plists in the sequence take precedence over earlier ones."
...
+           ("editor-version" . "Neovim/0.10.0")
+           ,@headers)))
+    (with-current-buffer (url-retrieve-synchronously url)
+      (goto-char url-http-end-of-headers)
+      (gptel--json-read))))
+
+(defun gptel--restore (file)
+  "Restore saved object from FILE."
+  (when (file-exists-p file)
+    (car (read-from-string
```
karthink, line 952: I suggest switching to the simpler `(read (current-buffer))` instead of converting the buffer to a string and then running `read-from-string`.

If this is not desirable for some reason, please let me know.

gptel.el
```diff
@@ -912,6 +912,60 @@ Later plists in the sequence take precedence over earlier ones."
...
+          (let ((coding-system-for-read 'utf-8-auto-dos))
```
karthink, line 953: I'm confused about why the coding system for read is `utf-8-auto-dos` while the coding system for write below in `gptel--save` is `utf-8-unix`.

Since the token is alphanumeric, the difference is going to be a an extra `^M` char at the end, which probably won't matter.  But is there any reason these two are different?

karthink (changes requested):
1. I think this code is doing enough differently that it can be moved into a separate file, perhaps `gptel-gh.el`, or `gptel-github.el`?  (I would avoid `gptel-copilot.el` since that makes it seem like gptel provides completion-at-point like copilot does, which it doesn't.)
2. The additions to `gptel.el` are specific to copilot, so please move them to `gptel-gh.el` as well.
3. Ensure all functions and variables in this file use the `gptel--gh-` or `gptel--github-` prefix.  (Except for the defcustoms and `gptel-make-gh` which will use single dashes)
4. Please skim through the changes in PR #670, in case there's something that can be shared.

gptel.el
kiennq: The reason of using `utf-8-auto-dos` for read is so that the `^M` character will be ignored. So for example, the save file is opened by some editor in Windows that automatically changing the EOL to CRLF, we can still properly read it.
Whereas on saving, we should save as `utf-8-unix` style so that no `^M` characters got in

gptel.el
kiennq: Oh, I forgot those extra headers, will move those to another file while keeping the general-purpose function here

gptel.el
karthink: Please add a short comment with this explanation (1-line) where you set `utf-8-auto-dos`, otherwise I'll be confused again in the future.

kiennq: Done

karthink:
@kiennq Thanks for making the adjustments.  Please see point 2 above, after which I can merge this.

> 2. The additions to gptel.el are specific to copilot, so please move them to gptel-gh.el as well.

More specifically, I think `gptel--url-retrieve` can stay in `gptel.el`, as it has other planned uses.  Please move the rest of the changes from `gptel.el` into `gptel-gh.el`.

gptel-gh.el
```diff
@@ -0,0 +1,211 @@
...
+  (pcase-let (((map :access_token :expires_at)
+               (gptel--gh-token gptel-backend)))
+    (when (or (null access_token)
+              (and expires_at
+                   (> (round (float-time (current-time)))
+                      expires_at)))
+      (gptel--gh-renew-token))))
+
+;;;###autoload
+(cl-defun gptel-make-gh-copilot
```
karthink, line 177: Can we call this just `gptel-make-gh`?  I don't feel particularly strongly about this, but I'm trying to avoid the name `copilot` since that immediately gives people the idea of tab-completion.

kiennq: I think that's a misconception that should be corrected. You can see that the website https://github.com/copilot is `copilot` but it's not tab-completion.
And there's even a M365 copilot search (apps and websites) that has nothing to do with tab-completion at all, all under the same Copilot brand.
I don't think the gptel is providing code completion at the current capacity, but at that time we should name it `gptel-completion` instead of using copilot. As it will eventually causes confusion when the user is using more and more product from Copilot branding.

kiennq: > By whom? This naming confusion is Microsoft's fault, they don't seem to be clear on what copilot is either. 

Yeah, by Microsoft, or Github as the naming is involving from a single function to a brand. There's even Microsoft Copilot Studio on Azure now and they're doing quite a different thing.

> Maybe we can add a line to the README where the instructions for Github Copilot are clarifying that this is the chat component only.

I think I mentioned it as Github CopilotChat in the README already, is that also okay?

> call it the more explicit `gptel-make-github-copilot`

I concise `github` to `gh` as their CLI tool so can we keep the `gh` shortcut name?

karthink: > I think I mentioned it as Github CopilotChat in the README already, is that also okay?

Yeah, I think it's fine as is.

> I concise github to gh as their CLI tool so can we keep the gh shortcut name?

Ah, I didn't see the connection with the gh CLI tool name until now.  That makes sense.

No actions needed here. :+1:

gptel-gh.el
```diff
@@ -0,0 +1,211 @@
...
+     :output-cost 15
+     :cutoff-date "2025-02")
+    (gemini-2.0-flash-001
+     :description "Next gen, high speed, multimodal for a diverse variety of tasks"
+     :capabilities (tool-use json)
+     :context-window 1000
+     :input-cost 0.10
+     :output-cost 0.40
+     :cutoff-date "2024-08")
+    ))
```
karthink, line 87: Move dangling parens to previous line.

gptel-gh.el
```diff
@@ -0,0 +1,211 @@
...
+  :type 'string
+  :group 'gptel)
+
+(defcustom gptel-gh-token-file (expand-file-name ".cache/copilot-chat/token"
+                                                 user-emacs-directory)
+  "The file where the GitHub token is cached."
+  :type 'string
+  :group 'gptel)
+
+(defvar gptel--gh-auth-common-headers
```
karthink, line 106: Change to `defconst`

gptel.el
```diff
@@ -912,6 +912,58 @@ Later plists in the sequence take precedence over earlier ones."
         (setq rtn (plist-put rtn p v))))
     rtn))
 
+;; https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)
+(defun gptel--uuid ()
```
karthink, line 916: Move to `gptel-gh.el` with appropriate prefix.

gptel.el
```diff
@@ -912,6 +912,58 @@ Later plists in the sequence take precedence over earlier ones."
...
+  "Generate a UUID v4-1."
+  (format "%04x%04x-%04x-4%03x-8%03x-%04x%04x%04x"
+          (random #x10000) (random #x10000)
+          (random #x10000)
+          (random #x1000)
+          (random #x1000)
+          (random #x10000) (random #x10000) (random #x10000)))
+
+(defun gptel--machine-id ()
```
karthink, line 925: Move to `gptel-gh.el` with appropriate prefix.

gptel.el
```diff
@@ -912,6 +912,58 @@ Later plists in the sequence take precedence over earlier ones."
...
+        (url-request-data (encode-coding-string (gptel--json-encode data) 'utf-8))
+        (url-mime-accept-string "application/json")
+        (url-request-extra-headers
+         `(("content-type" . "application/json")
+           ,@headers)))
+    (with-current-buffer (url-retrieve-synchronously url 'silent)
+      (goto-char url-http-end-of-headers)
+      (gptel--json-read))))
+
+(defun gptel--restore (file)
```
karthink, line 946: Move to `gptel-gh.el` with appropriate prefix.

gptel.el
```diff
@@ -912,6 +912,58 @@ Later plists in the sequence take precedence over earlier ones."
...
+    ;; We set the coding system to `utf-8-auto-dos' when reading so that
+    ;; files with CR EOL can still be read properly
+    (let ((coding-system-for-read 'utf-8-auto-dos))
+      (with-temp-buffer
+        (set-buffer-multibyte nil)
+        (insert-file-contents-literally file)
+        (goto-char (point-min))
+        (read (current-buffer))))))
+
+(defun gptel--save (file obj)
```
karthink, line 958: Move to `gptel-gh.el` with appropriate prefix.

gptel.el
```diff
@@ -912,6 +912,58 @@ Later plists in the sequence take precedence over earlier ones."
...
+  (let ((hex-chars "0123456789abcdef")
+        (length 65)
+        hex)
+    (dotimes (_ length)
+      (setq hex (nconc hex (list (aref hex-chars (random 16))))))
+    (apply #'string hex)))
+
+(cl-defun gptel--url-retrieve (url &key method data headers)
+  "Retrieve URL synchronously with METHOD, DATA and HEADERS."
+  (let ((url-request-method (if (eq method'post) "POST" "GET"))
```
karthink, line 936: Add `(declare (indent 1))`

gptel-gh.el
```diff
@@ -0,0 +1,211 @@
...
+     :method 'get
+     :headers `(("authorization"
+                 . ,(format "token %s" (gptel--gh-github-token gptel-backend)))
+                ,@gptel--gh-auth-common-headers))
+    (gptel--save gptel-gh-token-file)
+    (setf (gptel--gh-token gptel-backend))))
+
+(defun gptel--gh-auth ()
+  "Authenticate with GitHub Copilot API.
+We first need github authorization (github token).
```
karthink, line 155: Add a blank line above line 155.

gptel-gh.el
```diff
@@ -0,0 +1,211 @@
...
+  :group 'gptel)
+
+(defcustom gptel-gh-token-file (expand-file-name ".cache/copilot-chat/token"
+                                                 user-emacs-directory)
+  "The file where the GitHub token is cached."
+  :type 'string
+  :group 'gptel)
+
+(defvar gptel--gh-auth-common-headers
+  `(("editor-plugin-version" . "CopilotChat.nvim/2.0.0")
```
karthink, line 107: I'm assuming it's required to use nvim/neovim in these headers?

gptel-gh.el
```diff
@@ -0,0 +1,211 @@
...
+    ))
+
+(cl-defstruct (gptel--gh (:include gptel-openai)
+                         (:copier nil)
+                         (:constructor gptel--make-gh))
+  token github-token sessionid machineid)
+
+(defcustom gptel-gh-github-token-file (expand-file-name ".cache/copilot-chat/github-token"
+                                                        user-emacs-directory)
+  "The file where to find GitHub token."
```
karthink, line 96: Change description to "File where the GitHub token is stored."

gptel-gh.el
```diff
@@ -0,0 +1,252 @@
...
+          (header (lambda ()
+                    (gptel--gh-auth)
+                    `(("openai-intent" . "conversation-panel")
+                      ("authorization" . ,(concat "Bearer " (plist-get (gptel--gh-token gptel-backend) :token)))
+                      ("x-request-id" . ,(gptel--gh-uuid))
+                      ("vscode-sessionid" . ,(or (gptel--gh-sessionid gptel-backend) ""))
+                      ("vscode-machineid" . ,(or (gptel--gh-machineid gptel-backend) ""))
+                      ("copilot-vision-request" . "true")
+                      ("copilot-integration-id" . "vscode-chat"))))
+          (host "api.githubcopilot.com")
```
eotsn, line 229: Based on [this discussion](https://github.com/BerriAI/litellm/pull/9079#discussion_r2022334840) it seems like this should be configurable to support business users? Though I haven't encountered any issues using [copilot-chat.el](https://github.com/chep/copilot-chat.el) with a business subscription, and it uses the same domain name as here.

kiennq: Same, I have github enterprise and it works just fine. I see in the token it has the `endpoints` which point to `api.enterprise.githubcopilot.com` for me, but in the `token` field, it includes that sku as well. So my guess is that the `api.githubcopilot.com` is smart and can dispatch it to appropriated service based on the token as needed.

Anyway, this is configurable and a parameter of `gptel-make-gh-copilot` already so the user can change as they want.

eotsn: Yeah, switching between the domains doesn't appear to make any difference, so that seems to be the case.

This is going to be a very nice addition to gptel, thank you for your work!

karthink:
@kiennq when trying to test this out before merging, I get the following error:

```

**Comment by karthink** [comment]
Ah, the following code:

```elisp
(gptel--url-retrieve
     "https://api.github.com/copilot_internal/v2/token"
     :method 'get
     :headers `(("authorization"
                 . ,(format "token %s" (gptel--gh-github-token gptel-backend)))
                ,@gptel--gh-auth-common-headers))
```
returns this structure:
```elisp
(:message "Not Found" :documentation_url "https://docs.github.com/rest" :status "404")
```
which is registered as the session token.  The error handling needs to be fixed.

kiennq:
> Ah, the following code:
> 
> ```emacs-lisp
> (gptel--url-retrieve
>      "https://api.github.com/copilot_internal/v2/token"
>      :method 'get
>      :headers `(("authorization"
>                  . ,(format "token %s" (gptel--gh-github-token gptel-backend)))
>                 ,@gptel--gh-auth-common-headers))
> ```
> 
> returns this structure:
> 
> ```emacs-lisp
> (:message "Not Found" :documentation_url "https://docs.github.com/rest" :status "404")
> ```
> 
> which is registered as the session token, but is actually an error. The error handling needs to be fixed.

That means you got an invalid github token or auth service in Github got timeout or something unexpected happens.
Even if that's registered as the session token, on the next query, __it will check if the actual token is in there and try to get a new token again__. It can keep failing if the github token in the `gptel-gh-github-token-file` is invalid. But in that case we can't even detect that, we can just follow the normal procedure in `gptel`, delete the file and recreate the client.

Btw, how did you get to that state? Did you manually put the token into the `gptel-gh-github-token-file` or something?

karthink:
> It can keep failing if the github token in the gptel-gh-github-token-file is invalid.

`gptel-gh-github-token-file` is written correctly, it contains a token of the form `ghu_...`.  The call that fails is the one that writes `gptel-gh-token-file`, in `gptel--gh-renew-token`.

> But in that case we can't even detect that, we can just follow the normal procedure in gptel, delete the file and recreate the client.

It's not clear how gptel can identify this error and recreate the token file.

> Btw, how did you get to that state? Did you manually put the token into the gptel-gh-github-token-file or something?

No, it consistently and repeatedly fails as described, and writes the wrong message to `gptel-gh-token-file`.

kiennq:
> It's not clear how gptel can identify this error and recreate the token file. One way might be to check if `(plist-get (gptel--gh-token gptel-backend) :token)` exists.

It did that already at L170.

> `gptel-gh-github-token-file` is written correctly, it contains a token of the form `ghu_...`. The call that fails is the one that writes `gptel-gh-token-file`, in `gptel--gh-renew-token`.

How did you get the token in the `gptel-gh-github-token-file`?
That's not a normal github token, it's a token that's acquired as part of a special `client-id` that is registered with Github and can request a login token to authenticate.
Can you try to delete the `gptel-gh-github-token-file`, set the backend to `(setq gptel-backend (gptel-make-gh-copilot "Copilot"))`  and try again?

karthink:
> > It's not clear how gptel can identify this error and recreate the token file. One way might be to check if `(plist-get (gptel--gh-token gptel-backend) :token)` exists.
> 
> It did that already at L170.

At L170 you're checking `gptel--gh-github-token`.  I'm referring to `gptel--gh-token`.

> > `gptel-gh-github-token-file` is written correctly, it contains a token of the form `ghu_...`. The call that fails is the one that writes `gptel-gh-token-file`, in `gptel--gh-renew-token`.
> 
> How did you get the token in the `gptel-gh-github-token-file`?

Through the browser auth process.

> That's not a normal github token, it's a token that's acquired as part of a special `client-id` that is registered with Github and can request a login token to authenticate. Can you try to delete the `gptel-gh-github-token-file`, set the backend to `(setq gptel-backend (gptel-make-gh-copilot "Copilot"))` and try again?

Yup, I did this a few times when testing.  `gptel-gh-github-token-file` is always written correctly, with contents of the form `"ghu_..."`.  The problem is that `(gptel--gh-renew-token)` returns
```elisp
(:message "Not Found" :documentation_url "https://docs.github.com/rest" :status "404")`
```

It's possible that my Github account does not have access to Copilot -- that's not something we can fix in gptel.  But the above return value of `gptel--gh-renew-token` should be treated as an error and gptel should message the user explaining what went wrong.

To test if there was an error you could check to see if the return value contains the `:token` key, or perhaps there is a better way.

> That's not a normal github token, it's a token that's acquired as part of a special `client-id` that is registered with Github and can request a login token to authenticate. Can you try to delete the `gptel-gh-github-token-file`, set the backend to `(setq gptel-backend (gptel-make-gh-copilot "Copilot"))` and try again?

Yup, I did this a few times when testing.  `gptel-gh-github-token-file` is always written correctly, with contents of the form `"ghu_..."`.  The problem is that `(gptel--gh-renew-token)` returns
```elisp
(:message "Not Found" :documentation_url "https://docs.github.com/rest" :status "404")`
```

It's possible that my Github account does not have access to Copilot -- that's not something we can fix in gptel.  But the above return value of `gptel--gh-renew-token` should be treated as an error and gptel should message the user explaining what went wrong.

To test if there was an error you could check to see if the return value contains the `:token` key, or perhaps there is a better way.

kiennq:
> At L170 you're checking `gptel--gh-github-token`. I'm referring to `gptel--gh-token`.

Sorry I pointed to a wrong line at the previous change. It should be L210, see `gptel--gh-auth`.

I guess I can double check if the token exists. But the current returned error from gptel, `bad request: Authorization header is badly formatted` is also pointing to the same thing already.
Do you think an explicit guess (`You might not have access to Github Copilot Chat`) is better?

karthink:
> > At L170 you're checking `gptel--gh-github-token`. I'm referring to
> > `gptel--gh-token`.
> 
> Sorry I pointed to a wrong line at the previous change. It should be L210, see
> `gptel--gh-auth`.

I don't follow.  L210 checks to see if `(gptel--gh-token gptel-backend)` contains an `:access_token` key.  If it doesn't, it runs `(gptel--gh-renew-token)` which fetches this again and stores it as the token:
``` emacs-lisp
(:message "Not Found" :documentation_url "https://docs.github.com/rest" :status "404")
```
So there is no change to the existing value of `(gptel--gh-token gptel-backend)`, and no error detection happening here.

> I guess I can double check if the token exists. But the current returned error
> from gptel, `bad request: Authorization header is badly formatted` is also
> pointing to the same thing already.  Do you think an explicit guess (`You might
> not have access to Github Copilot Chat`) is better?

`bad request: Authorization header is badly formatted` is a mysterious message -- the error is occurring during the authentication phase, before any chat completion request is sent, and it took some effort even for me to figure out what the problem was.  So I think if `(gptel--gh-renew-token)` doesn't receive an actual token, it should report this error via `gptel--update-status` (when in `gptel-mode`) and clear the value of `(gptel--gh-token gptel-backend)` instead of leaving it set to
``` emacs-lisp
(:message "Not Found" :documentation_url "https://docs.github.com/rest" :status "404")
```

kiennq:
I've updated to explicitly quit with `user-error` when the auth can't be done. I guess it's better than the update status since, well, it quits the whole stack and show up everywhere instead of just `gptel-mode` buffer

kiennq:
Done

karthink:
@kiennq Thank you for the PR!

I can't test it because of the above error, so we'll have to wait and see if it's working as intended for most users.

Issue #769, opened 2025-04-07
Insert space/newline after tool call when gptel-include-tool-results is nil


jdormit:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

If tool results are not being included in a chat buffer, and the LLM makes a tool call then follows it up with a chat message, no space or newline gets inserted before the new message comes in. Here's an example where I asked the LLM to make a tool call then comment on the results:

```
### Try calling a tool then commenting about it so I can see how it outputs

I'll do a web search and then make a comment about the results.Based on the search results, I see that Emacs 29.4 is the latest version. This version is an emergency maintenance release.
```

This response has three messages: the first sentence, then the tool call, then the second sentence. Note that there is no space or newline between the two sentences: `...results.Based on...`.

I've been trying to fix this on my own for a bit but I'm not having much luck, `gptel--insert-response` is a bit too complicated for me to wrap my head around. I would appreciate it if you could point me in the right direction - happy to open a PR once I figure out what needs to change.

### Backend

Anthropic, but I would expect this to happen for any backend.

### Steps to Reproduce

1. Enable some tool (I used a web search tool) and set `gptel-include-tool-results` to `nil` (or `auto`, and make sure the tool doesn't have `:include` set)
2. Chat until the following sequence occurs: the LLM sends a chat message, then makes a tool call, then sends another chat message. I can consistently get Claude to do this by asking it to call a tool then comment on the results
3. Note that there is no space between the chat message before the tool call and the chat message after

### Additional Context

Emacs version: 29.1
GPTel commit: 55355bd

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

jdormit:
Oh I guess I should specify, I'm using streaming responses.

jdormit:
...and it turns out that was the key, streaming responses are inserted using a different function 🤦. Here's a fix: https://github.com/karthink/gptel/pull/770

PR #770, opened 2025-04-07
Insert newlines after streaming tool results

jdormit:
This is a fix for #769. It ensures that we insert a `gptel-response-separator` after a tool result comes in in streaming mode. This prevents weird formatting when `gptel-include-tool-results` is `nil` and the tool result came in between two chat responses (see the linked issue for details).

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -264,6 +264,9 @@ Optional RAW disables text properties and transformation."
              (set-marker-insertion-type tracking-marker t)
              (plist-put info :tracking-marker tracking-marker))
            (goto-char tracking-marker)
+           (when (plist-get info :last-was-tool-result)
+             (insert gptel-response-separator)
+             (plist-put info :last-was-tool-result nil))
            (unless raw
              (when transformer
                (setq response (funcall transformer response)))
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR.

This is trickier than it appears.  I need to test some edge cases when I can before merging.

jdormit:
Think I've run into one of those edge cases you were talking about @karthink: we should really be inserting the separator before a chat response, rather than after a tool response - otherwise multiple tool calls in a row leads to a bunch of newlines. Not sure if I'll have time to tackle that today but should be able to soon-ish.

gptel-curl.el
```diff
@@ -266,6 +266,9 @@ Optional RAW disables text properties and transformation."
              (set-marker-insertion-type tracking-marker t)
              (plist-put info :tracking-marker tracking-marker))
            (goto-char tracking-marker)
+           (when (plist-get info :last-was-tool-result)
+             (insert gptel-response-separator)
```
karthink, line 270: Here's another edge case.

When you test Claude it usually responds with text ("I'll run this tool for you"), then a tool call, then more text ("The results of the tool are X").

Suppose instead the LLM responds instead with a tool call followed by a response, as Gemini and OpenAI models do.  (No preamble before the tool call.)

Then you get a response that looks like this (with response-prefix "Response: ")
``` emacs-lisp
User: What is the time in Greece?

Response:

The time in Greece is ...
```

The extra `gptel-response-separator` between `Response: ` and the result is from your `:last-was-tool-call` tracking.

karthink:
Also, a similar fix might be needed for non-streaming responses.

karthink:
@jdormit I see a couple of new commits from you here.  Is the edge case from my comment above handled now?

karthink:
@jdormit I'd like to merge this PR but I'm not sure all the edge cases are covered.  If you're still interested in it, please let me know -- we'll need some tests to ensure that they are, and I can work on merging.

Please let me know if you've moved on as well, and I can close the PR.

jdormit:
Hey @karthink, this one very much dropped off my radar. To be honest I'm not sure I'll have the time to take it up again in the near future - feel free to take it over or close it out.

PR #775, opened 2025-04-11
docs: update xAI backend with Grok 3 model variants

axelknock:
Note: xAI has an image generation model, but it's served from a different endpoint, and I'm not sure how well it work with GPTel. I also opted to choose the "latest" variants for these models based off of the description on the model page:

[xAI Model Overview](https://docs.x.ai/docs/models)

### DIFF
--- a/README.org
+++ b/README.org
@@ -864,14 +864,9 @@ The above code makes the backend available to select.  If you want it to be the 
 
 Register a backend with
 #+begin_src emacs-lisp
-;; xAI offers an OpenAI compatible API
-(gptel-make-openai "xAI"           ;Any name you want
-  :host "api.x.ai"
-  :key "your-api-key"              ;can be a function that returns the key
-  :endpoint "/v1/chat/completions"
+(gptel-make-xai "xAI"                   ; Any name you want
   :stream t
-  :models '(;; xAI now only offers `grok-beta` as of the time of this writing
-            grok-beta))
+  :key "your-api-key")                  ; can be a function that returns the key
 #+end_src
 
 You can pick this backend from the menu when using gptel (see [[#usage][Usage]])
@@ -880,17 +875,11 @@ You can pick this backend from the menu when using gptel (see [[#usage][Usage]])
 
 The above code makes the backend available to select.  If you want it to be the default backend for gptel, you can set this as the value of =gptel-backend=.  Use this instead of the above.
 #+begin_src emacs-lisp
-;; OPTIONAL configuration
-(setq
- gptel-model   'grok-beta
- gptel-backend
- (gptel-make-openai "xAI"           ;Any name you want
-   :host "api.x.ai"
-   :key "your-api-key"              ;can be a function that returns the key
-   :endpoint "/v1/chat/completions"
-   :stream t
-   :models '(;; xAI now only offers `grok-beta` as of the time of this writing
-             grok-beta)))
+(setq gptel-model 'grok-3-latest
+      gptel-backend
+      (gptel-make-xai "xAI"               ; Any name you want
+                      :key "your-api-key" ; can be a function that returns the key
+                      :stream t))
 #+end_src
 
 #+html: </details>
--- a/README.org
+++ b/README.org
@@ -878,8 +878,8 @@ The above code makes the backend available to select.  If you want it to be the 
 (setq gptel-model 'grok-3-latest
       gptel-backend
       (gptel-make-xai "xAI"               ; Any name you want
-                      :key "your-api-key" ; can be a function that returns the key
-                      :stream t))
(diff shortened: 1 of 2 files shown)


karthink:
@axelknock This is a pretty long block of elisp to paste into the user's configuration.  What do you think about just adding a `gptel-make-grok` in `gptel-openai-extras.el` and reducing the configuration to just
```elisp
(gptel-make-grok "Grok" :key "...")
```

Also, gptel does not support image output so image generation models are completely unsupported and can be omitted from the configuration.

axelknock:
That sounds good, except that I would call it "xAI" instead of grok incase they change the model's name in the future. Note also that xAI's API is apparently unique in its ability to accept [unordered message types](https://docs.x.ai/docs/guides/chat#message-role-order-flexibility).

axelknock:
Squashed.

karthink:
Unlike other openai-derived backends (like Perplexity or Deepseek), `gptel-xai` is exactly the same as the `gptel-openai` type with no extra fields or special behavior.  So we don't need `gptel-xai` at all.  We can get rid of it and just use `gptel--make-openai` instead of `gptel--make-xai` in `gptel-make-xai`.  Could you make the change?

In the future if we add any xAI-specific features, I can define the derived type `gptel-xai`.

gptel-openai-extras.el
```diff
@@ -363,5 +363,71 @@ The Deepseek API requires strictly alternating roles (user/assistant) in message
...
+(cl-defun gptel-make-xai
+    (name &key curl-args stream key request-params
+          (header (lambda () (when-let* ((key (gptel--get-api-key)))
+                               `(("Authorization" . ,(concat "Bearer " key))))))
+          (host "api.x.ai")
+          (protocol "https")
+          (endpoint "/v1/chat/completions")
+          (models '((grok-3-latest
+                     :description "Grok 3"
+                     :capabilities '(text tool-use json)
```
karthink, line 381: The `text` capability is not recognized -- it is assumed that all models gptel supports output text.  Please remove text from the `:capabilities` lists.

gptel-openai-extras.el
```diff
@@ -363,5 +363,71 @@ The Deepseek API requires strictly alternating roles (user/assistant) in message
...
+                    (grok-2-vision-1212
+                     :description "Grok 2 Vision"
+                     :capabilities '(text tool-use json)
+                     :mime-types '("image/jpeg" "image/png" "image/gif" "image/webp")
+                     :context-window 32768
+                     :input-cost 2
+                     :output-cost 10))))
+  "Register an xAI backend for gptel with NAME."
+  (declare (indent 1))
+  (let ((backend (gptel--make-xai
```
karthink, line 416: Use `gptel--make-openai`, see comment in main thread.

gptel-openai-extras.el
```diff
@@ -363,5 +363,71 @@ The Deepseek API requires strictly alternating roles (user/assistant) in message
     (setf (alist-get name gptel--known-backends nil nil #'equal) backend)
     backend))
 
+;;; xAI
+(cl-defstruct (gptel-xai (:include gptel-openai)
```
karthink, line 367: The type `gptel-xai` is not required, see comment in main thread.

README.org
```diff
@@ -865,13 +865,10 @@ The above code makes the backend available to select.  If you want it to be the
...
-(gptel-make-openai "xAI"           ;Any name you want
-  :host "api.x.ai"
-  :key "your-api-key"              ;can be a function that returns the key
-  :endpoint "/v1/chat/completions"
+(gptel-make-xai "xAI"           ;Any name you want
   :stream t
-  :models '(;; xAI now only offers `grok-beta` as of the time of this writing
-            grok-beta))
+  :key "your-api-key"              ;can be a function that returns the key
+  )
```
karthink, line 871: Move paren to previous line, like
```
    :key "your-api-key")                     ;can be a function that returns the key
```

README.org
```diff
@@ -882,15 +879,12 @@ The above code makes the backend available to select.  If you want it to be the
...
  gptel-backend
- (gptel-make-openai "xAI"           ;Any name you want
-   :host "api.x.ai"
+ (gptel-make-xai "xAI"           ;Any name you want
    :key "your-api-key"              ;can be a function that returns the key
-   :endpoint "/v1/chat/completions"
    :stream t
-   :models '(;; xAI now only offers `grok-beta` as of the time of this writing
-             grok-beta)))
+   :models 'grok-3-latest))
```
karthink, line 887: This line is incorrect and also unnecessary now that we handle the models, it can be removed.

karthink:
A couple of README tweaks before I can merge this.

karthink:
Thank you for the PR @axelknock!

PR #777, opened 2025-04-11
allow customize-variable to set temperature to nil

nlevnaut:
Usually I set the temperature to nil to allow the backend to use its default temp, but I noticed customize-variable doesn't provide that as an option, and setopt complains about nil not being a number.


karthink:
Thanks for the PR @nlevnaut!

PR #780, opened 2025-04-12
Fix pairing of details tags

Arclite:
It looks like there was a mismatched copy/paste in #767 which caused the entire "Usage" section of the README to be hidden by default in GitHub's viewer (revealed only by expanding the Copilot Chat instructions). This is just a change to close the tag as needed and bring back the missing section.

Before:
image

After:
image


karthink:
Thank you @Arclite!

PR #781, opened 2025-04-12
gptel-gemini: Add support for Gemini 2.5 Pro Preview

benthamite:
This PR provides support for the Gemini 2.5 Pro Preview model. This models differs from Gemini 2.5 Pro Exp in that (1) it is paid rather than free and (2) it is not rate-limited to 5 requests per minute and 25 requests per day.


karthink:
Thanks @benthamite!

PR #789, opened 2025-04-16
gptel--gh-models: add new models

kiennq:


niv-l:
It's a shame that new models were dropped just after this PR was posted. Would love an update with the new `o3` and `o4-mini` models :)

karthink:
Thank you @kiennq!

karthink:
> It's a shame that new models were dropped just after this PR was posted. Would
> love an update with the new `o3` and `o4-mini` models :)

Added.

PR #794, opened 2025-04-18
gptel--make-tool: remove preprocess-args advice

kiennq:
This `cl-callf` advice somehow got optimized out (??) or just not working in native compilation.
It caused the `args` cleared out whenever I tried to add a new tool.
Since we're inside a same package, it would be better to call `gptel--preprocess-tool-args` on `gptel--make-tool` parameters directly.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1547,11 +1547,6 @@ feed the LLM the results.  You can add tools via
   (confirm nil :type boolean :documentation "Seek confirmation before running tool?")
   (include nil :type boolean :documentation "Include tool results in buffer?"))
 
-(define-advice gptel--make-tool (:filter-args (rest) preprocess-args)
-  "Convert symbol :type values to strings in the args in REST."
-  (cl-callf gptel--preprocess-tool-args (plist-get rest :args))
-  rest)
-
 (defun gptel--preprocess-tool-args (spec)
   "Convert symbol :type values in tool SPEC to strings destructively."
   ;; NOTE: Do not use `sequencep' here, as that covers strings too and breaks
@@ -1699,7 +1694,7 @@ callback as its first argument, which it runs with the result:
                    (lambda (_)
                      (let ((result (parse-this-buffer)))
                        (funcall callback result)))))"
-  (let* ((tool (apply #'gptel--make-tool slots))
+  (let* ((tool (apply #'gptel--make-tool (gptel--preprocess-tool-args slots)))
          (category (or (gptel-tool-category tool) "misc")))
     (setf (alist-get
            (gptel-tool-name tool)


karthink:
This won't work, since this preprocessing has to happen in the tool constructor `gptel--make-tool`, not in the user-facing `gptel-make-tool`.

The difference between the two is that `gptel--make-tool` creates unregistered tools, which is useful for gptel-requests that shouldn't modify gptel's state and don't use its UI..

kiennq:
Shouldn't `gptel-requests` send the tool with correct format already?
But we want to pre-process the input in case the user creates tool with some nuance difference in elisp (vector vs list, symbol vs string for name ...). After the processing happens, the tool is registered and then the next time when `gtpel--make-tool` is called, it's called with the correct parameters that we prepare.

kiennq:
The issue here is that gptel tool args doesn't work with native compilation now, that pretty much make gptel broken with tool use

karthink:
If `gptel--make-tool` doesn't run advice when native-compiled, why was it working until now?  I only changed a couple of details in `gptel--preprocess-tool-args` in the last few commits.

kiennq:
> If `gptel--make-tool` doesn't run advice when native-compiled, why was it working until now? I only changed a couple of details in `gptel--preprocess-tool-args` in the last few commits.
> 
> I'm trying to figure out why this is happening.

It runs the advice, but I think the preprocess function got collapsed somehow (from that `cl-callf`??) and always return nil for args.
It got away when I reevaluate the preprocess function again (not the define-advice one).
And the issue gone away if I change like in this function. It might be a bug on the native compiler but since we have a work-around and should avoid advice for internal function, may be that's a better approach?

karthink:
What's going on here is much more subtle.  See the latest commit message (b30cee0) for more details.  I need to ask about this on the emacs-devel mailing list.

Anyway, I removed the advice and added an extra function, please test.

kiennq:
Just saying but the advice did run for me, if it's not run the the `args` will retain value (working) instead of becomes `nil`.

What's not working is that the `preprocess` function under the native compilation decide to collapse and always set the args to nil. That might be because of native code optimization as it might see some functions / variable is not accessed and decide to optimize them out, leaving the default value of `nil`.

If I manually eval the `preprocess` function, not the `define-advice` clause, everything will work again.

karthink:
> What's not working is that the `preprocess` function under the native
> compilation decide to collapse and always set the args to nil. That might be
> because of native code optimization as it might see some functions / variable is
> not accessed and decide to optimize them out, leaving the default value of
> `nil`.

This problem was fixed right away, it's not the main issue.

> Just saying but the advice did run for me, if it's not run the the `args` will
> retain value (working) instead of becomes `nil`.

This problem is more complicated than this, whether or not the advice runs depends on many factors I don't understand yet.  Interpreted vs compiled is just one of the factors.  I'll wait to hear from emacs-devel.

It's fixed in gptel for now.

kiennq:
> This is the main problem -- the advice does not run reliably. This reason appears complicated. Whether or not the advice runs depends on many factors I don't understand yet. Interpreted vs compiled is just one of the factors. I'll wait to hear from emacs-devel.

I'm trying to say that the advice __run__ for me.
But the function which is set to the advice doesn't behave correctly.

If the function which is set to the advice (`gptel--preprocess-...`) is running correctly, it would be no issue since that function will do nothing in case of correct parameters are passed anyway.
However, since it doesn't behave correctly between the byte compile and native compile mode, it causes the issue.

I know the issue has been fixed, just want to clarify that piece of information since this doesn't relate to whether the advice on a function is working or not. It's the attached function that behaves incorrectly.
My guess here is that the `(cl-callf gptel--preprocess...)` is calling without the function quote for `gptel--preprocess...`, it affects the native compilation to think the `gptel--process-..` is not referred elsewhere and optmize it out.

karthink:
> I know the issue has been fixed, just want to clarify that piece of information since this doesn't relate to whether the advice on a function is working or not. It's the attached function that behaves incorrectly.

Indeed. I fixed this first, but then discovered that the advice function wasn't even being called some of the time.

> My guess here is that the (cl-callf gptel--preprocess...) is calling without the function quote for gptel--preprocess..., it affects the native compilation to think the gptel--process-.. is not referred elsewhere and optmize it out.

In the `cl-callf` macro-expansion, `gptel--preprocess-tool-args` is called as a regular function so this wasn't the cause.  `gptel--preprocess-tool-args` was setting its spec to `nil` sometimes.

I've asked about advising cl-struct constructors on emacs-devel: https://yhetil.org/emacs-devel/87r01oudee.fsf@gmail.com

PR #801, opened 2025-04-21
Improve gptel-org-set-topic to not include tags in the default value

akirak:
If the source heading has any tag, the result of `org-get-heading` contains the tag string. By using `org-entry-get` to retrieve the `ITEM` property of the entry, you can get the true heading of the entry, which would be more suitable unless you want to include the tags in the topic.

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -184,7 +184,7 @@ current heading and the cursor position."
                                  (substring-no-properties
                                   (replace-regexp-in-string
                                    "\\s-+" "-"
-                                   (org-get-heading)))
+                                   (org-entry-get nil "ITEM")))
                                  50))))))
   (when (stringp topic) (org-set-property "GPTEL_TOPIC" topic)))
 


karthink:
Thanks for the PR.  Any idea if this is preferable to  `(org-get-heading t t t t)`?

akirak:
> Thanks for the PR. Any idea if this is preferable to (org-get-heading t t t t)?

`org-entry-get` utilizes org-element cache when available, so it should be slightly faster than `org-get-heading`.

The following benchmark is unrealistic, but it indicates they have different performance profiles:

```emacs-lisp
(benchmark-run 10000
  (org-entry-get nil "ITEM"))
```

> | 0.13730204799999998 | 0 | 0.0 |

```emacs-lisp
(benchmark-run 10000
  (org-get-heading t t t t))
```

> | 0.27516740100000003 | 1 | 0.12469920199999951 |

karthink:
@akirak merged, thanks for the comparison.

PR #809, opened 2025-04-26
Allow uppercase characters when creating tool key shortcuts

kiennq:
This PR allows the use of uppercase characters when creating key shortcuts for tools. 
This would be beneficial for tools like github mcp which provides more than 26 tools

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -803,9 +803,11 @@ only (\"oneshot\")."
         (cl-loop                    ;for each category, collect tools as infixes
          for (name . tool) in tools-alist
          with tool-keys = (delete category-key (nconc (number-sequence ?a ?z)
-                                                      (number-sequence ?0 ?9)))
-         for tool-key = (seq-find (lambda (k) (member k tool-keys)) name
-                                  (seq-first tool-keys))
+                                                      (number-sequence ?0 ?9)
+                                                      (number-sequence ?A ?Z)))
+         for tool-key = (or (seq-find (lambda (k) (member k tool-keys)) name)
+                            (seq-find (lambda (k) (member k tool-keys)) (upcase name)
+                                      (seq-first tool-keys)))
          do (setq tool-keys (delete tool-key tool-keys))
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR @kiennq.

A couple of notes:

- The general convention I follow in gptel is that (with `x` = any lower case char, `X` = any upper case char)
  - `-x` sets a variable, i.e. some state that persists after the menu is closed,
  - `x`sets a transient menu infix option (i.e. not a suffix that quits the transient)
  - `X` or a special key (like `RET`) runs a suffix and quits the menu.
  Going by this convention, we should not use upper case letters for selecting tools.
  
  That said, I just broke this convention in the `gptel-integrations` menu items, so it's not a hard rule.
  
- However, a more serious problem is that the tools menu is almost unusable when you have a couple of MCP servers providing 25 tools each.  Because it doesn't scale, we are thinking about replacing individual tool selection with `completing-read-multiple` in #653.  Categories (i.e. all tools in the category) will still be selected with transient menu key bindings.

kiennq:
> * However, a more serious problem is that the tools menu is almost unusable when you have a couple of MCP servers providing 25 tools each.  Because it doesn't scale, we are thinking about replacing individual tool selection with `completing-read-multiple` in [gptel-transient: tool selection using completing-read-multiple #653](https://github.com/karthink/gptel/pull/653).  Categories (i.e. all tools in the category) will still be selected with transient menu key bindings.

I have the tool selection using completing-read-multiple enabled as well, however I still find myself time to time go to the full table view of all tools, especially for a quick glance and quickly enable a tool in the list.
Another thing is that the current table view provides a quick visual check of which tools have been enabled, and I think that's a good thing to keep.
Here is the commit that follows the completing-read-multiple that I used btw https://github.com/kiennq/gptel/commit/e35cec00170b31aee75b0895ef500b1dd3e6a823

karthink:
Closing as uppercase characters were added a while ago.  Further, the tools menu design has changed significantly so this no longer applies.

Issue #750, opened 2025-03-26
Gemini Google Search?


ekenberg:
Looking at [the documentation for Gemini API](https://ai.google.dev/gemini-api/docs/grounding?lang=rest#configure-search) I see that Google Search can be used as a tool even with the REST API. Does that mean that the Google search is run on the server and not locally? And could we get this functionality using Gemini with gptel?

I tried adding
```
"tools": [
          {
              "google_search": {}
          }
      ]
```
to the JSON request sent by gptel but that resulted in a `HTTP/2 404` error `"Could not parse HTTP response"`.

Sorry if I don't understand how this works. Web search enhanced Gemini answers in gptel chat would be so awesome!

knightsmove:
I tried :around advice on the Gemini-specific gptel--request-data.  Couldn't get the elisp data structure to serialize into the correct JSON using json-encode. I can barely code in elisp - just vibe-coded a couple hours with gemini-2.5-pro-exp-03-25 but failed. Would really rock if someone got this to work.

eslgastal:
This seems to work:

```elisp
(gptel-make-gemini "Gemini"
    :stream t
    :key "YOUR_GEMINI_API_KEY"
    :request-params '(:tools [(:google_search ())])
    )
```

But makes tool-use not available. I think the correct procedure would be to add the "google_search" tool directly in gptel--parse-tools for gptel-gemini backend, based on a configurable option.

avishefi:
@eslgastal looks like this is mutually exclusive with adding gptel tools, any idea how to combine these?

eslgastal:
> [@eslgastal](https://github.com/eslgastal) looks like this is mutually exclusive with adding gptel tools, any idea how to combine these?

I think you could explicitly add the google_search tool to the [return value of gptel--parse-tools for the gptel-gemini backend](https://github.com/karthink/gptel/blob/0da503faa08ab2320d272e407f93e3a1bdf6ba01/gptel-gemini.el#L226).

avishefi:
You mean using an advice?

cashpw:
Ading the `:google_search` around Gemini's `gptel--request-data` does the trick.

```elisp
(cl-defmethod gptel--request-data :around ((backend gptel-gemini) prompts)
  "Add search ability."
  (plist-put (cl-call-next-method) :tools (append '(:google_search ()))))
```

stefanv:
That adds search grounding, but it also seems to mess up other tools like macher?

[...]

PR #810, opened 2025-04-27
Gemini: parse code execution and grounding result

blahgeek:
(I'm not sure if this feature is desired or not. This is a draft version. Please let me know how you think.)

Gemini model support builtin tools including google search (grounding) and code execution. Now users can use `:request-params` to add support for them (https://github.com/karthink/gptel/issues/750#issuecomment-2815590261 ). This PR extends the parsing function for Gemini to print grounding and code execution results. (We may need better ways other than custom `:request-params` to support these tools, but that's for future PRs)

Screenshot 2025-04-27 at 21 55 21
Screenshot 2025-04-27 at 21 55 56

### DIFF
--- a/gptel-gemini.el
+++ b/gptel-gemini.el
@@ -72,36 +72,70 @@ Mutate state INFO with response metadata.
 If INCLUDE-TEXT is non-nil, include response text in the prompts
 list."
   (let* ((cand0 (map-nested-elt response '(:candidates 0)))
-         (parts (map-nested-elt cand0 '(:content :parts))))
+         (parts (map-nested-elt cand0 '(:content :parts)))
+         (grounding-meta (plist-get cand0 :groundingMetadata)))
     (plist-put info :stop-reason (plist-get cand0 :finishReason))
     (plist-put info :output-tokens
                (map-nested-elt
                 response '(:usageMetadata :candidatesTokenCount)))
-    (cl-loop
-     for part across parts
-     for tx = (plist-get part :text)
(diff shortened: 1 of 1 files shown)


karthink:
@blahgeek Thanks for the PR.  This feature is welcome.  Can the codeExecution parsing be folded into the main cl-loop?  Otherwise this looks fine.

karthink:
> (We may need better ways other than custom `:request-params` to support these tools, but that's for future PRs)

It's best to expose these features as backend-specific tools, I think.  There's no way in gptel or its UI to have tools that only work with one backend, so yeah, `:request-params` will do for now.

blahgeek:
Thanks. I've updated the code according to your suggestion.

eslgastal:
@blahgeek Nice! I think it would be interesting to have an option to extract the target redirected URL from the obfuscated "vertexaisearch.cloud.google.com" URLs. In Linux I've been doing this with `curl -s -o /dev/null -w '%header{location}' --head <URL>`.

Also, a couple of weeks ago I hacked together the insertion of grounding URLs with the associated citations throughout the LLM response (which shows where each grounding applies in the response). Unfortunately gptel's markdown->org conversion makes the :startIndex and :endIndex returned by Gemini incorrect for inserting the citations.

Here's the code, which may help in this PR:

https://gist.github.com/eslgastal/6700418bbad6ff9e49a4d6e28fe3dd83

And a screenshot:

gemini_grounding

avishefi:
This looks great, are there any blockers to merging this?

PR #815, opened 2025-04-29
Add AUGMENT stage to request pipeline

jwiegley:
This PR lays the groundwork for implementation many forms of RAG in gptel, including:

- Document inclusion, as selected by vector similarity
- "Agents" that augment the context when referenced
- "Dynamic context" that modifies the context is response to inputs or conditions

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -2451,13 +2451,13 @@ be used to rerun or continue the request at a later time."
   (let ((data (plist-get info :data)))
     (if (not (plist-member data :args))
         info
-      (let (((full-prompt (plist-get data :full-prompt))
-             (stream (plist-get data :stream))
-             (callback (plist-get data :callback))
-             (context (plist-get data :context))
-             (in-place (plist-get data :in-place))
-             (gptel-include-reasoning (plist-get data :gptel-include-reasoning))
-             (gptel-use-tools (plist-get data :gptel-use-tools))))
+      (let ((full-prompt (plist-get data :full-prompt))
+            (stream (plist-get data :stream))
+            (callback (plist-get data :callback))
+            (context (plist-get data :context))
+            (in-place (plist-get data :in-place))
+            (gptel-include-reasoning (plist-get data :gptel-include-reasoning))
+            (gptel-use-tools (plist-get data :gptel-use-tools)))
         ;; overwrite the "initial arguments" with the realized data
         (plist-put info :data (gptel--request-data backend full-prompt))
         (when stream (plist-put info :stream t))
@@ -2473,8 +2473,9 @@ be used to rerun or continue the request at a later time."
 
 (defsubst gptel-prepare-fsm (fsm)
   "Prepare the FSM for operation, preparing all context arguments."
-  (setf (gptel-fsm-info fsm)
-        (gptel-realize-info (gptel-fsm-info fsm))))
+  (let ((info (gptel-fsm-info fsm)))
+    (setf (gptel-fsm-info fsm)
+          (gptel-realize-info (plist-get info :backend) info))))
 
 (defun gptel-abort (buf)
   "Stop any active gptel process associated with buffer BUF.
--- a/gptel.el
+++ b/gptel.el
@@ -2385,13 +2385,8 @@ be used to rerun or continue the request at a later time."
   (gptel--sanitize-model)
   (let* ((directive (gptel--parse-directive system))
          ;; DIRECTIVE contains both the system message and the template prompts
-         (system-message
-          ;; Add context chunks to system message if required
-          (unless (gptel--model-capable-p 'nosystem)
-            (if (and gptel-context--alist
-                     (eq gptel-use-context 'system))
-                (gptel-context--wrap (car directive))
-              (car directive))))
+         (system-message (car directive))
+         (collection (gptel-context--collect))
          ;; TODO(tool) Limit tool use to capable models after documenting :capabilities
          ;; (gptel-use-tools (and (gptel--model-capable-p 'tool-use) gptel-use-tools))
          (stream (and stream gptel-use-curl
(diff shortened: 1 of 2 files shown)


gptel.el
```diff
@@ -1991,6 +1997,97 @@ buffer."
...
+(defun gptel--augment-info (fsm fns)
+  "Augment the request contained in state machine FSM's info.
+
+Each function receives a CALLBACK and an INFO, and must call the
+callabck with their augemented version of the INFO (or the
+original INFO) once completed. This permits augmentations to be
+asynchronous.
+
+Returns the transformed info at the end."
+  (if (null fns)
```
karthink, line 2078: This causes the augmentors to run sequentially -- not an issue for synchronous augmentors, but is unnecessarily slow if you have async ones.

I thought about why you might have designed it this way instead of using a counting callback and running through `gptel-augment-handler-functions` with `run-hook-with-args` -- the only reason I can think of is that alterations to `fsm` might be noncommutative, and that the ordering of `gptel-augment-handler-functions` might not be respected when there are async callbacks involved.  Is this something you foresee as a problem?

Otherwise we can switch to a counting callback, which can shorten the augment wait considerably.

gptel.el
```diff
@@ -1991,6 +1997,97 @@ buffer."
...
+      (gptel--finish-augmentation fsm)
+    (funcall (car fns)
+             #'(lambda (fsm) (gptel--augment-info fsm (cdr fns)))
+             fsm)))
+
+(defun gptel--handle-augment (fsm)
```
karthink, line 2084: If we switch to a counting callback in `gptel--augment-info` (see comment), we don't need a (recursive) `gptel--augment-info` or `gptel--finish-augmentation` any more, and  `gptel--handle-augment` can handle the whole augment process, localizing all the augment code.

gptel.el
```diff
@@ -2379,53 +2471,83 @@ be used to rerun or continue the request at a later time."
                  gptel-backend (and gptel--num-messages-to-send
                                     (* 2 gptel--num-messages-to-send))))))
             ((consp prompt) (gptel--parse-list gptel-backend prompt)))))
-         (info (list :data (gptel--request-data gptel-backend full-prompt)
+         (info (list :data (list :args t
+                                 :full-prompt full-prompt
```
karthink, line 2475: The main problem here is that `full-prompt` is not a backend-agnostic format -- it can be of a different type/schema for each backend.  This makes modifying the prompt in an augmentor difficult.

This can be papered over somewhat with `gptel--inject-prompt`, which can act as a backend-agnostic interface.  Even if I add `gptel--retrieve-prompt` and make these `setf`able, it's not a great interface.

My original idea was that the copied buffer itself is the backend-agnostic prompt list, and another intermediate, universal format for prompts between it and the final messages array is unnecessary.  (It can also produce a lot of garbage.)

Now we need a new backend-agnostic prompts list format.  Note that it can include many roles: `user`, `llm`/`assistant`, `tool-call`, `tool-result` and possibly more in the future.

I still think using the buffer can work because it's a simple data structure and Emacs is great at modifying it.  Adding text to the top and bottom of the buffer (corresponding to before the first messages array and after the last) is very simple, as is deleting or adding text anywhere in the middle.  This doesn't have to be done via `gptel-prompt-filter-hook` -- it can be part of the augment process.

gptel.el
```diff
@@ -2379,53 +2471,83 @@ be used to rerun or continue the request at a later time."
...
-    (when (and gptel-use-tools gptel-tools)
-      (plist-put info :tools gptel-tools))
-    ;; Add info to state machine context
     (setf (gptel-fsm-info fsm) info))
   (unless dry-run (gptel--fsm-transition fsm)) ;INIT -> WAIT
   fsm)
 
+(defun gptel--realize-info (info)
+  (let ((data (plist-get info :data)))
+    (if (not (plist-member data :args))
```
karthink, line 2494: What is the purpose of `:args` here?

gptel.el
```diff
@@ -2379,53 +2471,83 @@ be used to rerun or continue the request at a later time."
                  gptel-backend (and gptel--num-messages-to-send
                                     (* 2 gptel--num-messages-to-send))))))
             ((consp prompt) (gptel--parse-list gptel-backend prompt)))))
-         (info (list :data (gptel--request-data gptel-backend full-prompt)
+         (info (list :data (list :args t
```
karthink, line 2474: It's not clear to me yet how to avoid this step of storing and rehydrating this state in `:data`.  I'll continue to think about this.  But at minimum I think capturing many of these keys is unnecessary since you pass `info` (actually the whole `fsm`) to the augmentors, and they are available as `:keys` inside `info` in the current implementation on master.

- `:context`: Accessible inside `info` as `:context` anyway.
- `:gptel-include-reasoning`: Available as `:include-reasoning` in `:info`.
- `:callback`: Do we expect the augmentors to change the callback?
- `:in-place`: Available as `:in-place` in `:info`.

gptel.el
```diff
@@ -2829,7 +2951,13 @@ the response is inserted into the current buffer after point."
                              (kill-buffer buf)))
                          nil t nil)))
       ;; TODO: Add transformer here.
-      (setf (alist-get proc-buf gptel--request-alist) fsm))))
+      (setf (alist-get proc-buf gptel--request-alist)
+            (cons fsm
```
karthink, line 2955: We don't need to capture `proc-buf` into the closure, since this information is found in `gptel-abort` anyway and we can pass it as an argument to the abort-callback.  This should make the callbacks "static" functions that can be native-compiled.

karthink (commented):
Recording some thoughts/questions about the code, let me know what you think.

I can start making some changes to simplify this patch soon, can I push them to your branch?

jwiegley:
> Recording some thoughts/questions about the code, let me know what you think.
> 
> I can start making some changes to simplify this patch soon, can I push them to your branch?

I will read through your changes in more detail tomorrow; but in the meantime, please push whatever changes you wish!

karthink:
Thanks John.

The `feature-augment-buffer` branch, which used this PR for inspiration, is ticking along well.  A couple of commits from this branch have been added there.  But since this approach is otherwise incompatible with the prompt-buffer-based approach planned for gptel, I'm closing it.

jwiegley:
Understood, I prefer the buffer-based approach for a whole bunch of reasons.

Issue #688, opened 2025-03-08
No Response When Using `gptel-make-tool` with Zero Arguments


norisuke3:
## Description
I'm trying to create a simple tool with `gptel-make-tool` that takes no arguments, but I'm not getting any response when attempting to use it. When I try a similar tool that takes arguments, it works fine.

## Environment
- gptel package-version: 20250306.310
- Provider: OpenRouter
- Model: anthropic/claude-3.7-sonnet

## Reproduction Steps

1. Created a tool with arguments that works fine:
```elisp
(gptel-make-tool
 :name "greeting_with_a_name"
 :function (lambda (name) (concat name " Hello World!"))
 :description "A simple tool that returns a greeting message."
 :args (list '(:name "name"
               :type string
               :description "the name of the person to which the greeting is made for"))
 :category "debug")
```

2. Created a similar tool with no arguments:
```elisp
(gptel-make-tool
 :name "greeting_witout_a_name"
 :function (lambda () "Hello World!")
 :description "A simple tool that returns a greeting message."
 :args nil
 :category "debug")
```

## Expected Behavior
The tool without arguments should respond with "Hello World!" when invoked.

## Actual Behavior
When trying to use the zero-argument tool, there's no response. The OpenRouter Activity page shows no activity either, suggesting the request isn't being sent or is being rejected.

## Additional Information
I suspect there might be an issue with the `:args nil` setting, as the only difference between the working and non-working tools is the argument handling.

Any guidance would be appreciated.

**Additional context**
Emacs version: GNU Emacs 30.1 (build 1, aarch64-apple-darwin21.6.0, NS appkit-2113.65 Version 12.7.6 (Build 21H1320)) of 2025-02-24
Operating system: MacOS sequoia 15.3.1

norisuke3:
Just found that it's a model issue, not gptel. It works with the following models:
`openai/gpt-4o-2024-11-20`
`google/gemini-2.0-flash-001`

Somehow `openai/o3-mini` go into a loop until killing a gtpel session (*OpenRouter*).

A screen shot of Openrouter activity
<img width="1110" alt="Image" src="https://github.com/user-attachments/assets/391d9682-1215-4969-bf71-361bfa668ed7" />

**\*gptel-curl\***
```
HTTP/2 200 
date: Sat, 08 Mar 2025 20:32:34 GMT
content-type: text/event-stream
access-control-allow-origin: *
cache-control: no-cache
[...]

PR #818, opened 2025-04-30
send parameter in request when args is nil

longlene:


karthink:
As you can see from the comment in the original code, OpenAI complains if the function takes no arguments but the tool spec contains a `:parameters` field.  From #688, it looks like some OpenAI-compatible APIs (like Grok?) complain if the tool spec does _not_ contain a `:parameters` field.

So this needs to be tested carefully against all OpenAI compatible APIs.

longlene:
@karthink Thank you for the explanation. 
If so, maybe we can add a defcustom variable, gptel will still not cast `parameters` field by default, then we can customize the value to change the behavior.
What do you think?

longlene:
At least, llama-server expect the `parameters` field: https://github.com/ggml-org/llama.cpp/blob/13c9a3319b65469e883c49dd1c478abedc410157/common/chat.cpp#L232

longlene:
@karthink I updated the code, the default behavior will be still the same as before, then for some user who knows what they are doing, there is still a way for them to always send parameters when needed.

karthink:
@longlene As I mentioned above, a user option is not the solution here.

Could you check if the OpenAI API works correctly after your original change?  It did not use to, but things might be different now.  If it works with `:parameters` specified, we merge the change without the user option.

Also `:parameters` should probably be set to `:null`, not `nil`.

longlene:
@karthink ok, let me see if I can test it with OpenAI API.

longlene:
(list :paramters nil) would cause gptel send `paramters: {}` to the API.

karthink:
> (list :paramters nil) would cause gptel send `paramters: {}` to the API.

Yes.  In your testing, do `null` and `{}` both work?

longlene:
@karthink Yes, both OpenAI and llama-server can work with paramters: {} and null. I updated pr as null.

longlene:
In the log file I uploaded, the value is null. gpt can call `git_status` tool correctly.

karthink:
@longlene I'm merging this PR now, but please stick around in case OpenAI tool calls for functions without args start failing.

PR #825, opened 2025-05-05
Added instructions on how to install via use-package

dotemacs:
The change is super small, but figured why not have it, since `use-package` is arguably default package manager.


karthink:
`use-package` is not a package manager, and the provided code may not work depending on the value of `use-package-always-ensure`.

Moreover, the user will most probably be copying a use-package block with the setup instructions for their backend of choice from further below in the README anyway.

Closing as not planned.

PR #828, opened 2025-05-06
gptel-openai: gptel--request-data get reasoning capability from prop

ahmed-shariff:
The models with reasoning capabilities were hard-coded in openai's `gptel--request-data`. This PR replaces it with `gptel--model-capable-p` to get the reasoning capabilities from the model properties.

This came about when I started playing with `gpt-4o-search-preview` and `gpt-4o-mini-search-preview`, which seems to not support `temperature`. I was adding the reasoning capability, which seems to resolve that issue for now. Is it also worth considering `temperature` as another capability?

### DIFF
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -296,8 +296,8 @@ Mutate state INFO with response metadata."
          `(:model ,(gptel--model-name gptel-model)
            :messages [,@prompts]
            :stream ,(or gptel-stream :json-false)))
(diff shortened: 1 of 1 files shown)


karthink:
The `reasoning` capability in the model metadata is not currently used for anything.  Using it to determine if certain options apply will break `gptel-make-openai` use with all OpenAI-compatible backends that provide access to reasoning models.

Closing as not planned.

PR #830, opened 2025-05-07
gptel-openai: Prevent null tool-calls from causing error

ragnard:
Ensure that tool-calls are processed only if they are not null. This issue was discovered when using LiteLLM, and is possibly an implementation bug there. At least differs from the real OpenAI.com API, which doesn't have this issue. However, for gptel to be usable with LiteLLM, this check is necessary.

### DIFF
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -260,7 +260,8 @@ Mutate state INFO with response metadata."
     ;; OpenAI returns either non-blank text content or a tool call, not both.
     ;; However OpenAI-compatible APIs like llama.cpp can include both (#819), so
     ;; we check for both tool calls and responses independently.
-    (when-let* ((tool-calls (plist-get message :tool_calls)))
+    (when-let* ((tool-calls (plist-get message :tool_calls))
+                ((not (eq tool-calls :null))))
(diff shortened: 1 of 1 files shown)


karthink:
@ragnard Thanks for the PR.

Is a similar fix needed for streaming tool calls, or does LiteLLM not support streaming tool calls?

ragnard:
Not sure tbh, and not really sure how to test it either, but I will try to take a look tomorrow.

Thanks for gptel by the way, really fantastic software. I made a tiny extension ([gptel-magit](https://github.com/ragnard/gptel-magit/)) that extends magit with commands to generate commit messages (like gptcommit, but different) and using gptel programatically was a joy too.

karthink:
@ragnard Any idea about the non-streaming case?

If not we don't need to wait, I can merge this PR and add a todo to the non-streaming parser for later.

muep:
So like the reporter of #848, I was running into a similar problem with the Mistral API. Merging this on top of current `master` seemed to fix the problem there as well.

karthink:
Thanks for the PR @ragnard.  I'll keep an eye out for reports of failure in the streaming case.

PR #832, opened 2025-05-08
C-y can be used to add the current kill to the context

jwiegley:
This saves the user from having to first create a new buffer, yank the current kill into that buffer with C-y, and then add that buffer as context. With this change, they can just type C-y in the transient gptel-send buffer and the temporary is created as needed.

If there are already kills that have been yanked, the new kills will be added to the context. In order to reset this buffer, use C-u C-y when yanking from the transient menu.


jwiegley:
@karthink I'm thinking now that `C-y` should always just paste the current kill-ring as context, and they `C-u C-y` should append. Otherwise, you have to remember to reset the context as you switch queries.

gptel-context.el
```diff
@@ -84,6 +84,23 @@ context chunk.  This is accessible as, for example:
   :group 'gptel
   :type 'function)
 
+(defun gptel-context-add-current-kill (&optional arg)
+  "Add current-kill to gptel, accumulating if arg is non-nil"
+  (interactive "P")
+  (let ((kill (current-kill 0)))
+    (with-current-buffer (get-buffer-create "*current-kill*")
```
karthink, line 91: Could we rename this to `*gptel-current-kill*` or `*gptel-kill-ring-context*`, something like that?  It should be easy to identify the provenance of this buffer in the buffer list.

On another note, we don't need undo enabled in this buffer and it can be a hidden buffer, so the name should start with a space, like ` *gptel-current-kill*`.

gptel-context.el
```diff
@@ -84,6 +84,23 @@ context chunk.  This is accessible as, for example:
...
+      (if (not arg)
+          (kill-region (point-min) (point-max))
+        (goto-char (point-max))
+        (unless (bobp)
+          (unless (bolp)
```
karthink, line 96: I don't understand the double `(unless (bolp) ...)`, but I'm assuming it's required?

gptel-transient.el
```diff
@@ -1128,6 +1129,16 @@ supports.  See `gptel-track-media' for more information."
 (declare-function gptel-context--at-point "gptel-context")
 (declare-function gptel-add "gptel-context")
 
+(transient-define-suffix gptel--infix-context-add-current-kill (&optional arg)
+  "Add current kill to gptel's context."
+  :transient 'transient--do-stay
+  :key "C-y"
+  :description
```
karthink, line 1136: Please make this an expert-command, as I'm trying to keep first-time users from running away after looking at the menu.

You can add
```emacs-lisp
:if (lambda () gptel-expert-commands)
```
to the transient-suffix specification.

jwiegley: This is a really good point.

jwiegley:
It looks like this would would be a way of adding an “invisible” suffix, but I can’t find how to do that.

Interestingly, Preplexity is hallucinating _hard_, after many attempts at correction, and insists that `transient-define-prefix foo` creates a keymap named `foo-map` where you could bind a key to do just this. But, you can’t.

PR #834, opened 2025-05-10
gptel-gemini: Fix filtering of unsupported attributes in tool calls

necaris:
Previous implementation (#827) didn't recurse correctly into nested sub-schemas -- this does a better job.

### DIFF
--- a/gptel-gemini.el
+++ b/gptel-gemini.el
@@ -147,20 +147,21 @@ list."
   "Destructively filter unsupported attributes from SCHEMA.
 
 Gemini's API does not support `additionalProperties'."
+  (cl-remf schema :additionalProperties)
   (when (plistp schema)
     (cl-loop for (key val) on schema by #'cddr
              do (cond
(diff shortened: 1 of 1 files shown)


karthink:
@necaris: Thank you for the PR!

I haven't tested it yet, but it looks fine so I'm inclined to merge it.

However, could you add a couple of unit tests to check that this function is doing what it says?

You can raise a PR (separately) against the `gptel-test` repo with the unit tests, OR just include the `ert-deftest` forms here and I can add it, OR (if you're not comfortable with `ert-deftest`) give me a couple of sample inputs and expected outputs, including a case where no modification is expected.

necaris:
Thanks for the quick review! I went ahead and raised the PR on https://github.com/karthink/gptel-test -- https://github.com/karthink/gptel-test/pull/5 _should_ do what it says on the tin, but I'd be grateful for another pair of eyes.

karthink:
Thank you @necaris!

PR #836, opened 2025-05-12
gptel-gh: Add media capability to GitHub Copilot gpt-4.1 model

huonw:
The gpt-4.1 models appear to support media/"vision" when used via GitHub Copilot, and thus this PR aligns the capabilities of this model in `gptel--gh-models` to the capabilities of this model in `gptel--openai-models`.

### DIFF
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -41,7 +41,7 @@
      :cutoff-date "2023-10")
     (gpt-4.1
      :description "Flagship model for complex tasks"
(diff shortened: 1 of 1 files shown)


karthink:
@huonw Thank you!  This is actually unnecessary since the capabilities are stored as symbol properties and thus shared between gptel-openai and gptel-gh, since they use the same model symbols.  But this does fix a potential inconsistency where the capabilities are determined by the order in which you load those two libraries..

PR #845, opened 2025-05-14
gptel--display-tool-calls: allow to limit tool display length

kiennq:
This would be useful to allow to peek at called tools and its result


karthink:
Thanks for the PR @kiennq.

When included in the buffer, the tool result is fed back to the LLM in all subsequent turns.  gptel does not maintain the state of the conversation anywhere outside of the buffer.  The tool result is thus the source of truth for the LLM, and should not be truncated.

kiennq:
You're right. I should just implement the folding for markdown case instead.
Since there's no folding for markdown now, I've implemented a simple `gptel--hs-fold`, hope it's okay.

karthink:
I will review this when I have more time, but folding tool blacks in markdown buffers is a welcome feature.

I have some concerns about the spacing changes around the `begin_tool` block -- there are _many_ edge cases here.  I'll test it when I can.

karthink:
@kiennq I'm taking a look at this PR now with the intention to merge soon.  Here are a couple of suggestions:

1. Please don't touch any of the org-mode handling in this PR, **even if the code looks wrong**.  An example of such a change is in lines 3516-3521.  Another example is at line 3549, where you changed `(forward-line -1)` to `(search-backward ...)`.
2. ~~You've tried to unify the tool block folding behavior for Markdown and Org-mode in `gptel--hs-fold` by creating a new overlay with the invisible property.  This is not a good idea -- Org's folding is very complicated and there is much more to it than that.  It is  best done by Org functions like `org-cycle`.  Please restore the folding handling for Org mode and leave it untouched.~~

EDIT: Sorry, I read the code wrong, Org's folding is still handled by `org-cycle`, so it's just item 1.

I can look at the Markdown-specific changes after this.

karthink:
I also had a look at `gptel--hs-fold`.  Some notes:
- It hides the block but I found no way to reveal it except by running `hs-show-block`.
- When I run `hs-show-block`, the overlay is deleted and there's no way to hide it again.  I tried `hs-hide-block` and `hs-hide-all` but it doesn't work since the `hs-` properties are gone.

What would be good is a way to show hide the tool block like in Org-mode, with just `TAB`.

i also think `gptel-display-max` is not required.  We can just fold to the first line of the tool block, to
`````
``` tool (read_file :filepath "/path/...")…
`````
like we do in Org mode.  Pressing `TAB` on  this line should show the block, pressing `TAB` again should hide it.

karthink:
@kiennq are you still interested in this PR?

karthink:
Implemented in 9d685b9b3400337d6137a2c94903e15e18736aea.

PR #850, opened 2025-05-20
Make streaming for rewrites an independent variable

jwiegley:


dschrempf:
I am supporting this PR.
On my setup, I have to set `gptel-stream` to `nil` even though I have disabled streaming in the backend, otherwise rewriting does not work.

PR #856, opened 2025-05-22
Update gemini-2.5-flash-preview to latest model code.

bjodah:
I guess there is no rush since the 04-17 endpoint still works (not sure if that is still the old model or if it is already being redirected to the new version). But eventually the old preview models are usually phased out leading to 404 or similar.

See:
 - https://ai.google.dev/gemini-api/docs/models#gemini-2.5-flash-preview
 - https://ai.google.dev/gemini-api/docs/models#previous-experimental-models

EDIT: ~~Or perhaps I should duplicate the existing entry? Would be less disruptive to peoples configs I guess?~~ ✓

### DIFF
--- a/gptel-gemini.el
+++ b/gptel-gemini.el
@@ -463,6 +463,15 @@ files in the context."
      :input-cost 0.15
      :output-cost 0.60 ; 3.50 for thinking
      :cutoff-date "2025-01")
+    (gemini-2.5-flash-preview-05-20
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR.  Yes, please add this as a new option instead, we want to avoid breaking people's configs.

bjodah:
I rebased as an additional entry. Tried this branch in my setup and it seems to work alright.

karthink:
Thank you @bjodah.

PR #867, opened 2025-05-23
AWS Bedrock support (follow up from #670)

akssri:
This PR is a follow-up from #670 

Modifications to the parent PR's commits:
- uses call-next-method instead of gptel--tool-args-to-json-schema
- handle user-provided curl-args in backend
- remove white-space changes

New commits:
- Add AWS profile support

TODO:
- Add documentation blurb

PR #871, opened 2025-05-23
GPTEL changes for bedrock

akssri:
GPTEL-only changes for bedrock.
(follow-up from #867 )

### DIFF
--- a/gptel-curl.el
+++ b/gptel-curl.el
@@ -137,13 +137,13 @@ the response is inserted into the current buffer after point."
          (args (gptel-curl--get-args info token))
          (stream (plist-get info :stream))
          (process (apply #'start-process "gptel-curl"
-                         (gptel--temp-buffer " *gptel-curl*") "curl" args)))
+                         (gptel--temp-buffer " *gptel-curl*") (gptel--curl-path) args)))
     ;; Don't try to convert cr-lf to cr on Windows so that curl's "header size
     ;; in bytes" stays correct. Explicitly set utf-8 for non-win systems too,
     ;; for cases when buffer coding system is not set to utf-8.
     (set-process-coding-system process 'utf-8-unix 'utf-8-unix)
     (when (eq gptel-log-level 'debug)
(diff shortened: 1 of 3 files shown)


karthink:
Made some minor changes.

I think the only thing missing from this PR is `url-retrieve` handling.  Have you tested it with `gptel-use-curl` set to `nil`?  I expect that it won't work, as this seems to need some special Curl flags.

So one of two things is required:

- The bedrock backend is made to work with `url-retrieve` in `gptel--url-get-response`, making further changes to gptel.  This outcome would be preferable.  But it doesn't seem like running `set-buffer-multibyte` in the `url-retrieve` buffer will be enough to support bedrock.

 OR

- It is made clear that gptel-bedrock requires Curl.  This should be mentioned prominently in the README, and if a user tries to use gptel-bedrock with `gptel-use-curl` set to `nil`, a helpful message explaining the requirement should be provided instead of an error.  I don't want a special case checking for `gptel-bedrock` in `gptel--url-get-response` (or as function advice) -- no "leakage".  So this check should be handled in `gptel-bedrock.el` itself somehow.

akssri:
AWS needs sigv4 signing of requests, so Bedrock can't work without curl currently.
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv.html

Option 1, is "theoretically" doable, but I'm not really familiar with it (or how feasible it is in Emacs Lisp); there's an example impl. in CL, so may be not too difficult,
https://github.com/JGalego/cl-bedrock/blob/main/cl-bedrock.lisp#L132

I'd prefer keeping Option 1. as a "feature-request"; and go with Option 2.
(I can add the curl-checks in the Bedrock PR).

karthink:
Let's go with option 2 for now then.  I took a look at the sigv4 spec, and best I can tell it's not feasible to implement it from scratch in Emacs.  But perhaps an elisp wizard can help us out or point us to a library that does it in the future.

akssri:
SG; looks like someone has already ported this to elisp,
https://github.com/atamariya/aws-sign4/blob/master/aws-sign4.el

PR #888, opened 2025-06-01
README: Update link

samihda:
Link for the llama.cpp requirement is currently outdated. This small patch
updates it based on changes in ggml-org/llama.cpp#13249.


karthink:
Thanks for the PR @samihda!

PR #894, opened 2025-06-04
Fix `gptel-auto-scroll` not working when insertion is at end of buffer

### LINKED ISSUE #893: gptel-auto-scroll doesn't work at end of buffer

### Bug Description

`(add-hook 'gptel-post-stream-hook 'gptel-auto-scroll)` doesn't work correctly for me when I'm inserting text at the end of a buffer. I think the issue is that `scroll-up-command` is called from within `save-excursion`, where `point` is potentially already at the end of the buffer, causing it to error out rather than scroll up.

Note also that due to the excursion the comment `Note: This will move the cursor.` is incorrect ([code](https://github.com/karthink/gptel/blob/c6e2339d43dc8c6d7a1ef50e8fe1e5e4b67c74ea/gptel.el#L1011)).

The fix I found is to run `scroll-up-command` outside of the excursion using a delay:
```
(defun gptel-auto-scroll-fixed ()
  "Scroll window if LLM response continues below viewport."
  (when-let* ((win (get-buffer-window (current-buffer) 'visible))
              ((not (pos-visible-in-window-p (point) win)))
              (scroll-error-top-bottom t))
    (condition-case nil
        (with-selected-window win
          (run-at-time 0 nil #'scroll-up-command))
      (error nil))))
```

I'll put up a PR, but wanted to open this bug for discoverability.

### Backend

None

### Steps to Reproduce

1. Enable auto scroll: `(add-hook 'gptel-post-stream-hook 'gptel-auto-scroll)`
2. Move cursor to end of buffer.
3. Invoke `gptel-send` and have it generate text longer than the current buffer.

Expected behavior: as inserted text goes past end of buffer, it scrolls up to follow along with the inserted text.

Actual behavior: it doesn't scroll, and text goes off the end of the page.

### Additional Context

Emacs 27.2

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

JDNdeveloper:
Fixes #893, see issue for details.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1006,15 +1006,13 @@ Later plists in the sequence take precedence over earlier ones."
       (gptel--json-read))))
 
 (defun gptel-auto-scroll ()
-  "Scroll window if LLM response continues below viewport.
-
-Note: This will move the cursor."
+  "Scroll window if LLM response continues below viewport."
   (when-let* ((win (get-buffer-window (current-buffer) 'visible))
               ((not (pos-visible-in-window-p (point) win)))
               (scroll-error-top-bottom t))
     (condition-case nil
         (with-selected-window win
-          (scroll-up-command))
+          (run-at-time 0 nil #'scroll-up-command))
       (error nil))))
 
 (defun gptel-beginning-of-response (&optional _ _ arg)
--- a/gptel.el
+++ b/gptel.el
@@ -1007,7 +1007,7 @@ Later plists in the sequence take precedence over earlier ones."
 
 (defun gptel-auto-scroll ()
   "Scroll window if LLM response continues below viewport.
-	
+
 Note: This will move the cursor."
   (when-let* ((win (get-buffer-window (current-buffer) 'visible))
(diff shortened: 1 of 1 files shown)


karthink:
Using `run-at-time` in the condition case makes the let binding and the condition-case ineffective, since it then runs at top level.

I moved the whole function into the `run-at-time`, but I'm not sure if this works as intended either.  Could you test the latest version and let me know?

JDNdeveloper:
Unfortunately that doesn't work correctly, the check of `(point)` needs to occur within the excursion otherwise it will just remain where the user's cursor is and not detect the inserted text has scrolled off the page. I tried this as well, but it didn't work either:
```
...
  (when-let* ((win (get-buffer-window (current-buffer) 'visible))
              ((not (pos-visible-in-window-p (point) win))))
    (run-at-time
     0 nil
     (lambda ()
       (condition-case nil
           (let ((scroll-error-top-bottom t))
             (when (window-live-p win)
               (with-selected-window win
                 (scroll-up-command))))
         (error nil))))))
```

PR #895, opened 2025-06-05
gptel-bedrock: Add support for IAM role credentials

liangfu:
The changes add support for IAM role-based credentials in gptel-bedrock by implementing:

  1. Modified gptel-bedrock--fetch-aws-profile-credentials function to retrieve AWS credentials from EC2 instances without the need for a profile name.
  2. An expanded credential provider chain that tries:
    - Environment variables first
    - AWS profile settings second
    - Instance credentials third
  3. More informative error messages when authentication fails

  This enables gptel to work in AWS environments using IAM roles for authentication without requiring manual credential configuration, following AWS security best practices.

### DIFF
--- a/gptel-bedrock.el
+++ b/gptel-bedrock.el
@@ -506,9 +506,9 @@ Non-nil CLEAR-CACHE will refresh credentials."
              (or (and (not clear-cache) (cdr cell))
                  (setf (cdr cell)
                        (with-temp-buffer
-		           (unless (zerop (call-process "aws" nil t nil "configure" "export-credentials"
-					                (format "--profile=%s" profile)))
-		             (user-error "Failed to get AWS credentials from profile"))
+		           (unless (zerop (apply #'call-process "aws" nil t nil "configure" "export-credentials"
+                                     (if profile (list (format "--profile=%s" profile)) nil)))
+		             (user-error "Failed to get AWS credentials from profile %s" (if profile (format "from profile: %s" profile) "from IAM role")))
 		         (json-parse-string (buffer-string)))))))
 	 (expiration (if-let (exp (gethash "Expiration" creds-json))
 			     (date-to-time exp))))
@@ -521,22 +521,6 @@ Non-nil CLEAR-CACHE will refresh credentials."
       ((not clear-cache)
        (gptel-bedrock--fetch-aws-profile-credentials profile t))
       (t (user-error "AWS credentials expired for profile: %s" profile)))))
-
-(defun gptel-bedrock--fetch-instance-credentials ()
-  "Fetch AWS credentials using aws configure export-credentials.
-
-Returns a list of access key, secret key, and session token."
-  (with-temp-buffer
-    (condition-case nil
-        (progn
-          (call-process "aws" nil t nil "configure" "export-credentials")
-          (unless (zerop (buffer-size))
-            (let* ((creds (json-parse-string (buffer-string) :object-type 'plist))
-                   (access-key (plist-get creds :AccessKeyId))
-                   (secret-key (plist-get creds :SecretAccessKey))
-                   (session-token (plist-get creds :SessionToken)))
-              (cl-values access-key secret-key session-token))))
-      (error nil))))
 
 (defun gptel-bedrock--get-credentials ()
   "Return the AWS credentials to use for the request.
@@ -558,7 +542,7 @@ Convenient to use with `cl-multiple-value-bind'"
     (cond
       ((and key-id secret-key) (cl-values key-id secret-key token))
       ((and profile) (gptel-bedrock--fetch-aws-profile-credentials profile))
-      ((gptel-bedrock--fetch-instance-credentials))
+      ((gptel-bedrock--fetch-aws-profile-credentials nil))
       (t (user-error "Missing AWS credentials; tried environment variables, AWS profile, and instance credentials")))))
 
 (defvar gptel-bedrock--model-ids


gptel-bedrock.el
```diff
@@ -513,13 +513,34 @@ Non-nil CLEAR-CACHE will refresh credentials."
...
+(defun gptel-bedrock--fetch-instance-credentials ()
+  "Fetch AWS credentials using aws configure export-credentials.
+
+Returns a list of access key, secret key, and session token."
+  (with-temp-buffer
+    (condition-case nil
+        (progn
+          (call-process "aws" nil t nil "configure" "export-credentials")
+          (unless (zerop (buffer-size))
+            (let* ((creds (json-parse-string (buffer-string) :object-type 'plist))
```
karthink, line 525: 1. Prefer buffer parsing to buffer->string->json-parse.
2. Use `gptel--json-read`, which is just `json-parse-buffer` with `:object-type 'plist`.

gptel-bedrock.el
```diff
@@ -528,7 +549,8 @@ Convenient to use with `cl-multiple-value-bind'"
     (cond
       ((and key-id secret-key) (cl-values key-id secret-key token))
       ((and profile) (gptel-bedrock--fetch-aws-profile-credentials profile))
-      (t (user-error "Missing AWS credentials; currently only environment variables are supported")))))
+      ((gptel-bedrock--fetch-instance-credentials))
```
karthink, line 552: This runs on every request.  `call-process` is synchronous and blocks Emacs.  I would like gptel to _never_ block Emacs.

Making this asynchronous in the gptel-request pipeline is more work, and probably best avoided for now.

If these credentials can be cached, would you consider caching them instead?  Then it will block the first time but subsequent requests should be processed instantaneously.

gptel-bedrock.el
```diff
@@ -513,13 +513,34 @@ Non-nil CLEAR-CACHE will refresh credentials."
...
+    (condition-case nil
+        (progn
+          (call-process "aws" nil t nil "configure" "export-credentials")
+          (unless (zerop (buffer-size))
+            (let* ((creds (json-parse-string (buffer-string) :object-type 'plist))
+                   (access-key (plist-get creds :AccessKeyId))
+                   (secret-key (plist-get creds :SecretAccessKey))
+                   (session-token (plist-get creds :SessionToken)))
+              (cl-values access-key secret-key session-token))))
+      (error nil))))
```
karthink, line 530: The error message must be more informative, depending on whether `aws` was (i) available, (ii) called and (iii) failed to return a token.

Currently the request fails silently with no message, since you've condition-cased the `call-process`.

karthink (commented):
Thanks for the PR @liangfu.  I don't understand it but I trust you know what you're doing.

Some comments below.

jhoodsmith (commented):
Hi, 
I'm a long-term Emacs user and a big fan of this repo. Thank to @karthink for all that you have done.
I've followed with interest the recent Bedrock support. as this is used extensively at my work. That said, I'm not sure about this PR for a couple of reasons:

1.  `aws configure export credentials` is generally used once to inject the 3 AWS env variables into a shell prior to interacting with AWS services . It is not called on every interaction.
2. Calling `aws configure export-credentials` on its own, without a prior step, assumes we are using long-lived IAM credentials. If this is the case, then this is not good. AWS advices against this approach, preferring as it does short-lived credentials acquired through  AWS IAM Identity Center (formerly AWS Single Sign-On).

I've been using gptel-bedrock fine for the last few weeks, but before doing so I would make a single call to the following function:
```
(defun my/aws-login (profile)
  "Login to AWS SSO with PROFILE and export credentials to environment."
  (interactive (list (completing-read "AWS Profile: " 
                                     (split-string (shell-command-to-string 
                                                    "aws configure list-profiles") "\n" t))))
  (message "Logging in to AWS SSO with profile %s..." profile)
  (if (zerop (call-process "aws" nil nil nil "sso" "login" "--profile" profile))
      (condition-case err
          (let* ((json-string (shell-command-to-string
                              (format "aws configure export-credentials --profile %s" profile)))
                 (json-data (json-parse-string json-string :object-type 'plist)))
            (setenv "AWS_ACCESS_KEY_ID" (plist-get json-data :AccessKeyId))
            (setenv "AWS_SECRET_ACCESS_KEY" (plist-get json-data :SecretAccessKey))
            (setenv "AWS_SESSION_TOKEN" (plist-get json-data :SessionToken))
            (message "AWS credentials for profile %s exported to environment" profile))
        (error (message "Failed to export credentials: %s" (error-message-string err))))
    (message "AWS SSO login failed for profile %s" profile)))
```
(taken from my [Emacs config](https://github.com/jhoodsmith/emacs.d))
The result is that the short lived credentials get injected into the env variables, which are then picked up by gptel-bedrock.
This works fine for me, supports AWS best practice and makes only a single set of calls to the aws CLI. It also doesn't require bringing additional functionality into gptel (which is better from a separation of concerns point of view - i.e., gpel only cares about using the env variables, not about setting and refreshing them)

karthink:
@akssri, @felipeochoa: Could you take a look at this PR and @jhoodsmith's comment?

felipeochoa:
I agree that `call-process` should not be called inline, but I do like the use of `export-credentials`. If I had known about that command, I probably would have used it instead of relying on environment variables. 

Caching is a good idea, especially for EC2 instances or other places that have to do a network request to get these credentials, but that raises the question of cache invalidation.

@jhoodsmith what happens when your session expires?

Perhaps what we should do is put credentials in a `defvar` and then there could be different "log in" functions that would populate that variable. That would provide flexibility to run it from init.el for people using long-lived credentials, `my/aws-login` for jhoodsmith, `call-process` for liangfu, etc., and even have custom functionality to refresh credentials as needed.

akssri:
@liangfu @jhoodsmith Isn't this just replicating what the AWS_PROFILE feature is doing ? It's the exact same command that called from by ```gptel-bedrock--fetch-aws-profile-credentials``` ? What's missing over there ? The function also does caching and checks for token ttl.

Edit: it looks like while the command is the same, instance credentials are fetched when there is no profile keyword being passed ? If this is the only difference please modify the profile-specific function (with a special profile argument)  instead of duplicating the function.

jhoodsmith:
@akssri - at my work I have to access AWS using AWS SSO, which does the session token management. I thus need that initial `aws sso login` command to generate the token. Yes, the next bit is the same as what is in `gptel-bedrock--fetch-aws-profile-credentials`, but then I would have to run two commands. The single command works better for me

@felipeochoa - alas, when my session expires I have to login again.

akssri:
@jhoodsmith Yes, this is also how I use it - which is why gptel-bedrock--fetch-aws-profile-credentials checks for credential staleness and raises a user error if it has expired (instead of having curl fail semi-silently). 

This function covers your earlier criticism.

@liangfu  Please modify gptel-bedrock--fetch-aws-profile-credential (also see #925)


On ಜುಲೈ 13, 2025 07:25:59 ಅಪರಾಹ್ನ GMT+09:00, James Hood-Smith ***@***.***> wrote:
>jhoodsmith left a comment (karthink/gptel#895)
>
>@akssri - at my work I have to access AWS using AWS SSO, which does the session token management. I thus need that initial `aws sso login` command to get the token. Yes, the next bit is the same as what is in `gptel-bedrock--fetch-aws-profile-credentials`, but then I have would to run two commands. The single command works better for me
>
>@felipeochoa - alas, when my token expires I have to login again. 
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/karthink/gptel/pull/895#issuecomment-3066899430
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
- ಅಕ್ಷಯ
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jhoodsmith:
@akssri - thank you for explaining. I see now how my function can be simpler, and how that second bit that replicates what you've done is not needed if my only interaction with AWS is through gptel and bedrock. To be honest, I was never criticising `gptel-bedrock--fetch-aws-profile-credentials`. I was just questioning `gptel-bedrock--fetch-instance-credentials` (introduced in this PR). That said, I may have been confused about that too! Please ignore my previous comment.

karthink:
@jhoodsmith What's the status of this PR?  I can merge it with a minimal review after @felipeochoa or @akssri give us the go-ahead.

jhoodsmith:
@karthink - this is actually @liangfu's PR, not mine. I personally don't think it is needed. At the very least, we should follow @akssri's advice [above](https://github.com/karthink/gptel/pull/895#issuecomment-3067309936) and make changes to `gptel-bedrock--fetch-aws-profile-credential` instead.

liangfu:
thanks for the feedback. i addressed the review comments via reusing `gptel-bedrock--fetch-aws-profile-credentials`. 

take another look?

akssri:
(Moved to #1053 )

karthink:
Closing in favor of #1053.

PR #906, opened 2025-06-12
ensure that zero-arg tools have valid parameters struct

tazjin:
Some implementations of the OpenAI API (notably, when talking to OpenAI's o3 and up) require tools that have no parameters to still supply a valid object describing them instead of just `null`.

OpenAI's own older implementations (e.g. when talking to 4.1) handle `null` fine, and so do basically all other OpenAI-like endpoints I've tested. All implementations seem to work with the empty but valid object.

I read the docs for json-serialize and couldn't figure out another way to ensure that we get an empty `{}` instead of just a `:null`, hence the empty hash-table.


karthink:
Thanks for the PR.

Is there some reason this doesn't work?

```elisp
(gptel--json-encode '(:parameters (:type "object" :properties nil)))
;; => {"parameters":{"type":"object","properties":{}}}
```

> All implementations seem to work with the empty but valid object.

It was changed recently to `:parameters :null` because of compatibility issues (#818), so we need to be sure of this.  What providers have you tried besides OpenAI?

tazjin:
> Is there some reason this doesn't work?

Huh, that does work. For some reason I was sure that `nil` would serialise as an empty list, but in testing this does yield the correct `{}` object.

>  What providers have you tried besides OpenAI?

OpenAI (various models), company-internal proxy to various models, Anthropic.

I'm going to be running this patch for a while and try some more models. For now I haven't run into any issues. It could at least be a config option if you want to be overly cautious, but I don't see much reason why a "more valid" object would cause failures than a "less valid" one.

By the way, thanks for gptel - it is by far the best interface to these things that exists so far. I mean in general, not just in Emacs.

karthink:
> OpenAI (various models), company-internal proxy to various models, Anthropic.

The non-uniformity of behavior is across "OpenAI-compatible" APIs, like OpenAI itself, Openrouter, Groq, Stability, Perplexity, Together.ai, Anyscale, Deepseek, Cerebras, xAI, Bedrock etc.

OpenAI does not publish a formal spec for its API, so every provider is free to interpret unspecified behavior however they'd like.  These are the ones that need testing with this patch applied.

> I'm going to be running this patch for a while and try some more models. For now I haven't run into any issues. It could at least be a config option if you want to be overly cautious,

Making this configurable is a worse option as I see it.  I think it's gptel's job to sort details like this out so the user doesn't have to worry about it.

> but I don't see much reason why a "more valid" object would cause failures than a "less valid" one.

True.  Maybe we should just apply it and see if anyone using one of the other APIs above reports a bug.

> By the way, thanks for gptel - it is by far the best interface to these things that exists so far. I mean in general, not just in Emacs.
 
Cheers!

tazjin:
> True. Maybe we should just apply it and see if anyone using one of the other APIs above reports a bug.

I've tried it with our local DeepSeek installation, but it is experiencing general issues with tool calls atm so the results are inconclusive. It does seem to accept the tool list fine though. The other APIs in the list I personally don't use.

tazjin:
@karthink Tested:

- OpenAI
- Anthropic
- OpenRouter
- aitunnel.ru
- DeepSeek (public)
- DeepSeek (local setup using `sglang`)
- Gemini (albeit through a proxy, but one that should just pass through requests transparently)

karthink:
Thanks for testing against many APIs!  That gives me enough confidence to merge this. 👍

tazjin:
Thanks! Note that within APIs there might also be differences (e.g. OpenAI specifically seems to have slightly different parsers for different models), so lets keep an eye out for problems anyways. If you see anything related, feel free to tag me, too, I'll be happy to take a look.

PR #910, opened 2025-06-15
gptel: Add custom 'gptel-auth-user-name' for different API keys.

mvaled:
I have two API keys, one personal and one for work.  This let's me to easily switch them.


karthink:
`gptel-api-key-from-auth-source` is meant as an example only, so I don't want to add more customization that depends on it.

I suggest writing your own function instead, or getting API keys from the environment with `getenv`.

PR #911, opened 2025-06-16
Only MCP server in the connected state are considered active.

lizqwerscott:
Modified the logic for determining the active state. Now, a server is considered active only when the MCP server is in the 'connected' state. Previously, if the server was started and then stopped, it would remain in the mcp-server-connections list with a 'stop' status, incorrectly allowing tool retrieval attempts (which would fail). With this change, if the server is stopped and a tool is added, the server will automatically restart.


karthink:
Thanks for the update @lizqwerscott!

PR #917, opened 2025-06-20
gptel-rewrite: reject changes before ediff

nano-o:
The ediff feature of gptel-rewrite is somewhat broken: 
* The overlay showing the LLM response hides the ediff information during the ediff session, so it's hardly usable
* Even if gptel would hide the overlay for the duration of the ediff session, it's not clear to me what should happen when the ediff session ends and changes have been made. Should the LLM-response overlay be restored, and with what contents?

I think it makes more sense to just discard the LLM response when entering ediff (then it's just a normal ediff session with no surprises). This is what this PR does. It might also make sense to make the LLM-response buffer read-only (not done in this PR).

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -298,29 +298,20 @@ BUF is the buffer to modify, defaults to the overlay buffer."
       (display-buffer diff-buf))))
 
 (defun gptel--rewrite-ediff (&optional ovs)
-  "Ediff pending LLM responses in OVS or at point."
-  (interactive (list (gptel--rewrite-overlay-at)))
-  (when-let* ((ov-buf (overlay-buffer (or (car-safe ovs) ovs)))
-              ((buffer-live-p ov-buf)))
-    (letrec ((newbuf (gptel--rewrite-prepare-buffer ovs))
-             (cwc (current-window-configuration))
-             (hideshow
-              (lambda (&optional restore)
-                (dolist (ov (ensure-list ovs))
-                  (when-let* ((overlay-buffer ov))
-                    (let ((disp (overlay-get ov 'display))
-                          (stored (overlay-get ov 'gptel--ediff)))
-                      (overlay-put ov 'display (and restore stored))
-                      (overlay-put ov 'gptel--ediff (unless restore disp)))))))
-             (gptel--ediff-restore
-              (lambda ()
-                (when (window-configuration-p cwc)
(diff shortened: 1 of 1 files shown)


karthink:
> * The overlay showing the LLM response hides the ediff information during the ediff session, so it's hardly usable

This is not the case, or at least not the intended behavior.  You should see the original text in one buffer and the LLM-response text in the other.

Can you provide a screenshot showing what you see during ediff?

> * Even if gptel would hide the overlay for the duration of the ediff session, it's not clear to me what should happen when the ediff session ends and changes have been made. Should the LLM-response overlay be restored, and with what contents?

It is currently restored because the user might want to make more changes.  It is not apparent that quitting ediff with `q` means that the user is done with the LLM response -- what if they just wanted to quit Ediff?

nano-o:
Here's a screenshot below. On the left side you can see that the overlay hides the ediff information (it shows the original text).

Regarding what should happen after the ediff session ends: my workflow is that I incorporate LLM changes during the ediff session. Then I don't want to redisplay the overlay with the orignal LLM changes, as I already acted on them.

image

nano-o:
Here's a different suggestion. I don't really know what I'm doing though. Is setting the face to nil the right thing to do?

karthink:
> * The overlay showing the LLM response hides the ediff information during the ediff session, so it's hardly usable

> Here's a screenshot below. On the left side you can see that the overlay hides the ediff information (it shows the original text).

I don't understand -- the buffer on the left shows the original text, and the buffer on the right shows the rewritten text.  This is exactly how it should work, isn't it?

nano-o:
Okay, my bad if this is the intended behavior. My issue with it is that the yellow face of the overlay on the left-hand-side hides color-coded ediff information. Instead I was expecting to see this something like this:
image

karthink:
> Instead I was expecting to see this something like this:

Your expectation makes sense -- that's better than the current behavior.

karthink:
I've changed it so the yellow overlay face is removed when Ediffing.  Please update and test.

> my workflow is that I incorporate LLM changes during the ediff session. Then I don't want to redisplay the overlay with the orignal LLM changes, as I already acted on them.

I understand.  But what if you Ediff and decide you don't want to make changes yet?  When you quit ediff, the rewrite information is lost.

nano-o:
Thanks!

PR #918, opened 2025-06-20
Bump curl version requirement for AWS Bedrock

leezu:
While Curl 8.5.0 suppots --aws-sigv4 parameter, it's implementation is buggy and versions prior to 8.9.0 trigger a "The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method." error when interacting with Bedrock.


karthink:
Thanks for the update!

PR #919, opened 2025-06-20
Update gptel-make-bedrock example in README

leezu:
Update to Sonnet 4 and better explain model-region parameter.


karthink:
Thanks for the update @leezu!

PR #926, opened 2025-06-24
Add gemini-2.5 pro and flash productin models

stribb:
Also introduce 2.5-flash-lite.


karthink:
Thanks for the PR.

Could you shorten the `:description` fields to 70 chars or less?  It needs to fit into various places in gptel, such as annotations in the minibuffer when choosing models.

You can look at the `:description` of the existing Gemini models in `gptel--gemini-models` to see what I mean.

karthink:
Tweaked descriptions and merged.  Thanks for the PR @stribb!

PR #936, opened 2025-06-27
gptel-org: Update gptel-org-set-topic docstring for clarity

rgkirch:
I thought it was saying the "current" headline but it actually finds whatever parent or grandparent headline that has the topic set.


karthink:
Thanks for the PR @rgkirch!

rgkirch:
You are welcome!

Issue #932, opened 2025-06-26
Better and more fine grained conflicts for rewrite merge


meain:
We could probably improve what we put into the buffer for `C-c RET` ie `merge`. Instead just putting the text as is, we could put the output of passing it through git with a command like below:

```emacs-lisp
(format "git merge-file --no-diff3 -L Original -L Empty -L 'LLM Response' -p %s %s %s" original-temp-file empty-temp-file new-temp-file)
```

This is useful if the update is only for a small section or multiple small sections. I use something similar in [yap for rewrite](https://github.com/meain/yap/blob/baf9b15453b2dcc53ccdc1e4478380d19eedd9f2/yap.el#L257). I'm happy to submit a PR if you are interested.

For example, if I were to call `gptel-rewrite` with "update the docstring", the previous version would have produced the following:

```
<<<<<<< original
(defun gptel--rewrite-ediff (&optional ovs)
  "Ediff pending LLM responses in OVS or at point."
  (interactive (list (gptel--rewrite-overlay-at)))
  (when-let* ((ov-buf (overlay-buffer (or (car-safe ovs) ovs)))
              ((buffer-live-p ov-buf)))
    (letrec ((newbuf (gptel--rewrite-prepare-buffer ovs))
             (cwc (current-window-configuration))
             (hideshow
              (lambda (&optional restore)
                (dolist (ov (ensure-list ovs))
                  (when-let* ((overlay-buffer ov))
                    (let ((disp (overlay-get ov 'display))
                          (stored (overlay-get ov 'gptel--ediff)))
                      (overlay-put ov 'face (and restore 'gptel-rewrite-highlight-face))
                      (overlay-put ov 'display (and restore stored))
                      (overlay-put ov 'gptel--ediff (unless restore disp)))))))
             (gptel--ediff-restore
              (lambda ()
                (when (window-configuration-p cwc)
                  (set-window-configuration cwc))
                (funcall hideshow 'restore)
                (remove-hook 'ediff-quit-hook gptel--ediff-restore))))
      (funcall hideshow)
      (add-hook 'ediff-quit-hook gptel--ediff-restore)
      (let ((ediff-window-setup-function #'ediff-setup-windows-plain)
            (ediff-split-window-function #'split-window-horizontally))
        (ediff-buffers ov-buf newbuf)))))
=======
(defun gptel--rewrite-ediff (&optional ovs)
  "Compare pending LLM responses with original text using Ediff."
[...]

PR #939, opened 2025-06-29
Use git to produce better merge conflicts for rewrite

meain:
Previously, we were producing simple conflict markers by just putting original and generated code into conflict markers. In this implementation, if we have git available, we pass the original and new version through git to produce a more fine grained set of conflicts.

Fixes https://github.com/karthink/gptel/issues/932


karthink:
Thanks @meain.  I rewrote the two functions to act on buffer regions instead of passing strings.  (The previous, "functional" approach is garbage heavy.)

Can you check if they still work correctly?

meain:
> The previous, "functional" approach is garbage heavy

Aah, that is something I had not considered.

> Can you check if they still work correctly?

Yup, looks good to me :D

karthink:
Merged, thanks for the PR @meain!

> Aah, that is something I had not considered.

To be fair, it's not much of an issue since the text being rewritten will usually consist of a few paragraphs or functions at most, which isn't much garbage even when it's duplicated twice as a string and again across three separate buffers (as happens when using `shell-command-to-string` and passing strings around.)

But I _never_ touch working gptel code to improve performance -- plenty of other things to do!  So I thought it might be worth getting it right the first time here.

Issue #905, opened 2025-06-12
[Feature Request] Allow values of presets to be function


ahmed-shariff:
Right now, the values in presets are set statically, except for tools, which has the special form `:append`. This would allow dynamically computing the values of presets when applying the preset.

I have a simple modified version of gptel--preset-apply in a branch (https://github.com/karthink/gptel/commit/bf91ff5ac8df10495026e5e9b918962a6083f5a8):
```diff
diff --git a/gptel.el b/gptel.el
index dd2dbe3..6524ff5 100644
--- a/gptel.el
+++ b/gptel.el
@@ -3614,7 +3614,8 @@ example) apply the preset buffer-locally."
                             (car preset) val))
             (funcall setter sym val))))
        (:backend
-        (setq val (cl-etypecase val
+        (setq val (if (functionp val) (funcall val) val)
+              val (cl-etypecase val
                     (gptel-backend val)
                     (string (gptel-get-backend val))))
         (unless val
@@ -3622,6 +3623,7 @@ example) apply the preset buffer-locally."
                       (car preset) val))
         (funcall setter 'gptel-backend val))
        (:tools                          ;TEMP Confirm this `:append' convention
+        (setq val (if (functionp val) (funcall val) val))
         (let* ((append (when (eq (car-safe val) :append) (setq val (cdr val)) t))
                (tools
                 (flatten-list
@@ -3641,7 +3643,7 @@ example) apply the preset buffer-locally."
                           (intern-soft
                            (concat "gptel--" (substring (symbol-name key) 1)))))
              (guard (and sym (boundp sym))))
-        (funcall setter sym val))
+        (funcall setter sym (if (functionp val) (funcall val) val)))
        (_ (display-warning
            '(gptel presets)
            (format "gptel preset \"%s\": setting for %s not found, ignoring."
```

Posting as an issue to see if this is of interest before I make a PR.
The main reason I started experimenting with this is in relation to `gptel-prompt-transform-functions`.

karthink:
@ahmed-shariff do you have an example of how this helps with setting `gptel-prompt-response-functions`?

ahmed-shariff:
The use case is similar to the `tool`s `:append`. The problem, however, is that the order is more important with transformation functions. So, without the function value option, I have separate presets for each combination of these transformation functions:

```emacs-lisp
(gptel-make-preset default
  ...
  :prompt-transform-functions `(gptel--transform-apply-preset gptel--transform-add-context))

(gptel-make-preset transform1
[...]

PR #940, opened 2025-06-29
Allow dynamically augmenting/overriding preset parameters during apply

kmontag:
If a preset's `:pre` callback returns a plist, use it to augment/override the preset's normal specification.

This is a proposed implementation for "functional" presets, as discussed in https://github.com/karthink/gptel/issues/905 (though it's a bit different than the solution discussed there).

I believe it's also a more generic solution to the issue that's currently being solved by the special handling of `:append` in presets that specify `:tools`. If a preset wants to specify `:prompt-transform-functions`, for example, this change allows appending or inserting into the current global `gptel-prompt-transform-functions`, rather than overwriting it.

Naturally this would somewhat change the contract of the `:pre` function, and if there's anyone out there who's idly returning a plist, this could introduce unexpected side effects. I don't expect that's a particularly common case. However, another option would be to use a more explicitly-named field like `:override` for this function, instead of `:pre`.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -3525,7 +3525,9 @@ PARENTS is a preset name (or list of preset names) to apply before this
 one.
 
 PRE and POST are functions to run before and after the preset is
-applied.  They take no arguments.
+applied.  They take no arguments.  If PRE returns a plist, its contents
+will be merged into the existing specification while appying the preset.
+Use this to adjust the preset based on the current environment.
 
 BACKEND is the gptel-backend to set, or its name (like \"ChatGPT\").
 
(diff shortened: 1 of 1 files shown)


kmontag:
By the way, I realized it's also already possible to do something like this directly in the `:pre` or `:post` function:

```elisp
(gptel-make-preset
 'remove-blacklisted-tool
 ;; ...
 :pre
 (lambda ()
   (set
    'gptel-tools
    (seq-remove (lambda (tool) (equal (gptel-tool-name tool) "blacklisted_tool_name")) gptel-tools))))
```

However, this would assume the use of `set` to apply the preset settings, which isn't always the case.

kmontag:
Update: only allow modifying keys which are already present in the spec (i.e. you can't add new ones). Otherwise, `gptel--preset-syms` wouldn't necessarily pick up all the symbols that are being set.

kmontag:
I believe this is now obsolete after the solution to https://github.com/karthink/gptel/issues/905, closing.

PR #952, opened 2025-07-03
gptel: Fix warnings when specifying `:pre` or `:post` in a preset

kmontag:
This fixes stray warnings like `Warning (gptel): gptel preset ":description": setting for :pre not found, ignoring.`.


karthink:
@kmontag Thank you for the PR!

PR #956, opened 2025-07-08
`gptel--insert-file-string` should tell GPT the full file path.

eval-exec:
I think it would be better to let GPT know the whole file path. It's very helpful. GPT will know the project's name and what your file's module.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -991,7 +991,7 @@ Later plists in the sequence take precedence over earlier ones."
 
 (defun gptel--insert-file-string (path)
   "Insert at point the contents of the file at PATH as context."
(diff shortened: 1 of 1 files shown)


meain:
It might be useful to strip out the project prefix if it is a project.el or projectile.el recognized project.

karthink:
This is a duplicate of #863, where @munen is looking at improving the context annotation by doing this + some other things.

karthink:
@meain See https://github.com/karthink/gptel/pull/863#issuecomment-2949319777

Closing as this is obsoleted by bed2402c5442fec21cc336a76821a5c89c1b254b and #863.

PR #958, opened 2025-07-09
Fix invalid use of transient-scope with arguments

xanalogica:
Hi!  Thank you for this excellent package — I've been integrating gptel into my Emacs workflow and ran into an issue when calling transient-based commands with the newest stable version of Emacs.

Some recent versions of Emacs (e.g. 30.1.50+) now enforce the arity of transient-scope.  It's a no-argument function, but in a few places in the code, it's currently being called like:

    _(transient-scope 'gptel-tools)_

This raises an error:

    _(wrong-number-of-arguments #<subr transient-scope> 1)_

This PR replaces all such calls with the correct zero-argument form:

    _(transient-scope)_
    
It was only a couple of occurrences:
    - gptel-integrations.el
    - gptel-transient.el

### DIFF
--- a/gptel-integrations.el
+++ b/gptel-integrations.el
@@ -220,7 +220,7 @@ from all connected servers if it is nil."
          nil (lambda () (when-let* ((transient--prefix)
                                ((eq (oref transient--prefix command)
                                     'gptel-tools)))
-                     (let ((state (transient-scope 'gptel-tools)))
+                     (let ((state (transient-scope)))
                        (plist-put state :tools
                                   (delete-dups
                                    (nconc (mapcar (lambda (tool)
@@ -250,7 +250,7 @@ from all connected servers if it is nil."
     (call-interactively #'gptel-mcp-disconnect)
     ;; gptel-tools stores its state in its scope slot.  Retain the scope but
     ;; remove tools from it that no longer exist, then set up gptel-tools
-    (cl-loop with state = (transient-scope 'gptel-tools)
+    (cl-loop with state = (transient-scope)
              with tools = (plist-get state :tools)
              for tool-spec in tools
              if (map-nested-elt gptel--known-tools tool-spec)
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -994,7 +994,7 @@ only (\"oneshot\")."
      (lambda (tools)
        ;; We don't care about the transient args of this prefix at all, since
        ;; the state is managed entirely through its transient-scope:
-       (interactive (list (plist-get (transient-scope 'gptel-tools) :tools)))
+       (interactive (list (plist-get (transient-scope) :tools)))
        (gptel--set-with-scope
         'gptel-tools
         (mapcar (lambda (category-and-name)


karthink:
`transient-scope` is not a zero-arity function for me in Transient versions 0.7.4 (minimum required by gptel) through 0.8.8 (what I'm using):

``` text
transient-scope is a native-comp-function in ‘transient.el’.

(transient-scope &optional PREFIXES CLASSES)

Return the scope of the active or current transient prefix command.

If optional PREFIXES and CLASSES are both nil, return the scope of
the prefix currently being setup, making this variation useful, e.g.,
in ‘:if*’ predicates.  If no prefix is being setup, but the current
command was invoked from some prefix, then return the scope of that.

If PREFIXES is non-nil, it must be a prefix command or a list of such
commands.  If CLASSES is non-nil, it must be a prefix class or a list
of such classes.  When this function is called from the body or the
‘interactive’ form of a suffix command, PREFIXES and/or CLASSES should
be non-nil.  If either is non-nil, try the following in order:
```

Calling it with an argument is necessary to avoid some edge-cases involving async operations.

Can you identify where this change was made?

munen:
I run Emacs via Guix and have just upgraded from Emacs 30.0.92-0.881d593 to 30.1 and can confirm that the signature of `transient-scope` has changed. Without this PR, I was not able to select any tool in gptel.

These are the docs for `transient-scope` for me:

```
transient-scope is a native-comp-function in ‘transient.el’.

(transient-scope)

Return the value of the ‘scope’ slot of the current prefix.
```

Interestingly, I have `transient` installed through a transient dependency of both gptel and magit (both of which show the same version):

```
Package gptel is installed.

     Status: Installed in ‘gptel-20250706.357/’ (unsigned). Delete
    Version: 20250706.357
     Commit: e9d92423d138e6f217a494ac4b28fe68d349233a
    Summary: Interact with ChatGPT or other LLMs.
   Requires: emacs-27.1, transient-0.7.4, compat-29.1.4.1
    Website: https://github.com/karthink/gptel
   Keywords: convenience tools 
     Author: Karthik Chikmagalur <karthik.chikmagalur@gmail.com>
Other versions: 20250605.1803 (installed), 20250604.140 (installed), 0.9.8 (installed), 20250713.2034 (melpa).
```

```
Package transient is installed.

     Status: Installed in ‘transient-20241023.1537/’,
             shadowing a built-in package (unsigned).
    Version: 20241023.1537
     Commit: cb11650a60728d214b2a4d7bb7535add2c4c5d4c
    Summary: Transient commands.
   Requires: emacs-26.1, compat-30.0.0.0, seq-2.24
Required by: magit-20241023.1526, gptel-20250706.357, git-commit-20240520.1135,
               forge-20241201.700, cider-20241031.717, aidermacs-20250508.34
    Website: https://github.com/magit/transient
   Keywords: extensions 
     Author: Jonas Bernoulli <emacs.transient@jonas.bernoulli.dev>
Other versions: 20240525.1118 (installed), 20231026.1048 (installed), 20250701.1223 (melpa), 0.7.2.2 (builtin).
```

And as you've mentioned 0.7.4 is the minimum requirement of gptel.

The latest version of transient also does not have an empty arg list: https://github.com/magit/transient/blob/main/lisp/transient.el#L4163

I'll try upgrading transient and will update this ticket here.

munen:
I see the problem: For whatever reason, Emacs 30.1 does *not* use the installed transient `transient` dependency from gptel, but the one that came from Emacs which is only version 0.7.2.2.

I'm trying to figure out why Emacs uses that one, instead of the new version installed through package management.

munen:
TL;DR:

Emacs 30.1 seems to have introduced different precedence of packages. Therefore it loaded transient 0.7.2.2 which actually has a function arity of 0. This is how I ensured that the version from Guix takes precedence over the one that comes pre-installed with Emacs: https://github.com/munen/emacs.d/commit/29420de597d8489993f7da8310cdf2f88df34c44

Now, I run transient 0.9.1 and can select tools, again.

Imo this PR could be valuable to ensure that old versions of transient also work. But it would require a check against the `transient-version` variable to ensure that it keeps on working with versions that are not ancient(;

karthink:
> Imo this PR could be valuable to ensure that old versions of transient also
> work. But it would require a check against the `transient-version` variable to
> ensure that it keeps on working with versions that are not ancient(;

I'd like to avoid this kludge.

Assuming you're using package.el to manage packages, what is the value of `package-install-upgrade-built-in`?

karthink:
@munen I'm trying to see if there is anything actionable in this PR, since there is no `transient-scope` arity bug with transient 0.7.4 or higher, which gptel requires.

Could you let me know about the `package-install-upgrade-built-in` question above?

munen:
Thanks for the ping, again :pray: 

```
package-install-upgrade-built-in is a variable defined in ‘package.el’.

Its value is nil

Non-nil means that built-in packages can be upgraded via a package archive.
If disabled, then ‘package-install’ will not suggest to replace a
built-in package with a (possibly newer) version from a package archive.

  This variable was introduced, or its default value was changed, in
  version 29.1 of Emacs.
  You can customize this variable.
```

I haven't configured or customized the variable.

I do agree that this PR would be a kludge. In any case, it looks like it's a bug somewhere outside of gptel, because you clearly specified a higher/appropriate version requirement! Maybe it's a bug in Emacs when installed via Guix. Whatever it is, it probably should not affect/complicate code in gptel. I agree this PR can be closed. It's still here for future reference when someone else has this issue.

xanalogica:
We can drop this PR I think.  My problem was I expected Emacs 30.1.x to be using a recent version of transient.el but instead it's very old.  Problem was though that it was hard to replace the version, for me at least here, because it gets loaded VERY early.  After many attempts to override, reorder etc. I finally had to just load transient.el 0.8.8 in init.el and it worked.  But it took me a couple of days to get it to work.

(defun my/checkout-transient-tag (tag)
  "Force checkout of a specific TAG in the transient repo."
  (let* ((default-directory (expand-file-name "straight/repos/transient/" user-emacs-directory)))
    (when (file-directory-p default-directory)
      (message "[transient-fix] Forcing checkout of tag: %s" tag)
      (shell-command (format "git fetch --tags && git checkout tags/%s" tag)))))
      
(my/checkout-transient-tag "v0.8.8")

(straight-register-package
 '(transient :type git
             :host github
             :repo "magit/transient"
             :tag "v0.8.8"
             :files ("*" (:exclude ".git"))))

(straight-rebuild-package "transient")

(straight-use-package 'transient)

(defun check-transient-version (expected-version)
  "Verify that the loaded `transient` package matches EXPECTED-VERSION."
  (require 'transient)
  (let ((actual-version (if (boundp 'transient-version)
                            (symbol-value 'transient-version)
                          "(no version found)")))
    (if (string= actual-version expected-version)
        (message "[transient-check] ✅ Version %s is correctly loaded." actual-version)
      (warn "[transient-check] ⛔ Version mismatch! Expected %s but got %s"
            expected-version actual-version))))

(check-transient-version "0.8.8")

PR #960, opened 2025-07-09
Fix log-level selection menu

nano-o:
If using a list of symbols then the menu displays information about the info and debug functions, which have nothing to do with gptel's log level.


karthink:
@nano-o Thanks for the PR!

Issue #946, opened 2025-06-30
Confused about the DX of rejecting the tool use call


Martinsos:
I love tool use in gptel, it makes integration with emacs really shine!

I wrote some of my tools, configured them to be included or not, confirmed or not, that works great, but one thing I can't figure out is what happens when I reject a tool call. Let's say that AI suggest to execute an `emacs_eval` tool call that will add something to the `display_buffer_alist`. I don't want to run it, because I just need the code, I will add it to init.el myself, so I do C-c C-k (tool reject). What happens now is that I get a message that I need to reactivate tools in the gptel-menu, and chat kind of just "stop" -> the tool call overlay dissapears, and there is no next step in the chat, no "*** You:".

I am not sure how to think about this state, what I am supposed to do at that moment? Why did the tool calling get canceled, why not just that one tool call? I probably don't know enough about the underlying mechanism of tool calling hm, but I wonder what is the intended workflow in gptel at this moment.

karthink:
> Let's say that AI suggest to execute an `emacs_eval` tool call that will add something to the `display_buffer_alist`. I don't want to run it, because I just need the code, I will add it to init.el myself, so I do C-c C-k (tool reject). What happens now is that I get a message that I need to reactivate tools in the gptel-menu, and chat kind of just "stop" -> the tool call overlay dissapears, and there is no next step in the chat, no "*** You:".
>
> I am not sure how to think about this state, what I am supposed to do at that moment? 

What is your mental model of a rejected tool call?  What should happen next?

> Why did the tool calling get canceled, why not just that one tool call?  I probably don't know enough about the underlying mechanism of tool calling hm, but I wonder what is the intended workflow in gptel at this moment.

You cannot continue a request with an LLM if it asks for tool results but you don't provide (all of) them.  No further interaction in that loop is possible.

Of course, you can always start a new request by editing the buffer however you want (or not editing it at all) and calling `gptel-send` again.  This is the intended workflow.

tazjin:
> You cannot continue a request with an LLM if it asks for tool results but you don't provide (all of) them. No further interaction in that loop is possible.

Technically, you can! You can return a result string saying "The user rejected the tool call"; you could even make it so that the user can optionally specify a reason, so the LLM gets information on why that tool call was rejected.

I've taken the approach of feeding errors and rejections verbatim to the LLMs, and it often leads to them modifying their approach and doing something closer to what you want, but YMMV.

karthink:
[...]

PR #968, opened 2025-07-13
gptel: allow functions for tool confirmation

tobjaw:
Hey there :wave:
This PR gives the `:confirm` argument tools a bit more flexibility.
Now it can also take a _function_, letting you conditionally confirm tool calls based on their arguments. I find it handy to reduce the amount of prompts I have to confirm while still keeping things safe.


karthink:
@tobjaw Thanks for the PR.

I don't understand the intended workflow when the confirm function returns `nil`.  Since it is not "auto-confirmed", a confirmation prompt is shown as usual: now what?

In your example of `my-filepath-within-current-directory-p`, you can't change the arguments or the return value of the tool, or continue the request.  It has to be rejected manually anyway.

I think a better approach is to build the safeguard into the tool, as (for example) the official `filesystem` MCP server does.  The _tool_ checks if the arguments are sane.  If they are not it can return a message tailored to the circumstance.  Something like:

```
"/path/to/foo": file access denied.  Only files inside /bar/... are accessible.
```

The other advantage of this approach is that this is the result of evaluating the tool, so it is fed back to the LLM in the same tool-use loop and the LLM can correct itself.  I don't think this is possible if you add the check as a `:confirm` function outside the tool instead.

Maybe I'm missing something, let me know your thoughts.

tobjaw:
Thanks for getting back to me @karthink.

> I don't understand the intended workflow when the confirm function returns `nil`. Since it is not "auto-confirmed", a confirmation prompt is shown as usual: now what?
> 
> In your example of `my-filepath-within-current-directory-p`, you can't change the arguments or the return value of the tool, or continue the request. It has to be rejected manually anyway.

In my use case, I don't want to be asked about file access happening inside a projects dir (e.g. let `:confirm` return `nil` to allow all access), but I also don't want to outright forbid file access outside it either, e.g. because I'm developing an API client and a tool might need to read the servers `openapi.yaml` located in another projects directory (let `:confirm` return `t` and prompt me if it's ok to proceed).
Basically I want the convenience of having few prompts to confirm while still maintaining control over potentially sensitive files.

> I think a better approach is to build the safeguard into the tool, as (for example) the official filesystem MCP server does. The tool checks if the arguments are sane. If they are not it can return a message tailored to the circumstance. 

That's a great approach for something that does not require user input and I will totally ~steal~ use that for something like explicitly denying certain shell commands. But for the use case mentioned above that requires user confirmation I'd rather use the existing prompt mechanism instead of asking for user input during tool invocation.

> The other advantage of this approach is that this is the result of evaluating the tool, so it is fed back to the LLM in the same tool-use loop and the LLM can correct itself. I don't think this is possible if you add the check as a :confirm function outside the tool instead.

I haven't considered that, that is pretty cool 👍🏻

meain:
I have another interesting use-case for this. We can use a function in `:confirm` for a function which lets users run arbitrary shell commands, but the lambda linked to `:confirm` can look at the shell command and return if it is safe(maybe using another llm) and if safe, just run it and if not, ask user for confirmation.

karthink:
> I have another interesting use-case for this. We can use a function in `:confirm` for a function which lets users run arbitrary shell commands, but the lambda linked to `:confirm` can look at the shell command and return if it is safe(maybe using another llm) and if safe, just run it and if not, ask user for confirmation.

Why can't the tool function run this check instead of spreading the logic across two fields, `(gptel-tool-function tool)` and `(gptel-tool-confirm tool)`?

karthink:
> In my use case, I don't want to be asked about file access happening inside a projects dir (e.g. let `:confirm` return `nil` to allow all access), but I also don't want to outright forbid file access outside it either, e.g. because I'm developing an API client and a tool might need to read the servers `openapi.yaml` located in another projects directory (let `:confirm` return `t` and prompt me if it's ok to proceed).  Basically I want the convenience of having few prompts to confirm while still maintaining control over potentially sensitive files.

I understand the use case.  What I'd like to know is: what is the advantage of prompting the user via the `:confirm` behavior vs prompting the user via the tool function itself?  Because the advantage of the latter over the former is that you can indicate the refusal (and refusal reason) to the LLM and get it to change its behavior.

Further, suppose you deny the request via the `:confirm` prompt.  What now?  The tool use loop is suspended and can only be resumed, not edited.  But you don't want to resume it because it's unsafe or undesirable.  What do you do next?

The proposed change is small and not difficult to add.  But so far the feature it adds looks like an antipattern to me.  I'm trying to understand if it has any use cases that are not covered by moving the `:confirm` logic into the tool function.

tobjaw:
> what is the advantage of prompting the user via the `:confirm` behavior [...]?

For me mainly re-use of the existing rich prompt, with possibilities to inspect the request etc.
Maybe there's an easy way to replicate it inside the tool itself, I don't know.

> Further, suppose you deny the request via the :confirm prompt. What now? The tool use loop is suspended and can only be resumed, not edited. But you don't want to resume it because it's unsafe or undesirable. What do you do next?

I typically edit the buffer and make some changes in the context so the LLM will not repeat the previously denied request which works well for me. Which approach do you like?

> I'm trying to understand if it has any use cases that are not covered by moving the :confirm logic into the tool function.

Why have a `:confirm` field in the first place if it isn't the place to ask for the users confirmation / configure that bevaviour?
Couldn't you just as easily move it into the tool function as well? Maybe I'm missing something about the design here.

meain:
> Why can't the tool function run this check instead of spreading the logic across two fields, `(gptel-tool-function tool)` and `(gptel-tool-confirm tool)`?

If permission logic is handled only within the tool function, it can either allow or deny access, but cannot prompt the user for confirmation. By using a `tool-confirm` mechanism, we gain flexibility:

For example, with a file access tool, we can auto-approve access for files inside the current project directory, but trigger a user confirmation for files outside it. This enables conditional user prompts based on context, instead of a strict allow/deny response from the tool itself.

Example:
- Accessing `proj1/info.txt`: auto-approved (inside current project)
- Accessing `proj2/info.txt`: user confirmation requested (outside current project)

---

In short, it is not that the tool itself cannot check it, but by the time the request reaches the tool, there is no way to go back and ask for confirmation in gptel UI.

karthink:
> I typically edit the buffer and make some changes in the context so the LLM will not repeat the previously denied request which works well for me. Which approach do you like?

This is the only approach supported right now.  In #946 I tried an alternate approach where instead of canceling the request you can send a refusal to evaluate as the tool result.  This should directly communicate to the LLM that the tool call was not accepted.

Unfortunately when I test this the LLM (Claude) keeps trying to make similar tool calls endlessly.

Irrespective of whether we merge this PR, the "you refused the tool call, what now?" question does not have a clear answer.

> For me mainly re-use of the existing rich prompt, with possibilities to inspect
> the request etc.  Maybe there's an easy way to replicate it inside the tool
> itself, I don't know.

Yes, the overlay prompt cannot be accessed from inside the tool itself without a lot of kludge.  That might be a good reason to add this feature, but there might also be a better design possible that solves the "what now?" question.  I'm not sure what that is yet.

> Why have a `:confirm` field in the first place if it isn't the place to ask for the users confirmation / configure that behavior?  Couldn't you just as easily move it into the tool function as well? Maybe I'm missing something about the design here.

The problem is not so much with reusing the UI for conditional confirmation, but what happens next, as explained above.

karthink:
Thanks for the PR @tobjaw, I've merged this with minor changes.

I spent a while thinking about this change.  I'm still not convinced it's the best design so I reserve the right to break this behavior in the future.  But in practice it's unlikely I'll need to do that, as the change is small.

karthink:
An update: gptel now provides the hooks `gptel-pre-tool-call-functions` and `gptel-post-tool-call-functions`.  You can also edit tool calls interactively when using `gptel-send`.

Between the three options, you can inspect and act on tool calls and tool results, modify arguments, short-circuit the call and provide results yourself, modify the result of the tool call, block tool calls and supply the LLM with a reason, or stop the request entirely.  See the documentation of the hooks.  (If you want more information see commit messages from a080e68 to af0d39fa.)

PR #976, opened 2025-07-18
Return to transient menu after prompt selection

nano-o:
I think it would make more sense to go back to the menu after selecting a prompt. That might interfere with workflows I haven't thought about though...


karthink:
@nano-o returning to the menu is fine, but this approach cannot work.  It breaks the behavior of `M-x gptel-system-prompt` (try it.)

To do this correctly is tricky, this is why I haven't implemented it.  Let me know if you figure out a better solution.

PR #983, opened 2025-07-22
Added support for Bedrock API keys

andykuszyk:
[Bedrock API keys](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) offer a convenient alternative to authenticating with AWS to use Bedrock. This commit introduces an optional `:api-key` parameter for providing an API key, which takes precedence over other authentication methods.

### DIFF
--- a/gptel-bedrock.el
+++ b/gptel-bedrock.el
@@ -36,7 +36,7 @@
 (cl-defstruct (gptel-bedrock (:constructor gptel--make-bedrock)
                              (:copier nil)
                              (:include gptel-backend))
-  model-region)
+  model-region api-key)
 
 (defconst gptel-bedrock--prompt-type
   ;; For documentation purposes only -- this describes the type of prompt objects that get passed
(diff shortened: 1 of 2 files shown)


karthink:
Thanks for the PR, @andykuszyk.

1. All gptel-backend structs including gptel-bedrock already provide a `:key` field, there is no need to add `:api-key`.
2. I'm not sure about how this interacts with the now many other ways of authenticating provided by gptel-bedrock.  @felipeochoa, @akssri: thoughts?

andykuszyk:
> 1. All gptel-backend structs including gptel-bedrock already provide a `:key` field, there is no need to add `:api-key`.

Ah right, I didnt realise that! I'll test Bedrock access using `:key` and see how it goes 👍

karthink:
>> 1. All gptel-backend structs including gptel-bedrock already provide a `:key` field, there is no need to add `:api-key`.
>
> Ah right, I didnt realise that! I'll test Bedrock access using `:key` and see how it goes 👍

I doubt it will work, since AFAICT the auth functionality isn't implemented for :key.  I was commention only on the name of the field.  The authors of gptel-bedrock (pinged above) might have more information for us.

andykuszyk:
> > 1. All gptel-backend structs including gptel-bedrock already provide a `:key` field, there is no need to add `:api-key`. Ah right, I didnt realise that! I'll test Bedrock access using `:key` and see how it goes 👍
> I doubt it will work, since AFAICT the auth functionality isn't implemented for :key. I was commention only on the name of the field. The authors of gptel-bedrock (pinged above) might have more information for us.

You're right, it doesn't work as-is:

image

gptel-bedrock.el
```diff
@@ -628,11 +637,12 @@ STREAM - Whether to use streaming responses or not."
            :header nil           ; x-amz-security-token is set in curl-args if needed
            :models (gptel--process-models models)
 	   :model-region model-region
+           :api-key api-key
```
felipeochoa, line 640: I think this just needs to be replaced with `:key api-key` to use the default gptel field. But can you call this `bedrock-api-key` so it's clear what it refers to?

felipeochoa:
I think it would be better to not touch `gptel-bedrock--get-credentials` since the AWS credential search space is complex enough. 

Better would be to handle this directly in `gptel-bedrock--curl-args` using the built-in `:key`  field as @karthink mentioned. Something like:

```elisp
(defun gptel-bedrock--curl-args (region backend)
  "Generate the curl arguments to get a bedrock request signed for use in REGION."
  (let* ((api-key (gptel-backend-key backend))
         (auth-args (if api-key (new-code) (existing-code))))
   (nconc etc.)))
```

gptel-bedrock.el
andykuszyk: @felipeochoa I've refactored this PR based on your comment, and changed the name of this parameter 👍

andykuszyk:
@felipeochoa @karthink I've re-worked this PR based on your feedback (thanks for your advice 🙏 🙂 ).

I've also tested this locally with both a Bedrock API key and an AWS profile for authentication, and it seems to work fine for both 👍

karthink:
@andykuszyk I can merge this with a minimal review once @felipeochoa or @akssri give us the go-ahead.

gptel-bedrock.el
```diff
@@ -635,7 +642,9 @@ parameters (as plist keys) and values supported by the API."
           (gptel--make-bedrock
            :name name
            :host host
-           :header nil           ; x-amz-security-token is set in curl-args if needed
```
leezu, line 638: Why not set the bearer token in curl-args as well to keep the logic centralized? See https://github.com/karthink/gptel/pull/1007/files as an example

andykuszyk: Good idea 👍

I guess we could allow the API key to be set as either an elisp parameter, or as an environment variable?

akssri:
(Moved to #1053)

andykuszyk:
Closed in favour of https://github.com/karthink/gptel/pull/1053

Issue #909, opened 2025-06-14
When using evil mode, "Send selection" goes away in transient menu on any menu choices


Martinsos:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

If I select a line or region, when in evil-mode, and then run `gptel-menu` which opens a transient menu, it says at the bottom "RET Send selection (lines 5-6) ..." which is what I want.

However, if I then change any of the settings in the menu, the bottom text changes to "RET Send <buffername> (lines 1-5) ...", which is the behaviour gptel normally exibits when nothing is selected (send buffer from start to the point).

There is no way for me to pick selection again, in "<Prompt from".

### Backend

None

### Steps to Reproduce

0. Have evil-mode on.
1. Select a line in any buffer.
2. Run `gptel-menu`.
3. Notice that "selection" is currently ready to be sent.
4. Change any setting in the transient menu.
5. Notice that "selection" is not anymore ready to be sent, instead buffer to the point is.
6. There is no way to pick "selection" again (e.g. from "<Prompt from").

When evil-mode is off, it works ok.

### Additional Context

Emacs version: 30.1
OS: Archlinux
Problem manifests when evil-mode is enabled, otherwise doesn't.

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

Martinsos:
I would be happy to contribute a fix for this, although I could use a bit of guidance on it!

I tried looking at the code in the gptel-transient.el, and found that `gptel--describe-suffix-send` constructs the string describing what is going to be sent, but I don't think it actually influences what is going to get send if I got that right.
It is testing for `(region-active-p)` so obviously that stops being true for some reason when anything is chosen in the transient menu, but I am not sure what would be affecting that.

karthink:
> It is testing for `(region-active-p)` so obviously that stops being true for some reason when anything is chosen in the transient menu, but I am not sure what would be affecting that.

The region in the source buffer should remain active no matter what key you press in the transient menu (which is another buffer).  So I'm not sure what's going on.

I don't use evil-mode.  With evil-mode turned on, if you select a line in one buffer/window and press a key in another window, does the selection disappear?

Martinsos:
I have been using evil-mode for years now, first in spacemacs, now in my vanilla config, and never hit issues like this, so they are not typical.

[...]

PR #987, opened 2025-07-23
Fix "Send selection" indicator in transient menu when using evil-mode (#909)

EyoelYT:
## Description
When using evil-mode's line-wise visual selection (`M-x evil-visual-line`), the transient menu's bottom text temporarily shows the correct region ("Send selection (lines X–Y)") but then a different range as soon as any menu option or interaction occurs (use the arrow keys for testing).

This patch modifies `gptel--describe-suffix-send` to detect when evil's visual state is active and use `evil-visual-beginning`/`evil-visual-end` for computing line numbers. The fallback to `use-region-p` remains for ordinary selections.

## Testing
1. emacs -Q; install evil-mode and gptel.
2. Open buffer, press V on a line (`M-x evil-visual-line`).
3. M-x gptel-menu → Indicator shows "Send selection (lines N–M)".
4. Change any transient option → Indicator should still show the original line range.


karthink:
This is a bug either with evil-mode or Transient, and I'd rather it be reported there instead of adding a special case here.  Could you bring it up on the Transient repo instead?

EyoelYT:
I was just looking into this, and I don't think it is a bug. It is coming from evil-mode. I think evil-mode just handles visual regions in a different way.

You can see by playing around with this in a fresh emacs instance that has evil installed:
```emacs-lisp
(defun region-debug-toggle ()
  (interactive)
  (if (not (bound-and-true-p debugger-timer--object))
      (progn
        (message "turned on")
        (setq debugger-timer--object
            (run-with-timer
             1 1
             (lambda ()
               (message
                " RegiBeg: %s | RegiEnd: %s\n eVisBeg: %s | eVisEnd: %s"
                (and (use-region-p) (region-beginning))
                (and (use-region-p) (region-end))
                (and (evil-visual-state-p) (marker-position evil-visual-beginning))
                (and (evil-visual-state-p) (marker-position evil-visual-end)))))))
    (message "turned off")
    (cancel-timer debugger-timer--object)
    (setq debugger-timer--object nil)))
```

Observe the region values when using `M-x evil-visual-char`, `M-x evil-visual-block`, and especially when using `M-x evil-visual-line`; by moving the point around whilst in evil's visual-state. The region boundaries have different values almost each time. You can go to those points with `M-x goto-char`.

`region-beginning` and `region-end` use the `mark` and the `point`, while evil completely uses its own thing to demarcate regions.

I would imagine there would be all kinds of off-by-one-or-more errors due to this when using evil.

I would say, in order to solve this and support both evil and emacs regions, a simple solution would be to define gptel's own region functions within a utils file, e.g. functions like `gptel-region-active-p`, `gptel-region-end`, `gptel-region-beginning`?? I'd be willing to give it a try. 

I will also understand if this is a little too much (out of scope) for the project.

EyoelYT:
I'll be closing this PR as the issue can be solved with this snippet, which affects how region delimiters are calculated depending on whether we are in evil's visual-state or not

```emacs-lisp
(defun +evil-region-beginning-a (orig-fn)
  "Return beginning position of selection or evil visual region"
  (or (and (fboundp 'evil-visual-state-p)
           (evil-visual-state-p)
           (markerp evil-visual-beginning)
           (marker-position evil-visual-beginning))
      (funcall orig-fn)))

(defun +evil-region-end-a (orig-fn)
  "Return beginning position of selection or evil visual region"
  (or (and (fboundp 'evil-visual-state-p)
           (evil-visual-state-p)
           (markerp evil-visual-end)
           (marker-position evil-visual-end))
      (funcall orig-fn)))

(advice-add 'region-beginning :around #'+evil-region-beginning-a)
(advice-add 'region-end :around #'+evil-region-end-a)
```

Issue #982, opened 2025-07-22
How to logout of copilot?


khinshankhan:
**Briefly describe what you are trying to do**

I was logged into copilot on company github account, but recently they're moving towards allowing personal accounts to be on the plan. I'd like to log out of the existing account and into my personal.

With https://github.com/copilot-emacs/copilot.el there was simply a command like `M-x copilot-logout` and `M-x copilot-login`. I don't see anything similar for gptel?

**Additional context**
Emacs version: GNU Emacs 29.1 (build 1, aarch64-apple-darwin23.4.0, Carbon Version 170 AppKit 2487.5) of 2024-06-05
Operating system: macos sonoma 14.4 (23E214)

karthink:
@kiennq Could you address this?

kiennq:
You can simply eval `(gptel--gh-login)` and it will prompt you to login with the new account.
A more complete solution would be:
- Delete the `gptel-gh-xxx-token-file`s
- `(setf (gptel--gh-xxx-token gptel-backend) nil`

khinshankhan:
Awesome, seems the token files were saved in `~/.emacs.d/.cache/copilot-chat` so I just `rm -rf` that.

Running `(setf (gptel--gh-xxx-token gptel-backend) nil)` has a few errors like
```lisp
Debugger entered--Lisp error: (void-function \(setf\ gptel--gh-xxx-token\))
  (\(setf\ gptel--gh-xxx-token\) nil v)
  (let* ((v gptel-backend)) (\(setf\ gptel--gh-xxx-token\) nil v))
  (progn (let* ((v gptel-backend)) (\(setf\ gptel--gh-xxx-token\) nil v)))
  eval((progn (let* ((v gptel-backend)) (\(setf\ gptel--gh-xxx-token\) nil v))) t)
  elisp--eval-last-sexp(nil)
  eval-last-sexp(nil)
  funcall-interactively(eval-last-sexp nil)
  call-interactively(eval-last-sexp nil nil)
  command-execute(eval-last-sexp)
```

So I opted to just run
```lisp
(setq gptel--gh-xxx-token nil)
```
(not sure why you would reset `gptel-backend`?)

Anyway, after those 2 steps, running `gpt-send` prompted for the token which lets me choose account to use. Thanks 👍 

This issue can be closed but I recommend having these steps documented or making 2 interactive functions 👀

karthink:
[...]

PR #989, opened 2025-07-24
Add GitHub Copilot interactive login

khinshankhan:
In reference to https://github.com/karthink/gptel/issues/982


karthink:
@khinshankhan thanks for the PR.

I don't think we need a `gptel-gh-login` and a separate `gptel--gh-login`.  My idea in #982 was to turn `gptel--gh-login` into `gptel-gh-login`:

> @khinshankhan: Does the first solution (`gptel--gh-login`) work for you?  If it does, it might be better to **make that an interactive function** so the user can just run `M-x gptel-gh-login` and not worry about deleting tokens and resetting gptel-backends.

Please rename `gptel--gh-login` and make it interactive instead, unless there's some reason that won't work.

khinshankhan:
No issue, I just misunderstood. I've updated it to combine the functions.

karthink:
@khinshankhan Thanks for the PR!

PR #992, opened 2025-07-25
gptel: support function values in gptel presets

ahmed-shariff:
This PR updates gptel--apply-preset to allow function values in presets. This applies to :backend, :tools, and other general preset keys during their setting process. This is the diff I have in #905 and is an alternative to #940.


karthink:
Thanks for the PR.  I haven't forgotten about it, just have some reservations and unsolved questions.  I've replied in detail in #905.

PR #996, opened 2025-08-01
gptel-bedrock: allow configuring AWS profile via gptel-make-bedrock

lgfang:
The `gptel-make-bedrock` function now accepts a `:profile` keyword argument. This allows an AWS profile to be specified directly in the backend configuration, overriding any environment variables.

The credential resolution order has been updated to prioritize the most specific configuration:
1. Explicit profile in backend or request parameters
2. `AWS_PROFILE` environment variable
3. `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` environment variables

This makes managing AWS credentials more flexible and robust, especially for users who interact with multiple AWS accounts or prefer to keep their configuration self-contained within Emacs.

Here are the key benefits:

- Context-Specific Configurations: Users often work with different AWS accounts for various purposes (e.g., a personal account, a work development account, and a work production account). This feature allows you to define multiple `gptel` backends, each configured to use a specific AWS profile. You can then easily switch between these backends within Emacs without altering your system-wide environment variables.

- Self-Contained Emacs Configuration: For users who prefer to manage their entire workflow within Emacs, this change allows them to define AWS profiles directly in their Emacs configuration files (e.g., `init.el`). This eliminates the dependency on external shell environment variables like `AWS_PROFILE`, making the setup cleaner and more portable across different machines.

- Clarity and Reproducibility: Explicitly defining the profile in the backend configuration makes it transparent which set of credentials is being used. This is particularly valuable for literate programming or computational notebooks, as it documents the execution context directly alongside the code, improving reproducibility.

- Override Global Defaults: A user might have a default `AWS_PROFILE` set in their shell for general command-line use. This feature provides a convenient way to override that default for tasks performed within Emacs, without affecting the environment in their terminal.


akssri:
(Moved to #1053)

karthink:
Closing in favor of #1053.

PR #998, opened 2025-08-05
README: document using current window for dedicated chat buffer

jasalt:
I was looking for this behavior for a while and did not find it very well documented.

Thinking of the placement, could also be part of e.g. https://github.com/karthink/gptel/wiki/Tweaking-the-UI but first suggestion would be to add this README.md section.

### DIFF
--- a/README.org
+++ b/README.org
@@ -1200,6 +1200,8 @@ That's it. You can go back and edit previous prompts and responses if you want.
 
 The default mode is =markdown-mode= if available, else =text-mode=.  You can set =gptel-default-mode= to =org-mode= if desired.
 
+By default, a new window is created for the chat buffer and it can be changed to use current window by setting Chat UI option `(setq gptel-display-buffer-action '(display-buffer-same-window))`. 
+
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR!

This is incorrect, a new window is not created -- by default, gptel will open a new buffer according to your `display-buffer-alist` settings.

I also think it makes sense to add this to the FAQ (further below) instead of this section.  I want the user's experience of reading the "Usage" section to be as fast as possible.  They can come back and search for the option they need after they have gptel working.

It is also better to specify an action function that ensures that it stays selected, for example:

``` emacs-lisp
(setopt gptel-display-buffer-action
        '(pop-to-buffer-same-window))
```

`display-buffer-same-window` might move the focus depending on other hooks and `display-buffer-alist` settings.

Could you add this to the FAQ instead?

karthink:
@jasalt Are you still interested in this PR?

jasalt:
@karthink Excuse me for the delay. I added the suggested changes, thank you for the corrections. Further suggestions for improvement are welcome.

Issue #1001, opened 2025-08-05
subsequent request fails if tool calls results are printed to the buffer and tool call contains symbols with accents


SamuelVanie:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

While having `gptel-include-tool-results` set to `t`, if history contains tool call with a character that has an accent the next call will fail  

### Backend

Other (please specify in Additional Context)

### Steps to Reproduce

1. Set gptel-include-tool-results to t
2. provoke a tool call that will use characters with accent (writing `café` to a file)
3. Try making another call (writing something else to the same file)
4. The encoded characters \303\251 will cause a json error

### Additional Context

curl 8.15.0
gptel-backend : gh-copilot (same with Openrouter too)

### Backtrace

```emacs-lisp
Debugger entered--Lisp error: (wrong-type-argument json-value-p "{\"content\":\"# Translations\\n\\n- coffee: caf\303\251\\n- Ivory Coast: C\303\264te d'Ivoire\\n- destroy: d\303\251truire\\n- mouse: souris\\n- keyboard: clavier\\n- tea: th\303\251\\n- spoon: cuill\303\250re\\n- maybe: peut-\303\252tre\\n- television: t\303\251l\303\251vision\\n- state: \303\251tat\\n- defense: d\303\251fense\\n- landing-on: atterrissage sur\\n- phone: t\303\251l\303\251phone\\n\",\"path\":\"/home/vanieb/projects/gbonhi/translation.md\"}")
[...]

PR #1004, opened 2025-08-05
fix: accentuated characters encoded values causes error if in history

SamuelVanie:
- Add a utility function to convert values into utf8 characters before adding them to history (buffer)
- Make sure these characters are converted before passing them as data to curl

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -23,6 +23,10 @@
 
 ** New models and backends
 
+- Add support for ~gpt-5~, ~gpt-5-mini~ and ~gpt-5-nano~.
+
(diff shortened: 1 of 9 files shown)


SamuelVanie:
I've made a fix for the issue #1001 that I've identified. Now having characters with accent in tool call results or anywhere in the history doesn't causes error even with `gptel-include-tool-results` set to `t`.

karthink:
@SamuelVanie Thanks for the PR!

I've attempted to fix it much more simply, by just decoding the tool arguments before they are encoded again as part of the payload.

Closing this PR as not required, but if my solution doesn't work as expected we can revisit this.

PR #1005, opened 2025-08-07
gptel: preserve global values within `gptel-with-preset`

kmontag:
Instead of binding symbols targeted by a preset to `nil` initially, bind them to their real current value.

Currently this change is only relevant for presets using `:tools` with the `:append` keyword (since all other relevant values will simply be overridden when applying the preset). If some version of dynamic/functional presets ends up getting merged, this would be relevant for a wider range of cases.

Here's a snippet that can be run to demonstrate the difference:

``` emacs-lisp
(progn
  (gptel-make-tool
   :name "global_tool"
   :function (lambda ())
   :description "global tool"
   :args nil)
  (gptel-make-tool
   :name "preset_tool"
   :function (lambda ())
   :description "preset tool"
   :args nil)

  (let ((gptel-tools (list (gptel-get-tool "global_tool"))))
    (message "Outside gptel-with-preset: %s"
             (mapcar #'gptel-tool-name gptel-tools))
    (gptel-with-preset
     '(:tools (:append "preset_tool"))
     (message "Inside gptel-with-preset: %s"
              (mapcar #'gptel-tool-name gptel-tools)))))
```

Output without this change:

``` text
Outside gptel-with-preset: (global_tool)
Inside gptel-with-preset: (preset_tool)
```

Output after applying this change:

``` text
Outside gptel-with-preset: (global_tool)
Inside gptel-with-preset: (global_tool preset_tool)
```

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -3800,9 +3800,13 @@ from a gptel preset applied.
 NAME is the name of a preset, or a spec (plist) of the form
  (:KEY1 VAL1 :KEY2 VAL2 ...).  It must be quoted."
   (declare (indent 1))
-  `(cl-progv (gptel--preset-syms ,name) nil
-    (gptel--apply-preset ,name)
-    ,@body))
+  ;; Use a unique uninterned symbol so that our let-binding doesn't
+  ;; interfere with symbols in the body.
+  (let ((preset-syms (gensym "gptel--preset-syms")))
+    `(let ((,preset-syms (gptel--preset-syms ,name)))
+       (cl-progv ,preset-syms (mapcar #'symbol-value ,preset-syms)
+         (gptel--apply-preset ,name)
+         ,@body))))
 
 ;;;; Presets in-buffer UI
 (defun gptel--transform-apply-preset (_fsm)


karthink:
@kmontag thanks for catching this!

> If some version of dynamic/functional presets ends up getting merged, this would be relevant for a wider range of cases.

Yes, this is planned.  Including yours, there are at least three different designs for this in the works.  I'm trying to find the best one.

Your change looks fine, but I had some questions.  

- Is the reason for the gensym and the outer let block to avoid calling `gptel--preset-syms` twice?
``` emacs-lisp
`(let ((,preset-syms (gptel--preset-syms ,name)))
   (cl-progv (gptel--preset-syms ,name) ;=> EVALUATION 1
       (mapcar #'symbol-value (gptel--preset-syms ,name)) ;=> EVALUATION 2
     (gptel--apply-preset ,name)
     ,@body))
```

- Is there some reason you don't use the simpler `(make-symbol "preset-syms")` instead of `gensym`?  IIUC there is no chance of a name collision since the symbol is not-interned.

- Also, it seems to me that there should be an easier way to let-bind a bunch of variables to their _current_ values than constructing a new list with `(mapcar #'symbol-value ...)`.  I took a look at `macroexp-let2*` but couldn't figure out how to use it here.

karthink:
Here is an example of what I mean.

Based on the definition of `cl-progv`, here's a version of `gptel-with-preset` that does not use an extra gensym or cons a list of `symbol-value`s only to quote them immediately.  It will help to compare against the definition of `cl-progv`.  It is identical except that `cl-progv` quotes its VALUES argument, we don't.

``` emacs-lisp
(defmacro gptel-with-preset (name &rest body)
  "Run BODY with gptel preset NAME applied.

This macro can be used to create `gptel-request' command with settings
from a gptel preset applied.

NAME is the name of a preset, or a spec (plist) of the form
 (:KEY1 VAL1 :KEY2 VAL2 ...).  It must be quoted."
  (declare (indent 1))
  (let ((syms (make-symbol "syms"))
        (binds (make-symbol "binds"))
        (bodyfun (make-symbol "body")))
    `(let* ((,syms (gptel--preset-syms ,name))
            (,bodyfun (lambda () (gptel--apply-preset ,name) ,@body))
            (,binds nil))
       (while ,syms (push (list (car ,syms) (pop ,syms)) ,binds)) ;=> HERE `cl-progv' DIFFERS
       (eval (list 'let (nreverse ,binds) (list 'funcall (list 'quote ,bodyfun)))))))
```

kmontag:
>
>     * Is the reason for the gensym and the outer let block to avoid calling `gptel--preset-syms` twice?
> 

Yes, exactly.

> 
>     * Is there some reason you don't use the simpler `(make-symbol "preset-syms")` instead of `gensym`?  IIUC there is no chance of a name collision since the symbol is not-interned.
>

The only reason is this line from the [relevant docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Symbols.html): "To avoid problems when accidentally interning printed representation of generated code (see [Printed Representation and Read Syntax](https://www.gnu.org/software/emacs/manual/html_node/elisp/Printed-Representation.html)), it is recommended to use gensym instead of make-symbol."

Otherwise I don't know of any reason that `gensym` is necessary/preferable to `make-symbol`.
 
>
>     * Also, it seems to me that there should be an easier way to let-bind a bunch of variables to their _current_ values than constructing a new list with `(mapcar #'symbol-value ...)`.  I took a look at `macroexp-let2*` but couldn't figure out how to use it here.
>

I was also a bit bothered by this but couldn't find a better way using `cl-progv`. But your example looks like it would do the job just fine!

karthink:
> I was also a bit bothered by this but couldn't find a better way using `cl-progv`. But your example looks like it would do the job just fine!

Let's go with my version then?  It's ugly but that's `cl-progv` for you 🤷🏼

Using `cl-progv` just hides the ugliness, I'm fine with it in `gptel-with-preset` instead.

kmontag:
Yep, fine by me 👍

karthink:
I've made the change.

> The only reason is this line from the [relevant
> docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Symbols.html):
> "To avoid problems when accidentally interning printed representation of
> generated code (see [Printed Representation and Read
> Syntax](https://www.gnu.org/software/emacs/manual/html_node/elisp/Printed-Representation.html)),
> it is recommended to use gensym instead of make-symbol."
> 
> Otherwise I don't know of any reason that `gensym` is necessary/preferable to
> `make-symbol`.

Ah I see.  This is just guarding against the typo
``` emacs-lisp
(let ((foo (make-symbol "foo")))
  `(... foo ...))
```
instead of what we want:
``` emacs-lisp
(let ((foo (make-symbol "foo")))
  `(... ,foo ...))
```

kmontag:
Ah that makes sense, I wasn't clear on whether `make-symbol` is more likely to interfere with bindings in the `@body` in certain forms. But I think you're right.

Thanks!

PR #1007, opened 2025-08-07
gptel-bedrock: AWS_BEARER_TOKEN_BEDROCK support

leezu:
AWS has introduced Bearer token based authentication for Bedrock https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html Bearer token based authentication only requires the `Authorization: Bearer` header and does not need the sigv4 signing and thus removes the current dependency of the Bedrock backend on curl >= 8.9.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -70,6 +70,11 @@
 - gptel now handles Ollama models that return both reasoning content
   and tool calls in a single request.
 
+- ~gptel-make-bedrock~ now checks for the ~AWS_BEARER_TOKEN_BEDROCK~ environment
+  variable parameter and uses it for Bedrock API key based authentication if
+  present. See
+  https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html.
+
 * 0.9.8.5 2025-06-11
 
 ** Breaking changes
--- a/README.org
+++ b/README.org
@@ -1060,10 +1060,10 @@ Register a backend with
   :model-region 'apac)
 #+end_src
 
-The Bedrock backend gets your AWS credentials from the environment variables. It expects to find either
+The Bedrock backend gets your AWS credentials from the environment variables. If ~AWS_BEARER_TOKEN_BEDROCK~ is present, it uses the bearer token. Otherwise it expects to find either
 ~AWS_ACCESS_KEY_ID~, ~AWS_SECRET_ACCESS_KEY~, ~AWS_SESSION_TOKEN~ (optional), or if present, can use ~AWS_PROFILE~ to get these directly from the ~aws~ cli.
 
-NOTE: The Bedrock backend needs curl >= 8.5 in order for the sigv4 signing to work properly,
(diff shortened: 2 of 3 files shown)


karthink:
Thanks for the PR @leezu.  Is this a duplicate of #983?

leezu:
@karthink they are similar. This PR adds support for the AWS_BEARER_TOKEN_BEDROCK environment variable which is the standard environment variable name for Bedrock https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-use.html #983 adds support for passing the key as an argument in elisp. Ideally we should support both ways to specify the key.

The other difference between the PRs is that #983 uses :header argument to pass the token AND modifies the gptel-bedrock--curl-args to remove the other arguments if bearer token is present, whereas this PR handles both inside gptel-bedrock--curl-args. I think handling both in gptel-bedrock--curl-args is cleaner.

leezu:
Also FYI, Bedrock now supports an OpenAI compatible endpoint (at least for the gpt-oss models): https://aws.amazon.com/blogs/aws/openai-open-weight-models-now-available-on-aws/ 

```
(gptel-make-openai "Bedrock-OpenAI"
    :host "bedrock-runtime.us-west-2.amazonaws.com"
    :protocol "https"
    :key "REDACTED"
    :endpoint "/openai/v1/chat/completions"
    :stream nil
    :models '((openai.gpt-oss-20b-1:0 :capabilities (reasoning))
              (openai.gpt-oss-120b-1:0 :capabilities (reasoning))))
```

Additional patches are required for the reasoning support with GPT-OSS as it uses `<reasoning> </reasoning>` tags instead of the currently supported `<think> </think>`.

karthink:
> leezu left a comment (karthink/gptel#1007)
>
> @karthink they are similar. This PR adds support for the AWS_BEARER_TOKEN_BEDROCK environment variable which is the standard environment variable name for Bedrock https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-use.html #983 adds support for passing the key as an argument in elisp. Ideally we should support both ways to specify the key.
>
> The other difference between the PRs is that #983 uses :header argument to pass the token AND modifies the gptel-bedrock--curl-args to remove the other arguments if bearer token is present, whereas this PR handles both inside gptel-bedrock--curl-args. I think handling both in gptel-bedrock--curl-args is cleaner.

Thanks for the clarification.  Any design changes to gptel-bedrock are the purview of @felipeochoa and @akssri, so I'll defer to them on this proposal.

leezu:
Thanks, @karthink. It's not really a design change but simply adjustment to support the new API key functionality. It became available only last month: https://aws.amazon.com/blogs/machine-learning/accelerate-ai-development-with-amazon-bedrock-api-keys/

akssri:
(Moved to #1053)

karthink:
Closing in favor of #1053.

PR #1013, opened 2025-08-08
gptel-gh: update model list

kiennq:
Update model list for github copilot to include gpt-5


karthink:
Thanks @kiennq!  Could you also update the NEWS file?

kiennq:
@karthink I saw that we have entries for `Update gpt-5 and opus-41` in the NEWS file already, albeit they're for open AI and Anthropic (not explicitly mentioned) respectively.
Do you think we need to add news for Github Copilot?

karthink:
Oh you're right, we don't need a separate entry as we already mention the models.  I'll merge this.

karthink:
Thanks @kiennq!

PR #1015, opened 2025-08-08
Add .gptel file extension support

La-Loutre:
Automatically enables gptel-mode for files with a .gptel extension. This allows users to create new chat files or reload saved conversations without having to manually enable the mode.

I find this extremely useful. Simply open a new file named myconv.gptel, and you're chatting with the AI in a buffer that can be saved and reopened without running any command. You don't need to run M-x gptel anymore.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1595,6 +1595,36 @@ which see for BEG, END and PRE."
               gptel--old-header-line nil)
       (setq mode-line-process nil))))
 
+;;; Automatically enable `gptel-mode` for .gptel files.
+;;; This allows a user to start a new chat by simply opening a new file
+;;; ending in .gptel, or to resume a saved session without having to run
+;;; `M-x gptel-mode` manually.
+(defun gptel--auto-enable-mode-for-extension ()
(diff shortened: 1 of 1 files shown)


La-Loutre:
https://github.com/user-attachments/assets/dcb2a6a8-9c39-4b40-b7c3-350ace18c630

Igrom:
Why is it necessary to define a new mode?
- gptel-mode is supposed to be a minor mode, not a major mode
- you're pinning the mode as `text-mode`. If opening a new buffer by calling `gptel` based on the value of `gptel-default-mode`.

You can achieve the same with a one-liner that respects the value of `gptel-default-mode`. See https://github.com/karthink/gptel/compare/master...Igrom:gptel:feature/add-gptel-file-association-to-auto-mode-alist. (though I'm not sure if I'm not abusing `auto-mode-alist` here).

I'm not even convinced if that's the way to go. File associations could be added to the installation section in the README (e.g. to the `:init` section of `use-package`), as it is the case for many other packages.

@karthink FYI (esp. regarding the last point)

Issue #995, opened 2025-07-30
Add "rolling window" feature (infinite buffer)


La-Loutre:
Hi, so basically I vibe coded a few lines that leverage a gptel hook, from what I understand, such that I can have an "infinite" conversation with the LLM, even though the context size isn't infinite.

It just takes the last N lines, and no more. I know there is a feature to limit the number of past messages, but I want this to apply even to the first prompt, which might already be larger than the context size.

I found this extremely useful. I have conversations with 40k lines in files, where I can just get back to where I was without having to worry about the context size. I can still search through what was done in the buffer, even though the AI has a limited view of the whole content.

I have no idea how to integrate it further than these few lines that were vibe coded. It's fine for my use like this, but because I find this useful, I'm just posting the idea.

Here are the lines I added into my .spacemacs to make it work.
```lisp
  (defvar my-gptel-max-prompt-characters 68000
    "The maximum number of characters to send in a gptel prompt.")

  ;; This is the function that will do the work.
  (defun my-gptel-truncate-prompt ()
    "Truncate the gptel prompt buffer to `my-gptel-max-prompt-characters`."
    ;; Check if our character limit is set and if the buffer exceeds it
    (when (and (bound-and-true-p my-gptel-max-prompt-characters)
               (> (buffer-size) my-gptel-max-prompt-characters))
      ;; This message is for you to confirm the hook is running.
      ;; It will appear in the *Messages* buffer.
      (message "gptel: Truncating prompt to the last %s characters." my-gptel-max-prompt-characters)
      ;; This is the action: delete everything except the last N characters.
      (delete-region (point-min) (- (point-max) my-gptel-max-prompt-characters))))

  ;; This is the crucial line that "relies on the hook".
  ;; We are adding our custom function to the list of functions that gptel
  ;; will run right before sending the prompt.
  (add-hook 'gptel-prompt-transform-functions 'my-gptel-truncate-prompt)
```

karthink:
This is a neat idea, but I think gptel can do better.  For most models we know the context window size, so it should be possible to cut it off at an (approximate) model-specific token count.

La-Loutre:
I just did a PR for this.

PR #1016, opened 2025-08-09
Add rolling window prompt truncation

La-Loutre:
Automatically truncate prompts to fit model context with safety margin. 
Transient supports enabling/disabling or setting a custom fraction (0.1-1.0).

#995

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -720,6 +720,7 @@ Also format its value in the Transient menu."
     (gptel--infix-variable-scope)
     (gptel--infix-provider)
     (gptel--infix-max-tokens)
+    (gptel--infix-rolling-window-prompt)
     (gptel--infix-num-messages-to-send
      :if (lambda () (and gptel-expert-commands
                     (or gptel-mode gptel-track-response))))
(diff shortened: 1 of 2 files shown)


meain:
A better way to manage long chats might be to warn users when we are reaching the end of the context window and allowing them to call something to compact the chat. This is what tools like [claude-code](https://docs.anthropic.com/en/docs/claude-code/slash-commands) or [opencode](https://opencode.ai/docs/cli/#slash-commands) does.

La-Loutre:
I don't think there is an objectively better way; it really depends on who is in front of the computer. For my use case, I do not want to call commands or modify the chat history. I just want what worked, the conversation with the LLM, to continue working seamlessly, without interruption.

It's disabled by default, so it will not be used without the user's understanding and choice.

bajsicki:
I don't know about other APIs, but [llama.cpp](https://github.com/ggml-org/llama.cpp/tree/master/tools/server) supports getting the context window via the /props endpoint. It's the `n_ctx` key.

Could query that for accuracy.

Here's some code that should be easy to adapt (I use this for automatically switching presets/ model backends with llama.cpp):

```
#+begin_src elisp :results none
(defun gptel-got--model ()
  "This function retrieves the model name from the GPTel backend.

It sends a GET request to the backend's models endpoint to fetch the
available models. The function returns the model name (string) from the
first entry in the backend's model list, which should be sufficient
for local, single-model inference."
  (let ((result (gptel--url-retrieve (concat "http://" (gptel-backend-host gptel-backend) "/v1/models") :method "GET")))
    (file-name-nondirectory(plist-get (aref (plist-get result :models) 0) :model))))

(defun gptel-got--match-model ()
  "Finds the closest matching model name from the backend's model list.

This function uses string distance (i.e. Levenshtein distance) to compare
the target model name (from `gptel-got--model') against all models defined
for the backend. It returns the best-matching model name (string) from
the backend."

  (let ((got-model (gptel-got--model))
        (min most-positive-fixnum)
        (best-match nil))
    (cl-loop for model in (gptel-backend-models gptel-backend)
             do (let ((distance (string-distance got-model (symbol-name model))))
                  (when (< distance min)
		    (setq min distance)
		    (setq best-match model))))
    best-match))
#+end_src
```

La-Loutre:
It looks interesting to fetch from the server that info. Although I think it's a slightly different topic. 
This PR truncate the window based on gptel context.
Now is the gptel context set by fetching info from the server itself is a different question.

Issue #1019, opened 2025-08-09
gpt-5 models not handling `gptel-max-tokens`


JDNdeveloper:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

The following error occurs when using the `gpt-5` family of models:
```
ChatGPT error: ((HTTP/2 400) invalid_request_error) Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

I believe this line needs to be updated to include all `gpt-5` models: https://github.com/karthink/gptel/blob/9c9af2cdfbec0f55a14591f0b1ffab0ebcd25e5e/gptel-openai.el#L300

I'll put up a PR.

### Backend

None

### Steps to Reproduce

1. Set `gptel-max-tokens` to some value
2. Issue request

### Additional Context

Emacs 27

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

PR #1020, opened 2025-08-09
Mark gpt-5 models as reasoning models

JDNdeveloper:
Fixes https://github.com/karthink/gptel/issues/1019

### DIFF
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -297,7 +297,7 @@ Mutate state INFO with response metadata."
            :messages [,@prompts]
            :stream ,(or gptel-stream :json-false)))
         (reasoning-model-p ; TODO: Embed this capability in the model's properties
-         (memq gptel-model '(o1 o1-preview o1-mini o3-mini o3 o4-mini))))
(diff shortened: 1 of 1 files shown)


karthink:
@JDNdeveloper Thanks.  Should be `gpt-5 gpt-5-mini gpt-5-nano` and not `5 5-mini 5-nano`, shouldn't it?

EDIT: While you fix that please also split the list across two lines.  It's not a hard rule but I try to wrap at <80 chars.

JDNdeveloper:
> @JDNdeveloper Thanks. Should be `gpt-5 gpt-5-mini gpt-5-nano` and not `5 5-mini 5-nano`, shouldn't it?
> 
> EDIT: While you fix that please also split the list across two lines. It's not a hard rule but I try to wrap at <80 chars.

You're right, I had misread `o4-mini` as `4o-mini` and thought the `gpt-` wasn't needed. Fixed.

karthink:
@JDNdeveloper Thanks for the PR!

PR #1054, opened 2025-08-30
gptel-openai: Update deepseek model specs

FrauH0lle:
DeepSeek V3.1 has been released, introducing a bigger context window for all models and new costs. See
https://api-docs.deepseek.com/news/news250821.

* gptel-openai-extras.el (gptel-make-deepseek): Adjust `:context-window`, `:input-cost` and `:output-cost` to newly documented values.


karthink:
@FrauH0lle Thanks for the update!

Issue #971, opened 2025-07-16
Start MCP servers with preset


FrauH0lle:
Hi @karthink, you create a fantastic package. The more I use it, the better it seems to get! Really great work.

I would like to ask a bit for your help with what I am trying to do:
Start an MCP server and connect to it when a preset is invoked via "@my_preset" in a chat buffer. 

First, my setup and what I tried to do. 

```elisp
  (gptel-make-deepseek "DeepSeek"
    :stream t
    :key (gptel-api-key-from-auth-source "api.deepseek.com" "apikey"))

  (setopt mcp-hub-servers
         `(("sequential-thinking" . (:command "npx" :args ("-y" "@modelcontextprotocol/server-sequential-thinking")))
           ("memory-bank" . (:command "npx" :args ("-y" "@allpepper/memory-bank-mcp")
                             :env (:MEMORY_BANK_ROOT ,(expand-file-name (projectile-project-root)))))))

```

I found the issues #896 and #881 read your comments and suggestions. In #896 you asked

> I also don't understand why you want to start an MCP server via the @preset-name prompt. The server stays active as an Emacs child process afterwards. Why not just start all the MCP servers you need when you start up Emacs or gptel?

And the `memory-bank-mcp` is actually the reason why: The server requires that the env var `MEMORY_BANK_ROOT` is set to the path of the memory bank when it starts and this cannot be changed as soon as it is running, it requires a restart. 

My initial idea was to use the `:pre` keyword of `gptel-make-preset` to make sure the server is available. 

```elisp
  (gptel-make-deepseek "DeepSeek"
    :stream t
    :key (gptel-api-key-from-auth-source "api.deepseek.com" "apikey"))

  (setopt mcp-hub-servers
         `(("sequential-thinking" . (:command "npx" :args ("-y" "@modelcontextprotocol/server-sequential-thinking")))
           ("memory-bank" . (:command "npx" :args ("-y" "@allpepper/memory-bank-mcp")
                             :env (:MEMORY_BANK_ROOT ,(expand-file-name (projectile-project-root)))))))

(defvar +gptel-memory-mcp-last-conn nil)

(gptel-make-preset 'memory-bank
      :description "Use Memory Bank MCP to add and update project context"
      :backend "DeepSeek"
      :model 'deepseek-chat
      :pre (lambda ()
              (let ((mcp-hub-servers
                     `(("memory-bank" . (:command "npx" :args ("-y" "@allpepper/memory-bank-mcp")
[...]

PR #1055, opened 2025-08-30
gptel-integrations: Add sync mcp server init

FrauH0lle:
This change requires `mcp.el` commit
5f06a78fe74f58888f115bf30838ede5e013f4af or later. See https://github.com/lizqwerscott/mcp.el/commit/5f06a78fe74f58888f115bf30838ede5e013f4af.

* gptel-integrations.el (gptel-mcp-connect): Add `syncp` argument which gets passed to `mcp-hub-start-all-server` in order to initialize the mcp server synchronously.

This completes #971

### DIFF
--- a/gptel-integrations.el
+++ b/gptel-integrations.el
@@ -58,7 +58,8 @@ considered.
 If INTERACTIVE is non-nil (or called interactively), guide the user
 through setting up mcp, and query for servers to retrieve tools from.
 
-Call SERVER-CALLBACK after starting MCP servers."
+Call SERVER-CALLBACK after starting MCP servers.  If SERVER-CALLBACK is
+not a function and non-nil, start SERVERS synchronously."
   (interactive (list nil nil t))
   (if (locate-library "mcp-hub")
       (unless (require 'mcp-hub nil t)
@@ -119,8 +120,9 @@ Call SERVER-CALLBACK after starting MCP servers."
                       (when (functionp server-callback) (funcall server-callback))))))
 
             (if inactive-servers        ;start servers
-                (mcp-hub-start-all-server
-                 add-all-tools (mapcar #'car inactive-servers))
+                (let ((syncp (and server-callback (not (functionp server-callback)) t)))
+                  (mcp-hub-start-all-server
+                   add-all-tools (mapcar #'car inactive-servers) syncp))
               (funcall add-all-tools (mapcar #'car servers))))
         (message "All MCP tools are already available to gptel!")
         (when (functionp server-callback) (funcall server-callback))))))


karthink:
@FrauH0lle Thanks for the PR!

Under these changes, defining a preset that starts an MCP server if required looks like:
``` emacs-lisp
(gptel-make-preset  'memory-bank
  :description "Use  Memory  Bank  MCP  to  add  and  update  project  context"
  :backend     "DeepSeek"
  :models      'deepseek-chat
  :pre         (lambda () (gptel-mcp-connect '("memory-bank") nil nil 'sync)))
;;                                   servers╶─╯               │   │    │
;;                           server-callback╶─────────────────╯   │    │
;;                               interactive╶─────────────────────╯    │
;;                         synchronous start╶──────────────────────────╯
```

I think the calling convention for `gptel-mcp-connect` can be simplified.  The `server-callback` argument is meaningless when the call is synchronous, so it can be reused.  Perhaps the `server-callback` argument can be the symbol `sync` instead of a function to start synchronously.  Then the above call becomes
``` emacs-lisp
(gptel-mcp-connect '("memory-bank") 'sync)
;;         servers╶─╯                │
;;   callback/sync╶──────────────────╯ 
;;   interactive not required
```
which is much simpler.

We can also accept any truthy value that's not a function in place of `sync`.  What do you think?

FrauH0lle:
Hey @karthink 

I think I'll have some time tomorrow to give it a shot to make the function a bit more compact.

Another approach could be to make it a `cl-defun` and use keyword args. I wouldn't say that `server-callback` is fully meaningless. but can be used as a way to control the return value of `gptel-mcp-connect`. Instead of always `nil` it could be used to display a message or simply return `t`. 

But as you are the maintainer, I would like to leave this to you. I think both approaches are fine.

karthink:
> Another approach could be to make it a `cl-defun` and use keyword args.

gptel-mcp-connect is a command, and interactive functions cannot be cl-defuns.

> I wouldn't say that `server-callback` is fully meaningless.  but can be used as a way to control the return value of `gptel-mcp-connect`.  Instead of always `nil` it could be used to display a message or simply return `t`.  

I don't understand the purpose of a server-callback during synchronous initialization.  In your example, how is it different from displaying the message after running gptel-mcp-connect synchronously?  If you have a more specific example of when it can be useful we can retain the callback and use a different argument for sync.

FrauH0lle:
> gptel-mcp-connect is a command, and interactive functions cannot be cl-defuns.

And learned something :)

> I don't understand the purpose of a server-callback during synchronous initialization. In your example, how is it different from displaying the message after running gptel-mcp-connect synchronously? 

Hmm, true. I was thinking out loud and should have thought a bit longer... You are right, it does not make sense here. 

I'll be sitting in a train tomorrow for some hours, so I should be able to adjust the PR.

FrauH0lle:
@karthink I updated the PR according to your feedback.

FrauH0lle:
@karthink Quick ping, is there anything missing in this PR?

karthink:
@FrauH0lle No, it looks good.  Merged, thanks for the PR!

FrauH0lle:
Thank you very much!

PR #1058, opened 2025-09-03
gptel-gh: Change how tokens are managed

marcus-lundgren:
Introduce functions that can be overloaded that controls how a token is restored/stored. This enables the user to make gptel-gh store the token in a secure manner, similar to what is recommended for the other backend variants.

In conjunction with this, proper support for multiple accounts has also been introduced by having each gptel-gh backend be marked with which account they should be considered to belong to.

**Old description:**
Removed the file based persistent storage of the tokens, as they were stored in clear text. They will only be stored in variables.

The initial value of the github-token can now be set by providing a function that provides it. This enables this token to be securely stored outside of Emacs.

### DIFF
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -319,7 +319,7 @@ instead of attempting to open a browser automatically."
   (let ((gh-backends (gptel--gh-get-backends-by-account-hint account-hint)))
     ;; It shall only be possible to login when there exists a corresponding backend
     (if (= (length gh-backends) 0)
-        (user-error "No GitHub CoPilot backend found for account hint '%s'" account-hint))
+        (user-error "No GitHub Copilot backend found for account hint '%s'" account-hint))
     (pcase-let (((map :device_code :user_code :verification_uri)
                  (gptel--url-retrieve
                      "https://github.com/login/device/code"
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -275,7 +275,7 @@
 (defun gptel-gh--get-save-token-function (account-hint token)
   (if #'gptel-oauth-token-save-function
       (gptel-oauth-token-save-function 'gptel-gh account-hint token)
-    (gptel--gh-save-token-to-file account-hint token))
+    (gptel--gh-save-token-to-file account-hint token)))
 
 ;; https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)
 (defun gptel--gh-uuid ()
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -257,10 +257,21 @@
    #'gptel-gh--get-load-token-function
    account-hint))
 
+(defun gptel--gh-validate-token-p (token)
+  "Check if TOKEN is a valid non-empty GitHub token string."
+  (and (stringp token) (not (string-empty-p token))))
+
 (defun gptel-gh--get-load-token-function (account-hint)
-  (if #'gptel-oauth-token-load-function
-      (gptel-oauth-token-load-function 'gptel-gh account-hint)
-    (gptel-oauth--read-token (gptel--gh-generate-token-filename account-hint))))
+  "Load GitHub token from file, ensuring it is a string."
+  (let ((token (if #'gptel-oauth-token-load-function
+                  (gptel-oauth-token-load-function 'gptel-gh account-hint)
+                (gptel-oauth--read-token (gptel--gh-generate-token-filename account-hint)))))
+    ;; Filter out nil and empty values (empty files do not contain tokens)
+    (when (and token (not (equal token "")))
+      (unless (gptel--gh-validate-token-p token)
+        (message "Ignoring invalid GitHub token format in file: expected string, got %S" token)
+        (setq token nil))
+      token)))
 
 (defun gptel--gh-save-token (account-hint token)
   "Function that updates the GitHub OAuth token cache and calls the save function."
@@ -273,6 +284,9 @@
    token))
 
 (defun gptel-gh--get-save-token-function (account-hint token)
+  "Save GitHub TOKEN to file for ACCOUNT-HINT, ensuring it is a string."
+  (unless (gptel--gh-validate-token-p token)
+    (user-error "Cannot save invalid GitHub token: expected string, got %S" token))
   (if #'gptel-oauth-token-save-function
       (gptel-oauth-token-save-function 'gptel-gh account-hint token)
     (gptel-oauth--write-token (gptel--gh-generate-token-filename account-hint) token)))
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -353,8 +353,7 @@ Then we need a session token."
                (gptel--gh-token gptel-backend)))
     (when (or (null token)
               (and expires_at
-                   (> (round (float-time (current-time)))
-                      expires_at)))
+                   (> (round (float-time)) expires_at)))
       (gptel--gh-renew-token))))
 
 (cl-defmethod gptel-curl--parse-stream ((backend gptel--gh) info)
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -275,7 +275,7 @@
 (defun gptel-gh--get-save-token-function (account-hint token)
   (if #'gptel-oauth-token-save-function
       (gptel-oauth-token-save-function 'gptel-gh account-hint token)
-    (gptel--gh-save-token-to-file account-hint token)))
+    (gptel-oauth--write-token (gptel--gh-generate-token-filename account-hint) token)))
 
 ;; https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)
 (defun gptel--gh-uuid ()
@@ -302,10 +302,6 @@
    gptel-gh-github-token-file
    #'gptel-oauth--validate-account-hint
    account-hint))
-
-(defun gptel--gh-save-token-to-file (account-hint token)
-  "Save GitHub TOKEN to file using ACCOUNT-HINT."
-  (gptel-oauth--save-token-to-file #'gptel--gh-generate-token-filename account-hint token))
 
 (defun gptel-gh-login (account-hint)
   "Login to GitHub Copilot API.
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -263,7 +263,7 @@
 
 (defun gptel-gh--get-load-token-function (account-hint)
   "Load GitHub token from file, ensuring it is a string."
-  (let ((token (if #'gptel-oauth-token-load-function
+  (let ((token (if gptel-oauth-token-load-function
                   (funcall gptel-oauth-token-load-function 'gptel-gh account-hint)
                 (gptel-oauth--read-token (gptel--gh-generate-token-filename account-hint)))))
     ;; Filter out nil and empty values (empty files do not contain tokens)
@@ -287,7 +287,7 @@
   "Save GitHub TOKEN to file for ACCOUNT-HINT, ensuring it is a string."
   (unless (gptel--gh-validate-token-p token)
     (user-error "Cannot save invalid GitHub token: expected string, got %S" token))
-  (if #'gptel-oauth-token-save-function
+  (if gptel-oauth-token-save-function
       (funcall gptel-oauth-token-save-function 'gptel-gh account-hint token)
     (gptel-oauth--write-token (gptel--gh-generate-token-filename account-hint) token)))
 
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -264,7 +264,7 @@
 (defun gptel-gh--get-load-token-function (account-hint)
   "Load GitHub token from file, ensuring it is a string."
   (let ((token (if #'gptel-oauth-token-load-function
-                  (gptel-oauth-token-load-function 'gptel-gh account-hint)
+                  (funcall gptel-oauth-token-load-function 'gptel-gh account-hint)
                 (gptel-oauth--read-token (gptel--gh-generate-token-filename account-hint)))))
     ;; Filter out nil and empty values (empty files do not contain tokens)
     (when (and token (not (equal token "")))
@@ -288,7 +288,7 @@
   (unless (gptel--gh-validate-token-p token)
     (user-error "Cannot save invalid GitHub token: expected string, got %S" token))
   (if #'gptel-oauth-token-save-function
-      (gptel-oauth-token-save-function 'gptel-gh account-hint token)
+      (funcall gptel-oauth-token-save-function 'gptel-gh account-hint token)
     (gptel-oauth--write-token (gptel--gh-generate-token-filename account-hint) token)))
 
 ;; https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -300,7 +300,6 @@
   "Generate token filename for GitHub backend with ACCOUNT-HINT."
   (gptel-oauth--generate-token-filename
    gptel-gh-github-token-file
-   #'gptel-oauth--validate-account-hint
    account-hint))
 
 (defun gptel-gh-login (account-hint)
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -227,17 +227,11 @@
 (cl-defstruct (gptel--gh (:include gptel-openai)
                          (:copier nil)
                          (:constructor gptel--make-gh))
-  token github-token sessionid machineid responses-backend)
+  token github-token sessionid machineid account-hint responses-backend)
 
 (defcustom gptel-gh-github-token-file (expand-file-name ".cache/copilot-chat/github-token"
                                                         user-emacs-directory)
   "File where the GitHub token is stored."
-  :type 'string
-  :group 'gptel)
-
-(defcustom gptel-gh-token-file (expand-file-name ".cache/copilot-chat/token"
-                                                 user-emacs-directory)
-  "File where the chat token is cached."
   :type 'string
   :group 'gptel)
 
@@ -247,6 +241,31 @@
     ("User-Agent" . ,(format "Emacs %s" emacs-version))))
 
 (defconst gptel--gh-client-id "Iv1.b507a08c87ecfe98")
+(defconst gptel--gh-default-username-placeholder "[Default account]")
+
+(defun gptel--gh-get-backends-by-account-hint (account-hint)
+  "Get all GitHub Copilot backends for a specific account hint."
+  (gptel-oauth--get-backends-by #'gptel--gh-p #'gptel--gh-account-hint account-hint))
+
+(defun gptel--gh-load-token (account-hint)
+  "Function that ensures that the GitHub OAuth token cache is used and is set."
+  (gptel-oauth--load-token
+   #'gptel--gh-p
+   #'gptel--gh-account-hint
+   #'gptel--gh-github-token
+   (lambda (b token) (setf (gptel--gh-github-token b) token))
+   (or gptel-oauth-token-load-function 'gptel--gh-restore-token-from-file)
+   account-hint))
+
+(defun gptel--gh-save-token (account-hint token)
+  "Function that updates the GitHub OAuth token cache and calls the save function."
+  (gptel-oauth--save-token
+   #'gptel--gh-p
+   #'gptel--gh-account-hint
+   (lambda (b token) (setf (gptel--gh-github-token b) token))
+   (or gptel-oauth-token-save-function 'gptel--gh-save-token-to-file)
+   account-hint
+   token))
 
 ;; https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)
 (defun gptel--gh-uuid ()
@@ -267,28 +286,40 @@
       (setq hex (nconc hex (list (aref hex-chars (random 16))))))
     (apply #'string hex)))
 
-(defun gptel-gh-login ()
+(defun gptel--gh-generate-token-filename (account-hint)
+  "Generate token filename for GitHub backend with ACCOUNT-HINT."
+  (gptel-oauth--generate-token-filename
+   gptel-gh-github-token-file
+   #'gptel-oauth--validate-account-hint
+   account-hint))
+
+(defun gptel--gh-restore-token-from-file (account-hint)
+  "Restore GitHub token from file using ACCOUNT-HINT."
+  (gptel-oauth--restore-token-from-file #'gptel--gh-generate-token-filename account-hint))
+
+(defun gptel--gh-save-token-to-file (account-hint token)
+  "Save GitHub TOKEN to file using ACCOUNT-HINT."
+  (gptel-oauth--save-token-to-file #'gptel--gh-generate-token-filename account-hint token))
+
+(defun gptel-gh-login (account-hint)
   "Login to GitHub Copilot API.
 
 This will prompt you to authorize in a browser and store the token.
 
+ACCOUNT-HINT is used to provide a hint which account to login to.
+It must match an existing backend.
+
 In SSH sessions, the URL and code will be displayed for manual entry
 instead of attempting to open a browser automatically."
-  (interactive)
-  ;; Determine which GitHub backend to use
-  (let ((gh-backend
-         (cond
-          ;; If current backend is GitHub, use it
-          ((and (boundp 'gptel-backend)
-                gptel-backend
-                (gptel--gh-p gptel-backend))
-           gptel-backend)
-          ;; Otherwise, find any GitHub backend
-          ((cl-find-if (lambda (b) (gptel--gh-p b))
-                       (mapcar #'cdr gptel--known-backends)))
-          ;; No GitHub backend found
-          (t (user-error "No GitHub Copilot backend found.  \
-Please set one up with `gptel-make-gh-copilot' first")))))
+  (interactive (list (gptel-oauth--read-account-hint
+                      #'gptel--gh-p #'gptel--gh-account-hint
+                      "Choose GitHub account: "
+                      gptel--gh-default-username-placeholder
+                      "No GitHub copilot backends registered")))
+  (let ((gh-backends (gptel--gh-get-backends-by-account-hint account-hint)))
+    ;; It shall only be possible to login when there exists a corresponding backend
+    (if (= (length gh-backends) 0)
+        (user-error "No GitHub CoPilot backend found for account hint '%s'" account-hint))
     (pcase-let (((map :device_code :user_code :verification_uri)
                  (gptel--url-retrieve
                      "https://github.com/login/device/code"
@@ -296,25 +327,22 @@ Please set one up with `gptel-make-gh-copilot' first")))))
                    :headers gptel--gh-auth-common-headers
                    :data `( :client_id ,gptel--gh-client-id
                             :scope "read:user"))))
-      (gptel-oauth--device-auth-prompt user_code verification_uri)
-      ;; Use gh-backend for token storage
-      (let ((resp-body (gptel--url-retrieve
-                           "https://github.com/login/oauth/access_token"
-                         :method 'post
-                         :headers gptel--gh-auth-common-headers
-                         :data `( :client_id ,gptel--gh-client-id
-                                  :device_code ,device_code
-                                  :grant_type "urn:ietf:params:oauth:grant-type:device_code"))))
-        (thread-last
-            (plist-get resp-body :access_token)
-          (gptel-oauth--write-token gptel-gh-github-token-file)
-          (setf (gptel--gh-github-token gh-backend)))))
-    ;; Check gh-backend for success
-    (if (and (gptel--gh-github-token gh-backend)
-             (not (string-empty-p
-                   (gptel--gh-github-token gh-backend))))
-        (message "Successfully logged in to GitHub Copilot.")
-      (user-error "Error: You might not have access to GitHub Copilot Chat!"))))
+
+      (gptel-oauth--device-auth-prompt user_code verification_uri account-hint)
+      (let ((github-token
+             (plist-get
+              (gptel--url-retrieve
+                  "https://github.com/login/oauth/access_token"
+                :method 'post
+                :headers gptel--gh-auth-common-headers
+                :data `( :client_id ,gptel--gh-client-id
+                         :device_code ,device_code
+                         :grant_type "urn:ietf:params:oauth:grant-type:device_code"))
+              :access_token)))
+        (if (or (null github-token) (string-empty-p github-token))
+            (user-error "Error: You might not have access to GitHub Copilot Chat!"))
+        (message "Successfully logged in to GitHub Copilot")
+        (gptel--gh-save-token account-hint github-token)))))
 
 (defun gptel--gh-renew-token ()
   "Renew session token."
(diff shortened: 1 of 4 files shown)


marcus-lundgren:
@karthink - Do you have any input in how you imagine the workflow to be for this API regarding the GitHub token?

With the changes made just recently to present the login function, is it reasonable to have a workflow to perform that in order to get the token that can then be stored elsewhere?

karthink:
> @karthink - Do you have any input in how you imagine the workflow to be for this API regarding the GitHub token?
>
> With the changes made just recently to present the login function, is it reasonable to have a workflow to perform that in order to get the token that can then be stored elsewhere?

Sorry, I haven't looked at this PR yet.  I also don't use Copilot so let's hear from @kiennq (the author of gptel-gh) on the proposed changes.

marcus-lundgren:
@kiennq  - A friendly reminder about this PR :)

kiennq:
Sorry for the late reply. Can you add a default/example for `github-token-init`? At least it should be defaulted to the current workflow where the token is saved/restored from a file, else we're looking at reauth everytime gptel is reloaded

marcus-lundgren:
Thanks for your reply, @kiennq!

This is what I am currently using:

```emacs-lisp 
(use-package gptel
  :defer t
  :config (setq
           gptel-backend (gptel-make-gh-copilot "Copilot"
                           :github-token-init (lambda () (read-passwd "Copilot token: "))))
  )
```

I store my token in a password manager, so I use `read-passwd` in order to enter it when calling `gptel`. I will only be prompted once per Emacs session for this, which is a very reasonable workflow for me as I rarely restart Emacs. As it takes just a function, it is trivial to make it read from a file instead.

With the current implementation in this PR, the login function is not able to persistently store the newly acquired token. The new token is only stored in a variable and I take the it from the message log, which is not an ideal scenario and hence my request for feedback earlier. Another approach could be to introduce load and save functions that can be given as parameters to `gptel-make-gh-copilot`. It would then allow the user to choose how it should be stored and saved.

Before making the changes in gptel itself, I used the code below as a patch to control how the token was stored. This also has the behavior of only requesting the token once. As you may notice, the logic for the chat message token was kept, but this was before me realising that that is also a thing to keep secret.

```emacs-lisp
(use-package gptel
  :config
  (progn
    (defvar my/gh-token nil "Store the Copilot token for GPTel commands.")
    (defalias 'original-gptel--gh-save (symbol-function 'gptel--gh-save))
    (defalias 'original-gptel--gh-restore (symbol-function 'gptel--gh-restore))
    (defun gptel--gh-restore (file)
      (if (equal file gptel-gh-github-token-file)
          (if my/gh-token
              my/gh-token
            (progn 
              (setq my/gh-token
                    (let ((token (read-passwd "Copilot token: ")))
                      (if (string= "" token) nil token)))
              my/gh-token))
        (original-gptel--gh-restore file)))
    (defun gptel--gh-save (file obj)
      (message "New GH token: %s" obj)
      (if (equal file gptel-gh-github-token-file)
          (setq my/gh-token obj)
        (original-gptel--gh-save file obj)))))
```

marcus-lundgren:
@kiennq - Any thoughts? :)

marcus-lundgren:
@kiennq - I've now changed the implementation so that it defaults to use the file as before for the github-token.

What has been additionally added is the ability to create a custom save function for the github-token. This solves the question that I raised before, as it is now up to the one redefining the save function to decide how to handle an update.

A sample use-package snippet to override it:

```emacs-lisp
(use-package gptel
  :defer t
  :config
  (defvar my/gh-token nil "Store the Copilot token for GPTel commands.")
  (setq
   gptel-gh-github-token-load (lambda ()
                                (if my/gh-token
                                    my/gh-token
                                  (progn 
                                    (setq my/gh-token
                                          (let ((token (read-passwd "Copilot token: ")))
                                            (if (string= "" token) nil token)))
                                    my/gh-token)))
   gptel-gh-github-token-save (lambda (token)
                                (setq my/gh-token token))
   gptel-backend (gptel-make-gh-copilot "Copilot"))
  )
```

marcus-lundgren:
@kiennq - a friendly reminder concerning this. I think the current solution in this MR strikes a good balance between ease of use and choice of security. If you agree, then I can start working on changes to the README.

@karthink - do you have any thoughts regarding the latest proposal?

marcus-lundgren:
@kiennq, @karthink - another reminder.

karthink:
@marcus-lundgren just dropping a note to say this PR remains on my radar -- I just need a longer chunk of time to review it than other PRs since I'm not the author of the token management code.  I'll do a general code review but I trust @kiennq's judgment on the design.

marcus-lundgren:
@karthink - thank you for the status update! As you may have noticed, I am rebasing the branch continuously to fix any conflicts (as happened a few days ago).

gptel-gh.el
```diff
@@ -317,38 +302,25 @@ If your browser does not open automatically, browse to %s."
...
-    (let ((token (gptel--gh-restore gptel-gh-github-token-file)))
-      (if token
-          (setf (gptel--gh-github-token gptel-backend) token)
-        (gptel-gh-login))))
-
-  (when (null (gptel--gh-token gptel-backend))
-    ;; try to load token from `gptel-gh-token-file'
-    (setf (gptel--gh-token gptel-backend)
-          (gptel--gh-restore gptel-gh-token-file)))
+  (unless (funcall gptel-gh-github-token-load)
```
kiennq: This function can be called quite frequently. Can you change the default `token-load` function to cache github-token to a variable instead of reading from file everytime?

marcus-lundgren: Good catch! I introduced a general cache for the loading function. This will also mimic the behavior from when the token was stored in the backend.

gptel-gh.el
```diff
@@ -180,26 +180,28 @@
...
   "File where the GitHub token is stored."
   :type 'string
   :group 'gptel)
 
-(defcustom gptel-gh-token-file (expand-file-name ".cache/copilot-chat/token"
-                                                 user-emacs-directory)
-  "File where the chat token is cached."
-  :type 'string
-  :group 'gptel)
+(defvar gptel-gh-github-token-load 'gptel--gh-restore-from-file
```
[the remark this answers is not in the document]
marcus-lundgren: Fixed! When I investigated defcustom previously, I for some reason was unable to find the function type.

marcus-lundgren:
Updated config example with the builtin cache functionality:

```emacs-lisp
(use-package gptel
  :defer t
  :config
  (setq
   gptel-gh-github-token-load-function (lambda () (read-passwd "Copilot token: "))
   gptel-gh-github-token-save-function (lambda (token) (message "Token saved: %s" token))
   gptel-backend (gptel-make-gh-copilot "Copilot"))
  )

```

gptel-gh.el
```diff
@@ -287,28 +291,24 @@ in your local browser, enter the code, and authorize.  Press ENTER after authori
...
-        (gptel--gh-save gptel-gh-github-token-file)
-        (setf (gptel--gh-github-token gh-backend))))
-    ;; Check gh-backend for success
-    (if (and (gptel--gh-github-token gh-backend)
-             (not (string-empty-p
-                   (gptel--gh-github-token gh-backend))))
-        (message "Successfully logged in to GitHub Copilot.")
-      (user-error "Error: You might not have access to GitHub Copilot Chat!"))))
+             user_code verification_uri))
+    (browse-url verification_uri)
```
kiennq: The indentation looks off, can you double check?

kiennq:
I left one comment, else it LGTM

marcus-lundgren:
@kiennq, @karthink - I resolved the review comments you had after fixing them. If the workflow is that the reviewer should do that, then please let me know.

I've now also updated the README file with information about the customizable variables.

Some suitable information for the NEWS file can be found below. Do you want me to add it?

- Breaking change: Removed the customizable variable `gptel-gh-token-file`, as the chat token is no longer persistently stored.
- Added the customizable variables `gptel-gh-github-token-load-function` and `gptel-gh-github-token-save-function`. This enables the GitHub OAuth token to be loaded from and saved to arbitrary places.

karthink:
If I understand the changes correctly, you can no longer have more than one Github copilot backend defined, since the authentication is now shared across all gh backends?  Is this correct?

marcus-lundgren:
@karthink - Yes, you are correct, but I would argue that while it was technically true before, it wasn't properly implemented as such. Simply due to the fact that one could only persistently store one token and there was no official way to set the GitHub token manually.

Calling the login function will overwrite the token stored in the file:

https://github.com/karthink/gptel/blob/master/gptel-gh.el#L267

It would also fall back to the persistently stored token on failure to renew the chat token:

https://github.com/karthink/gptel/blob/master/gptel-gh.el#L287 followed by
https://github.com/karthink/gptel/blob/master/gptel-gh.el#L298

With the changes I've made as a basis, one could:

- Reintroduce the tokens to the backend to be used as a cache variable
- Add the load and save functions as parameters to the gptel-make-gh-copilot function and add them as references to the backend. This would enable several backends to use the same overloaded functions.

But I think that might be suitable for a follow up PR to make it proper.

Any thoughts, @kiennq? Do you agree with my understanding of the logic and that it is suitable for another PR?

kiennq:
> Any thoughts, @kiennq? Do you agree with my understanding of the logic and that it is suitable for another PR?

Although we only store one token to file, we can have multiple backends and each backend can ask for a github token by using the login function. So, we can still use multiple github accounts/backends.

gptel-gh.el
```diff
@@ -182,25 +199,25 @@
       (setq hex (nconc hex (list (aref hex-chars (random 16))))))
     (apply #'string hex)))
 
-(defun gptel--gh-restore (file)
-  "Restore saved object from FILE."
-  (when (file-exists-p file)
+(defun gptel--gh-restore-from-file ()
```
kiennq: Can you keep this function the same signature as before, i.e., taking a file param.
I'm working on a change to automatically fetch the model list and will use this to store the cached model list as well

marcus-lundgren:
> > Any thoughts, @kiennq? Do you agree with my understanding of the logic and that it is suitable for another PR?
> 
> Although we only store one token to file, we can have multiple backends and each backend can ask for a github token by using the login function. So, we can still use multiple github accounts/backends.

As @karthink identified in his comment, that will no longer work in the current implementation in this PR.

As the GitHub account used for a backend can change behind the scenes as described earlier, I would like to suggest adding proper support for token handling. I am thinking something along the lines of adding a parameter to the gptel-`make-gh-copilot` function which contains the desired GitHub username. That would allow the user to explicitly state which account to use, which in turn makes it possible to store the token in account specific files (e.g. github_token_karthink). Then we will avoid the following scenario that I fall into:

1. Have one backend for a personal account using free tier and one backend for a work account using a paid subscription
2. The currently stored token in the file is for the personal account as that was logged in last
3. We receive an error in the following logic, triggering the work account to read and make use of the personal account's token from the file:

```
It would also fall back to the persistently stored token on failure to renew the chat token:

https://github.com/karthink/gptel/blob/master/gptel-gh.el#L287 followed by
https://github.com/karthink/gptel/blob/master/gptel-gh.el#L298
``` 

4. The backend meant for the work account will now burn through the free tier's limited usage limit.

`gptel-gh-github-token-load-function` and `gptel-gh-github-token-save-function` should then take the GitHub username as a parameter, so that the implementation can deduce how to save/load the account specific token. With this, it would probably be best to replace the `gptel-gh-github-token-file` custom variable with a variable that points to a directory, as the filename will have the username in it. What do you think, @kiennq?

kiennq:
How about making the `gptel-gh-github-token-save/load-function` as part of the `gptel--gh` backend? Same with the chat token cache. That would allow the user to be able to create multiple backends that target different github copilot backends.

marcus-lundgren:
I am still determined to follow through with this PR, but family and other things has to be prioritized at the moment. I am still using this branch every day at work, so it will not become stale.

marcus-lundgren:
@karthink, @kiennq - I have now implemented my vision of proper account handling. The backends are now aware of which account they belong to by setting the github-username parameter. Not explicitly setting it will fall back to the old behavior, with the difference that all backends with the same intended username will have their github-token value updated at the same time. It is  therefore now deterministic which account will be used for each backend. A new login for a given username will update all of the related backends.

gptel-gh.el
```diff
@@ -411,6 +464,7 @@ for."
                   :stream stream
                   :request-params request-params
                   :curl-args curl-args
+                  :github-username github-username
```
kiennq: Can we use the `:name` parameter instead? I'm not sure if adding new parameter is better since we need unique `name` for each backend anyway.

kiennq (approved):

kiennq:
Approve with a suggestion of not adding `:github-username` parameter for the class. We have `:name` for each backend already.
Also, `gptel-gh-login` needs to be able to run interactively (i.e., the `github-username`) should be able to be inputted from mini buffer when called via`M-x`

gptel-gh.el
```diff
@@ -219,32 +284,26 @@
     (write-region (prin1-to-string obj) nil file nil :silent)
     obj))
 
-(defun gptel-gh-login ()
+(defun gptel-gh-login (github-username)
```
kiennq: This function is interactive, the `github-username` should be able to be inputted from mini-buffer when calling this interactively

gptel-gh.el
marcus-lundgren: We could, but then:

- We would not be able to have the current implementation's benefit of updated all backends that are connected to the same account. But if the officially sanctioned logic is that one GitHub account should only be connected to one backend, then I can remove that logic no matter which solution we choose.
- How should illegal characters for filenames be handled? The `:github-username` parameter enforces the allowed characters for GitHub usernames, which fits within the valid character range for filenames. The `:name` parameter may currently take any character, such as `/`, which is problematic. A solution would be to add the enforcement to that parameter as well, but would confusingly only be a requirement for the copilot backend.

Even though it is unfortunate, I feel that having this separate from the `:name` parameter is the least bad choice.

gptel-gh.el
marcus-lundgren: Good catch! I'll fix it.

kiennq: Btw, here is how I fixed it in my branch. Your changes are working fine for me though. https://github.com/kiennq/gptel/commit/8e0bb5f1523802959c6c3860d3665717551b68b9

kiennq (approved):

gptel-gh.el
```diff
@@ -230,32 +310,37 @@
...
-          ((cl-find-if (lambda (b) (gptel--gh-p b))
-                       (mapcar #'cdr gptel--known-backends)))
-          ;; No GitHub backend found
-          (t (user-error "No GitHub Copilot backend found.  \
-Please set one up with `gptel-make-gh-copilot' first"))))
+  (interactive (list (let ((known-usernames (gptel--gh-get-registered-usernames)))
+                       (cond
+                        ((= 0 (length known-usernames))
+                         (user-error "No GitHub copilot backends registered"))
+                        ((= 1 (length known-usernames))
```
kiennq: In this case if we only have default user, it would return the place holder and then we can't find that placeholder on the backend list

kiennq (changes requested):

marcus-lundgren:
@karthink, @kiennq - A friendly reminder regarding this PR :-)

karthink:
@marcus-lundgren I'm trying to understand the changes, and I don't follow how multiple Copilot backends are now supported.

There is only one `gptel--gh-token` variable, so if I switch between two Copilot backends the second one will error.  Am I missing something?

marcus-lundgren:
@karthink - Are you looking at the latest version? I have reverted from the global gptel--gh-token so that it is placed in each backend as before.

karthink:
> marcus-lundgren left a comment (karthink/gptel#1058)
>
> @karthink - Are you looking at the latest version? I have reverted from the global gptel--gh-token so that it is placed in each backend as before.

I think I was, my apologies.  I'll check again.

karthink:
I tried to test it as follows:

```elisp
(gptel-make-gh-copilot "Copilot")
```

- Run `M-x gptel-gh-login`
- Authorize in the browser
- After authorizing and pressing `RET` in Emacs, I got an error with this backtrace:

``` elisp
Debugger entered--Lisp error: (wrong-type-argument gptel--gh #s(gptel-gemini :name "Gemini" :host "generativelanguage.googleapis.com" :header nil :protocol "https" :stream t :endpoint "/v1beta/models" :key gptel-api-key-from-auth-source :models (gemini-pro-latest gemini-flash-latest gemini-flash-lite-latest) :url #f(lambda () [(endpoint "/v1beta/models") (protocol "https") (host "generativelanguage.googleapis.com") (stream t)] (let ((method (if (and stream gptel-use-curl gptel-stream) "streamGenerateContent" "generateContent"))) (format "%s://%s%s/%s:%s?key=%s" protocol host endpoint gptel-model method (gptel--get-api-key)))) :request-params nil :curl-args nil :coding-system nil))
  (signal wrong-type-argument (gptel--gh #s(gptel-gemini :name "Gemini" :host "generativelanguage.googleapis.com" :header nil :protocol "https" :stream t :endpoint "/v1beta/models" :key gptel-api-key-from-auth-source :models (gemini-pro-latest gemini-flash-latest gemini-flash-lite-latest) :url #f(lambda () [(endpoint "/v1beta/models") (protocol "https") (host "generativelanguage.googleapis.com") (stream t)] (let ((method (if ... "streamGenerateContent" "generateContent"))) (format "%s://%s%s/%s:%s?key=%s" protocol host endpoint gptel-model method (gptel--get-api-key)))) :request-params nil :curl-args nil :coding-system nil)))
  (or (let* ((cl-x cl-x)) (progn (and (memq (type-of cl-x) cl-struct-gptel--gh-tags) t))) (signal 'wrong-type-argument (list 'gptel--gh cl-x)))
  (let* ((cl-x gptel-backend)) (or (let* ((cl-x cl-x)) (progn (and (memq (type-of cl-x) cl-struct-gptel--gh-tags) t))) (signal 'wrong-type-argument (list 'gptel--gh cl-x))) (let* ((v cl-x)) (aset v 14 token)))
  (funcall gptel-gh-github-token-save-function github-username (let* ((cl-x gptel-backend)) (or (let* ((cl-x cl-x)) (progn (and (memq (type-of cl-x) cl-struct-gptel--gh-tags) t))) (signal 'wrong-type-argument (list 'gptel--gh cl-x))) (let* ((v cl-x)) (aset v 14 token))))
  (gptel--gh-save-github-token "" "ghu_XXXXXXXXXXXXXXXXXXXXXXXXXX")
  (let ((github-token (plist-get (gptel--url-retrieve "https://github.com/login/oauth/access_token" :method 'post :headers gptel--gh-auth-common-headers :data (cons ':client_id (cons gptel--gh-client-id (cons ... ...)))) :access_token))) (if (or (null github-token) (string-empty-p github-token)) (user-error "Error: You might not have access to GitHub Copilot Chat!")) (message "Successfully logged in to GitHub Copilot") (gptel--gh-save-github-token github-username github-token))
  (progn (gui-set-selection 'CLIPBOARD user_code) (let ((username-text (if (string= github-username "") "Login for the default account." (format "Login for '%s'." github-username)))) (if in-ssh-session (progn (message "GitHub Device Code: %s (copied to clipboard)" user_code) (read-from-minibuffer (format "%s Code %s is copied. Visit %s in your local browser, enter the code, and authorize.  Press ENTER after authorizing. " username-text user_code verification_uri))) (read-from-minibuffer (format "%s Your one-time code %s is copied. Press ENTER to open GitHub in your browser. If your browser does not open automatically, browse to %s." username-text user_code verification_uri)) (browse-url verification_uri) (read-from-minibuffer "Press ENTER after authorizing."))) (let ((github-token (plist-get (gptel--url-retrieve "https://github.com/login/oauth/access_token" :method 'post :headers gptel--gh-auth-common-headers :data (cons ':client_id (cons gptel--gh-client-id ...))) :access_token))) (if (or (null github-token) (string-empty-p github-token)) (user-error "Error: You might not have access to GitHub Copilot Chat!")) (message "Successfully logged in to GitHub Copilot") (gptel--gh-save-github-token github-username github-token)))
  (let ((device_code x1894) (user_code x1895) (verification_uri x1896)) (progn (gui-set-selection 'CLIPBOARD user_code) (let ((username-text (if (string= github-username "") "Login for the default account." (format "Login for '%s'." github-username)))) (if in-ssh-session (progn (message "GitHub Device Code: %s (copied to clipboard)" user_code) (read-from-minibuffer (format "%s Code %s is copied. Visit %s in your local browser, enter the code, and authorize.  Press ENTER after authorizing. " username-text user_code verification_uri))) (read-from-minibuffer (format "%s Your one-time code %s is copied. Press ENTER to open GitHub in your browser. If your browser does not open automatically, browse to %s." username-text user_code verification_uri)) (browse-url verification_uri) (read-from-minibuffer "Press ENTER after authorizing."))) (let ((github-token (plist-get (gptel--url-retrieve "https://github.com/login/oauth/access_token" :method 'post :headers gptel--gh-auth-common-headers :data (cons ... ...)) :access_token))) (if (or (null github-token) (string-empty-p github-token)) (user-error "Error: You might not have access to GitHub Copilot Chat!")) (message "Successfully logged in to GitHub Copilot") (gptel--gh-save-github-token github-username github-token))))
  (let* ((x1894 (map-elt val :device_code)) (x1895 (map-elt val :user_code)) (x1896 (map-elt val :verification_uri))) (let ((device_code x1894) (user_code x1895) (verification_uri x1896)) (progn (gui-set-selection 'CLIPBOARD user_code) (let ((username-text (if (string= github-username "") "Login for the default account." (format "Login for '%s'." github-username)))) (if in-ssh-session (progn (message "GitHub Device Code: %s (copied to clipboard)" user_code) (read-from-minibuffer (format "%s Code %s is copied. Visit %s in your local browser, enter the code, and authorize.  Press ENTER after authorizing. " username-text user_code verification_uri))) (read-from-minibuffer (format "%s Your one-time code %s is copied. Press ENTER to open GitHub in your browser. If your browser does not open automatically, browse to %s." username-text user_code verification_uri)) (browse-url verification_uri) (read-from-minibuffer "Press ENTER after authorizing."))) (let ((github-token (plist-get (gptel--url-retrieve "https://github.com/login/oauth/access_token" :method ... :headers gptel--gh-auth-common-headers :data ...) :access_token))) (if (or (null github-token) (string-empty-p github-token)) (user-error "Error: You might not have access to GitHub Copilot Chat!")) (message "Successfully logged in to GitHub Copilot") (gptel--gh-save-github-token github-username github-token)))))
  (progn (ignore (mapp val)) (let* ((x1894 (map-elt val :device_code)) (x1895 (map-elt val :user_code)) (x1896 (map-elt val :verification_uri))) (let ((device_code x1894) (user_code x1895) (verification_uri x1896)) (progn (gui-set-selection 'CLIPBOARD user_code) (let ((username-text (if ... "Login for the default account." ...))) (if in-ssh-session (progn (message "GitHub Device Code: %s (copied to clipboard)" user_code) (read-from-minibuffer ...)) (read-from-minibuffer (format "%s Your one-time code %s is copied. Press ENTER to open GitHub in your browser. If your browser does not open automatically, browse to %s." username-text user_code verification_uri)) (browse-url verification_uri) (read-from-minibuffer "Press ENTER after authorizing."))) (let ((github-token (plist-get ... :access_token))) (if (or (null github-token) (string-empty-p github-token)) (user-error "Error: You might not have access to GitHub Copilot Chat!")) (message "Successfully logged in to GitHub Copilot") (gptel--gh-save-github-token github-username github-token))))))
  (let* ((val (gptel--url-retrieve "https://github.com/login/device/code" :method 'post :headers gptel--gh-auth-common-headers :data (cons ':client_id (cons gptel--gh-client-id '(:scope "read:user")))))) (progn (ignore (mapp val)) (let* ((x1894 (map-elt val :device_code)) (x1895 (map-elt val :user_code)) (x1896 (map-elt val :verification_uri))) (let ((device_code x1894) (user_code x1895) (verification_uri x1896)) (progn (gui-set-selection 'CLIPBOARD user_code) (let ((username-text ...)) (if in-ssh-session (progn ... ...) (read-from-minibuffer ...) (browse-url verification_uri) (read-from-minibuffer "Press ENTER after authorizing."))) (let ((github-token ...)) (if (or ... ...) (user-error "Error: You might not have access to GitHub Copilot Chat!")) (message "Successfully logged in to GitHub Copilot") (gptel--gh-save-github-token github-username github-token)))))))
  (let ((gh-backends (gptel--gh-get-backends-by-username github-username)) (in-ssh-session (or (getenv "SSH_CLIENT") (getenv "SSH_CONNECTION") (getenv "SSH_TTY")))) (if (= (length gh-backends) 0) (user-error "No GitHub CoPilot backend found for username '%s'" github-username)) (let* ((val (gptel--url-retrieve "https://github.com/login/device/code" :method 'post :headers gptel--gh-auth-common-headers :data (cons ':client_id (cons gptel--gh-client-id '...))))) (progn (ignore (mapp val)) (let* ((x1894 (map-elt val :device_code)) (x1895 (map-elt val :user_code)) (x1896 (map-elt val :verification_uri))) (let ((device_code x1894) (user_code x1895) (verification_uri x1896)) (progn (gui-set-selection 'CLIPBOARD user_code) (let (...) (if in-ssh-session ... ... ... ...)) (let (...) (if ... ...) (message "Successfully logged in to GitHub Copilot") (gptel--gh-save-github-token github-username github-token))))))))
  (gptel-gh-login "")
  (funcall-interactively gptel-gh-login "")
  (command-execute gptel-gh-login record)
  (execute-extended-command nil "gptel-gh-login" "gptel-gh-login")
  (funcall-interactively execute-extended-command nil "gptel-gh-login" "gptel-gh-login")
  (command-execute execute-extended-command)
```

`gptel--gh-save-github-token` seems to be trying to use my active `gptel-backend` (Gemini) instead of the Copilot backend.

marcus-lundgren:
@karthink - Good catch! I've pushed a fix and I am now unable to replicate the issue.

marcus-lundgren:
I made an unfortunate rebase that was pushed. It seems like this PR was automatically closed due to that. Reopening it.

karthink:
I made a similar rebase error yesterday and accidentally closed a PR.   I
guess you're never experienced enough with git to not shoot yourself in the
foot now and then!

Anyway, I still mean to merge this PR, sorry for the continued delays.  The
last time I looked at it, I wondered if gptel could use a more secure token
storage option out of the box *without* requiring the user to set up any
keys or passwords.  I looked at plstore and auth-source, both part of
Emacs.  The former requires a gpg key or requires the user to type in a
password.  The latter can integrate with Gnome and KDE keyrings and the
equivalent on MacOS, but won't work on Windows.

marcus-lundgren:
@karthink - Thank you for the update! It is always great to be humbled once in a while after having an accident like the one that we both seem to have shared. (Thank goodness for `git reflog show`!)

That sounds interesting. I may look into doing something with plstore instead of copying the secret manually from a password manager, as it has support for storing secrets (plstore-put). Will share a suitable snippet if I get around to it.

karthink:
Haven't lost track of this PR!  It's fast approaching the title of longest standing gptel PR that will eventually be merged.

Incidentally, see discussion of at-rest oauth token encryption in #1349.

marcus-lundgren:
> Haven't lost track of this PR! It's fast approaching the title of longest standing gptel PR that will eventually be merged.

Thanks for the update, @karthink! I will wear it as a badge of honor if that becomes the case :-)

> Incidentally, see discussion of at-rest oauth token encryption in #1349.

Yeah, I saw it whilst performing my daily browsing of issues and PRs in this repo. It is always nice when someone is interested in a thing that one has created (which I hope is a feeling you get for having created this wonderful package).

karthink:
@marcus-lundgren You'll see a merge conflict when you rebase this PR, but hopefully it makes things simpler in the end.

I moved some of the token persistence code into a dedicated `gptel-oauth.el`.  Your fix should still apply, just split across the two files.

This PR has two functions:

1. Provide per-gh-backend token storage
2. Make the token storage method customizable

1 can continue to live in this file.  2 needs to be moved to `gptel-oauth` and made backend-agnostic -- the token storage function/method should apply to all backends that use `gptel-oauth`.

More generally I'm looking into encrypting access tokens on disk with `plstore` or `auth-source`, the problem remains that there is no encryption method available in Emacs across OS platforms that does not bother the user with password entry or encryption key setup for something (the access token) they have no idea about.  Still trying to find a "just works" access token encryption solution.

marcus-lundgren:
Thank you for the update, @karthink.

I saw that it was merged and I have been working for some time on rebasing my changes to latest master this evening. In order to reduce the complexity of it, I've squashed some of the commits together. I don't have much time at the moment so it may take some time, but it will be done in the end.

I will first start with getting the branch in a state where the external functionality is the same. I will then proceed with making the custom functions to be placed in the new file. Will let you know as soon I have a need to discuss some implementation detail.

marcus-lundgren:
@karthink - The rebase was easier than I thought and this PR works as before.

There is a design decision however that will need to be taken. In the gptel-gh implementation, the storing of the token is based on the assigned github-username, which is enforced to only contain characters that are valid for filenames. The functions concerning where to then read/store tokens use the github-username in order to provide meaningful context to the custom function or to give a sane value to calculated the filepath in the fallback scenario based on this value.

In the gptel-openai-oauth implementation, there is no apparent existing value that matches github-username. The name of the backend could be used, but for legacy reasons a user may have used characters that are not then valid to use in filenames.

A solution to this is to add an equivalent of github-username to gptel-make-openai-oauth. By having a quick glance, it seems like an email address is used as a username. The @ symbol seems to be valid according, so it should hopefully be a suitable candidate to use.

I may just make a placeholder implementation to get things going, but it would be great if you could add your thoughts before I start doing the work.

marcus-lundgren:
I am working on finalising the implementation. The kids birthdays et al makes progress a bit slower.

karthink:
@marcus-lundgren I'm taking another crack at merging this now.  Can you rebase this on master?  If you're busy I can try the rebase too.

marcus-lundgren:
Hi, @karthink!

Things are now calmer and I started progressing with the implementation again.

I wanted to introduce some sort of validation of the token that is read/written. It was implemented yesterday, but I haven't hade the chance to test it. I do not have the ability to try out the OpenAI variant, but hopefully you or someone else is able to.

In the mean time, it would be nice to receive feedback on the design decisions I've made.

I will rebase on master later today and ping you afterwards.

marcus-lundgren:
@karthink - Rebase complete and some additional changes have been pushed. I don't remember when I last tried out the logic, but it wasn't too many commits ago. But it should be ready for a review.

marcus-lundgren:
@karthink - I fixed the latest merge conflicts, but haven't had the time to test it for regressions. The merge seemed clean and I do not get any parsing errors of the lisp code.

marcus-lundgren:
@karthink - The code is yet again rebased. I also fixed some mistakes that was introduced since I last tested the code regarding the custom function calling.

marcus-lundgren:
@karthink - I fixed another minor thing and rebased yet again. Do let me know if there is anything else I can do. The PR shall be considered to be ready for review.

marcus-lundgren:
@karthink - What is your plan for this PR? Is there anything else I can do?

marcus-lundgren:
@karthink - A gentle bump regarding this PR.

marcus-lundgren:
@karthink - Another gentle bump :-)

PR #1063, opened 2025-09-06
gptel-gemini: Update model information

benthamite:
This PR supplies missing `output-cost` and `cutoff-date` information for a few model; trims docstrings to ensure they fit the space allotted; and removes an unnecessary URL.


karthink:
@benthamite Thank you!

PR #1070, opened 2025-09-13
gptel-gh: add grok-code-fast-1 model

DiogoDoreto:
Reuse properties defined in gptel-openai-extras but update context-window to 128, as returned by the Copilot Models API


karthink:
@DiogoDoreto Thanks for the PR!

PR #1075, opened 2025-09-19
support api.business.githubcopilot.com

CsBigDataHub:
@karthink thank you

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,34 @@
 
 * 0.9.9.3-pre
 
+** Breaking changes
+
+- The models =gemini-1.5-flash-8b=, =gemini-1.5-flash=,
+  =gemini-1.5-pro-latest=, =gemini-2.0-flash-thinking-exp-01-21=,
+  =gemini-2.0-flash-lite-preview-02-05=,
+  =gemini-2.5-flash-lite-preview-06-17=,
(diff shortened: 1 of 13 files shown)


ChetanKoneru:
Hello @karthink , Any thoughts on this PR?

`api.githubcopilot.com` stopped working for me the day I created this PR, hence the PR, I don't think I have a unique use-case here. I think giving users option to customize the endpoint will be helpful.

Just realized that I need to update README, LMK. 

As always thank you.

karthink:
@CsBigDataHub Thanks for the PR.  I don't think this needs to be a `defcustom` -- I suspect this is going to coderot in a few months or an year.  We are essentially swapping one host for another, so it should be enough to add instructions in the README as a subsection of the github-copilot section.

If you add those instructions to the README, I can merge that.

ChetanKoneru:
Hello @karthink,

Does this look good?

=========================

Github copilot is configured to use public end-point `api.githubcopilot.com`.
However private endpoints can be configure like so -

```emacs-lisp
(gptel-make-gh-copilot "Copilot"
        :host "api.business.githubcopilot.com")
```
=========================

karthink:
@CsBigDataHub, that looks good to me.

ChetanKoneru:
closing in favor of https://github.com/karthink/gptel/pull/1116

PR #1084, opened 2025-10-01
add kagi api key to request headers

mzacho:
The key needs to be passed on from `gptel-make-kagi`.


karthink:
Thanks for the PR.  I don't understand this change though.  `gptel--get-api-key` looks up the key corresponding to the active gptel backend, so it should not be necessary to include the key here.

mzacho:
You're right, apologies, my problem was upstream (no more credits), mistakenly thought this was the issue, closing.

PR #1087, opened 2025-10-04
Fix attachment link resolution with org-attach ID-based storage

### LINKED ISSUE #1073: Media tracking implementation not compatible with customized org-attach-id-dir

### Bug Description

I believe that gptel-send does not correctly locate media files referenced by org-mode "attachment" links when using org-attach's "ID based" storage support. Instead, gptel will gather an incorrect path to the image file, determine that the file is not readable, and then fail to encode and send it to the LLM.

I spent some time debugging this and have a fix that works for me, but I am not sure if it is appropriate.

First, in `gptel--parse-media-links`, make sure to use `org-expand` if the org entity is an attachment. Without this, gptel will try to open a relative file path that does not exist.

```lisp
        (when-let* ((link (org-element-context))
                    ((gptel-org--link-standalone-p link))
                    (raw-link (org-element-property :raw-link link))
                    (raw-path (org-element-property :path link))
                    (type (org-element-property :type link))
                    (path (if (string= type "attachment")
                              (org-attach-expand raw-path)
                            raw-path))
```

Second, in `gptel--realize-query`, set the current buffer appropriately before calling `gptel--parse-buffer`. Without this, `org-attach-expand` will not function correctly.

```lisp
          (setq full-prompt (with-current-buffer (plist-get info :buffer)
                              (gptel--parse-buffer ;prompt from buffer or explicitly supplied
                               gptel-backend (and gptel--num-messages-to-send
                                                  (* 2 gptel--num-messages-to-send)))))
```

If this issue report and proposed resolution passes your sniff test, I can put together a pull-request.

### Backend

None

### Steps to Reproduce

1. Customize org-attach-id-dir to some random location on your machine, not in the same working directory as your chat buffer
2. Attach an image file and link to it
3. Enable media tracking
4. Try to send it, asking "What is in this image?"

Without workaround, llm will say that it cannot see images. With workaround, it will describe image content.

Alternatively, you can set gptel-expert-commands to t and use the JSON inspect query support;
1. Perform steps 1-3 above
2. Inspect query JSON in the gptel transient

Without workaround, JSON payload will reflect that the attachment link was not detected as media. You'll just see the link text in the preceding :text node. With workaround, the media will be represented as an encoded entry.

### Additional Context

Reproduced on 2 machines
* NixOS, Emacs 30.1
* Windows 11, Emacs 29.something IIRC

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

jwiegley:
## Summary

Fixes #1073

This PR fixes an issue where gptel does not correctly locate media files referenced by org-mode "attachment" links when using org-attach's ID-based storage support.

### 1. `gptel-org.el` (gptel--parse-media-links)
- Use `org-attach-expand` to properly resolve attachment-type links to their actual file paths
- This is necessary when org-attach uses ID-based storage where attachment directories are not relative to the current file
- Added `declare-function` for `org-attach-expand`

### 2. `gptel-request.el` (gptel--realize-query)
- Call `gptel--parse-buffer` in the original buffer context rather than the temp buffer
- This ensures `org-attach-expand` has access to Org properties (ID, ATTACH_DIR) needed to resolve attachment paths
- These properties are not copied to the temp buffer, so the context switch is necessary

## Testing

- Backward compatible: file-type links continue to work as before
- Attachment-type links now properly expand to absolute paths when using ID-based storage
- Existing test suite validates media link parsing functionality

## Test plan

1. Customize `org-attach-id-dir` to a location outside the working directory
2. Create an Org file with a heading that has an ID property
3. Attach an image file to the heading using `org-attach`
4. Link to the attachment using `[[attachment:filename.png]]`
5. Enable media tracking in gptel
6. Send a query asking about the image
7. Verify the image is correctly encoded and sent to the LLM

Without this fix, the LLM would not receive the image. With the fix, the image is properly encoded and included in the request.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -41,6 +41,7 @@
 (defvar org-link-angle-re)
 (defvar org-link-bracket-re)
 (declare-function mailcap-file-name-to-mime-type "mailcap")
+(declare-function org-attach-expand "org-attach")
 (declare-function gptel--model-capable-p "gptel")
 (declare-function gptel--model-mime-capable-p "gptel")
 (declare-function gptel--model-name "gptel")
@@ -365,8 +366,11 @@ for inclusion into the user prompt for the gptel request."
         (when-let* ((link (org-element-context))
                     ((gptel-org--link-standalone-p link))
                     (raw-link (org-element-property :raw-link link))
-                    (path (org-element-property :path link))
+                    (raw-path (org-element-property :path link))
                     (type (org-element-property :type link))
+                    (path (if (string= type "attachment")
+                              (org-attach-expand raw-path)
+                            raw-path))
                     ;; FIXME This is not a good place to check for url capability!
                     ((member type `("attachment" "file"
                                     ,@(and (gptel--model-capable-p 'url)
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -2033,9 +2033,12 @@ Initiate the request when done."
           (save-excursion (goto-char (point-min))
                           (gptel--parse-list-and-insert (cdr directive))))
         (goto-char (point-max))
-        (setq full-prompt (gptel--parse-buffer ;prompt from buffer or explicitly supplied
-                           gptel-backend (and gptel--num-messages-to-send
-                                              (* 2 gptel--num-messages-to-send))))
+        ;; Parse buffer in original buffer context so org-attach-expand can access
+        ;; Org properties (ID, ATTACH_DIR) not present in the temp copy buffer
+        (setq full-prompt (with-current-buffer (plist-get info :buffer)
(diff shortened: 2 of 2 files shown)


jwiegley:
I tested this in the debugger by hand with a chat file as described by the original reporter, and indeed without these changes it was not able to see my attached image, but with the changes it saw it correctly and added it to the prompt to be sent.

karthink:
This looks good, I'll take a closer look soon.  Thanks!

0WD0:
Friendly ping — this looks useful. 
@karthink Could you take a look when convenient? Thanks!

Issue #370, opened 2024-09-03
magit COMMIT_EDITMSG buffer interpreted as Prose editor context


nfedyashev:
I've tried calling gptel in the context of magit while writing commit message

```
# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
```

`(print major-mode)` in such buffer outputs `text-mode`

`You are a prose editor. Rewrite the following text to be more professional`
I suspect that's not the prefilled context that a typical user might want(no disrespect to all Prose editors Emacs users).

LLM as git commit message (re)write tool seems to be pretty popular topic accordingly to Google search results. Perhaps it is worth improving this workflow for end users.

My suggestion - instead of just checking major-mode name, prior to that check whether `buffer-file-name` contains `COMMIT_EDITMSG` string and in this case, set a special context instead of the default one for text-mode(Prose editor).

Perhaps something like "You are git expert. What you write will be passed to git commit -m \"[message]\". Rewrite the following message."

karthink:
The refactoring interface needs a fair bit of work instead of micro-fixes, unfortunately.  Let's leave this issue open for now while I figure out how the redesigned interface should work.

karthink:
The refactoring interface has been redesigned from scratch (the workflow change should be apparent when you use it).  

> My suggestion - instead of just checking major-mode name, prior to that check whether buffer-file-name contains COMMIT_EDITMSG string and in this case, set a special context instead of the default one for text-mode(Prose editor).

For the issue of specialized refactor messages like this one, I've added a user option (`gptel-rewrite-directives-hook`) to which you can add functions.  Each function in the hook is called until one returns a (non-nil) string, the default refactoring instruction.  The function can use any logic to prepare the instruction.  In this case, you could do something like:

```elisp
(defun gptel-rewrite-commit-message ()
  (when (and (string-match-p "COMMIT_EDITMSG" (buffer-name))
             (derived-mode-p 'text-mode))
    "You are a git expert.  What you write will be passed to git commit -m \"[message].  Rewrite the following message."))

(add-hook 'gptel-rewrite-directives-hook #'gptel-rewrite-commit-message)
```

I can think about including messages for some common special cases (like this one) to the default value of this hook in gptel, but it'll help to get some feedback on how this system works first.

nfedyashev:
This one sounds like perhaps it can be closed :+1: 
Thank you

karthink:
[...]

PR #1095, opened 2025-10-09
gptel-rewrite: Fix directive hook evaluation location

aagit:
The gptel-rewrite-directive-hook was being evaluated in the wrong buffer context, causing (buffer-name) to return "gptel-prompt" instead of the actual buffer name.

The gptel-rewrite menu correctly displayed the prompt generated by the hook, making the issue not apparent during normal usage.

This was discovered while testing gptel-rewrite-ripgrep-edit, where the generic hooks fail to handle the ripgrep-edit format, making the incorrect hook evaluation apparent.

Fixes: 89d1b4768d9a ("gptel: Switch to prompt-buffer centric parsing")

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -708,7 +708,9 @@ generated from functions."
                              "\n\n" (car prompt))))
     (prog1 (gptel-request prompt
              :dry-run dry-run
-             :system gptel--rewrite-directive
+             :system (if (functionp gptel--rewrite-directive)
+                         (funcall gptel--rewrite-directive)
+                       gptel--rewrite-directive)
              :stream gptel-stream
              :context
              (let ((ov (or (cdr-safe (get-char-property-and-overlay (point) 'gptel-rewrite))


karthink:
@aagit I think you've sent me emails with a couple of patches about a similar issue that I haven't got to yet.

Do you want me to ignore the emails and look at your PRs, or do the opposite?

aagit:
You can ignore the email and only look at the PR. I did the PR after a few days because I wasn't sure if the git-send-email went to the gmail spam folder.

aagit:
Hello,

I probably didn't flag this properly but this is a regression fix so it's technically higher priority than other optimizations. Without fixing this bug ripgrep-edit.el doesn't work, but it's not just ripgrep-edit.el that doesn't work: all other per-file prompts also are currently non functional including the default "programming" one, it's just more apparent with ripgrep-edit.

karthink:
@aagit Two notes:

1. The right way to do this would be to replace `gptel--rewrite-directive` with `(gptel--parse-directive gptel--rewrite-directive 'raw)` instead of checking if it's a function here.  (The directive can be more complex than a single function, see `gptel--parse-directive`)

2. The directive is expanded in the prompt construction buffer intentionally.  This way presets that modify the rewrite directive will be dealing with the object they expect.  For example:

``` emacs-lisp
(gptel-make-preset 'rewrite-extra
  :system (concat (funcall gptel--system-message)
                  "\nExtra rewrite instructions"))
```

If you now use `@rewrite-extra` in the rewrite message (entered at the minibuffer), it will cause an error when the preset is applied in the `*gptel-prompt*` buffer because `gptel--system-message` is not a function, `gptel--rewrite-directive` has already been evaluated to a string.

This is a niche case, but it gets in the way of building a clean mental model of what gptel is doing.  On balance, it might be worth the trade-off -- I'll think about it for a bit longer before deciding if I should merge this.

aagit:
Sorry, I didn't notice I could select oneshot presets with @preset-name in the prompt.

Another way I thought of was to switch back to the orig buffer during the callback execution... no idea if it's feasible but I though I should mention it just in case.

karthink:
After some more thinking, it makes sense for the system message to always be parsed (eval'd if it's a function) in the original buffer -- if it's a string or list, it doesn't matter where it is parsed.  And if it's a function, it needs to have access to the local state (the current buffer, in your case).

Applying presets via `gptel-with-preset` or the transient menu will continue to work fine.  The only thing this breaks is oneshot presets where the user assumes the original system message (before the preset is applied) will be a function because they set it to a function.  This is a special case that I need to avoid or at least document.

If I make this change I'll do it in `gptel-request`, since this problem is not unique to `gptel-rewrite`.

karthink:
I've had more time to ponder this behavior now, and I think the correct thing to do is to parse the directive at the time of request, in the context in which `gptel-request` was invoked.  So I'll make this change to `gptel-request` soon.

karthink:
It's just a little tricky to do correctly since the directive can take many forms -- a string, list, or a function that can return a string or a list.  I'll get to it when I can.

aagit:
Hello, what's the status of the fix? I think the gptel-rewrite-directives-hook working again is more important than the @preset feature I didn't know about. Perhaps you shall consider dropping the @preset feature if it made things too complicated that hard to maintain. Overall I don't think mixing metadata that has side effect with the prompt is particularly clean interface. Selecting the preset with ALT+RET+@ may actually be faster with fewer keypress and less prone to a typo and then it's also applicable per buffer or globally, not just oneshot.

karthink:
The status of the fix is that I've tried twice to move the directive expansion to the prompt buffer in `gptel-request`, and failed both times because of subtle issues.  As I expected, it's been surprisingly tricky to get it right without edge cases.

The `@preset` issue is not a concern for now.

aagit:
Thanks for the update!

I may have misunderstood, I thought the cons of evaluating the gptel--rewrite-directive before switching buffer was purely that @rewrite-extra couldn't be used in the minibuffer anymore, which again looks like a "dubious" feature as it's slower and more error prone to type than the alternative.

aagit:
Hi @karthink,

Thanks for the thoughtful responses and for taking the time to consider this carefully.

I wanted to share that I’ve found a working workaround for the longstanding regression in `gptel-rewrite-directive-hook` usage:  
https://gitlab.com/aarcange/ripgrep-edit/-/commit/8e3b82fd9dc3a368e496e2f0a4eb5a44e156347e

With this fix applied, `razor-coding` works out of the box with upstream `gptel`, and I no longer need to ask users of rg-edit to apply custom patches to gptel or use my gptel tree. However, since only `ripgrep-edit` has this workaround applied, all other hooks that rely on `gptel-rewrite-directive-hook` remain broken.

I’ve been using the fix from my PR for over a month as an heavy user of `gptel`, and it’s proven stable with zero issues.

Also worth noting: as you hinted I verified that `@rewrite-extra` works perfectly when used in the rewrite prompt.

I suspect many users have configured `gptel-rewrite-directive-hook` under the assumption it’s working, but it silently doing nothing. This can be misleading and hard to debug.

Thanks again for your work on `gptel`, it's an incredibly useful tool!

karthink:
@aagit wait, I didn't know that `gptel-rewrite-directives-hook` is non-functional.  This is different from the above bug report, that the directive expansion happens in the wrong buffer:

> Due to a regression introduced in gptel commit
> 89d1b4768d9ab1d2624cb68d826f2134cb4c067e, the
> gptel-rewrite-directives-hook dynamic prompt appears functional (TAB
> after gptel-rewrite shows the expected prompt), but the generic prompt
> is silently used instead. This can be confirmed using Inspect (`I` key
> with gptel-expert-commands enabled).

I can't reproduce this behavior.  I did the following:

``` emacs-lisp
(add-hook 'gptel-rewrite-directives-hook
          (lambda ()
            (when (derived-mode-p 'emacs-lisp-mode)
              "Replacement for rewrite directive here.")))
```

then issued a rewrite in an emacs-lisp buffer and checked the dry-run output.  The rewrite directive is replaced as expected.

aagit:
> @aagit wait, I didn't know that `gptel-rewrite-directives-hook` is non-functional. This is different from the above bug report, that the directive expansion happens in the wrong buffer:
> 
> > Due to a regression introduced in gptel commit
> > [89d1b47](https://github.com/karthink/gptel/commit/89d1b4768d9ab1d2624cb68d826f2134cb4c067e), the
> > gptel-rewrite-directives-hook dynamic prompt appears functional (TAB
> > after gptel-rewrite shows the expected prompt), but the generic prompt
> > is silently used instead. This can be confirmed using Inspect (`I` key
> > with gptel-expert-commands enabled).
> 
> I can't reproduce this behavior. I did the following:
> 
> ```emacs-lisp
> (add-hook 'gptel-rewrite-directives-hook
>           (lambda ()
>             (when (derived-mode-p 'emacs-lisp-mode)
>               "Replacement for rewrite directive here.")))
> ```
> 
> then issued a rewrite in an emacs-lisp buffer and checked the dry-run output. The rewrite directive is replaced as expected.

I'm sorry I thought this was already clear, I guess I should have been more explicit that gptel-rewrite-directives-hook cannot be used at all, it's broken and regressed in commit 89d1b4768d9ab1d2624cb68d826f2134cb4c067e as reported in the Fixes header of this PR.

You can reproduce with your example here:

https://github.com/karthink/gptel/issues/370#issuecomment-2332503104

Even if this was intentional to break the old example, the fact that TAB shows the prompt based on the buffer-name, but what goes on the wire is something else (verify it with Inspect what goes on the wire) is a bug.

So at the moment I have to workaround it by rewriting internal variables or I would need to ask the ripgrep-edit users reporting that it doesn't work, to switch to my gptel tree until this is fixed.

aagit:
More details: to reproduce it's sufficient to use buffer-name in your hook. The whole point of the fix is to run the hook *before* changing buffer name... so it is evaluated in TAB in the same buffer as Inspect/send.

I don't actually know why it even works with derived-mode-p, as it also should be evaluated in the wrong buffer... but I can't check modes, I use different directives for different filename extensions, even when they have identical mode. In the case of rg-edit.el at the moment there's no mode to check at all... I just trigger prog-mode, ripgrep-edit will linearize anything: a mix of bash, python, rust, C can land in the same buffer, it's up to the AI to figure things out.

This is the disabled gptel hook:

https://gitlab.com/aarcange/ripgrep-edit/-/blob/main/emacs/rg-edit.el?ref_type=heads#L50
https://gitlab.com/aarcange/ripgrep-edit/-/blob/main/emacs/rg-edit.el?ref_type=heads#L173

BTW, I used buffer-file-name in the hook, not sure if I should have used buffer-name, I don't think it makes any difference, but buffer-name as in your example with COMMIT_EDITMSG also doesn't work, it's run in the wrong buffer. 

ripgrep-edit buffers end with .rg-edit.

karthink:
Please update and test, the `gptel-rewrite-directive-hook` should now be evaluated in the request buffer, and not in a temporary buffer.

aagit:
> Please update and test, the `gptel-rewrite-directive-hook` should now be evaluated in the request buffer, and not in a temporary buffer.

Perfect, the fix works.

Is there now a way to runtime detect when a version is fixed? So I can disable the workaround based on that.

The fact this is so subtle and TAB showed the right prompt, makes me wonder users will try rg-edit and think the LLM just corrupts the rg-edit format when in fact with the system prompt it works pretty well. It's not like 100% perfect but no LLM output can be. The whole point is that rg-edit speeds things up so much that one can try 10-100 times instead of waiting 10 minutes for some agent to run grep and read snippets and find out the directive didn't work. It's like having x10 GPUs depending on the use case and accuracy goes up compared to doing each case individually as it wants to see and it all in the same kv input. It's not just about not having to run gptel-rewrite multiple times in different files.

I hope there will be a way standardize the rg-edit format and fine tune the mainstream LLM for it to make it even less likely it eats separators, which is the main annoyance one can run into.

If it eats the separators normally it's with smaller models and with lots of context (above 32k). Either removing context with a @profile-noctx (if context isn't too relevant)  or starting the gptel-rewrite directive with "Rewrite: do not delete the separators. real directive here" is enough most of the time. If even then, it still eats separators usually it means it can't be done and some code to complete the directive is missing in the kv input to be rewritten.

As opposed without this gptel fix (or the workaround) it just doesn't work...

karthink:
> Is there now a way to runtime detect when a version is fixed? So I can disable the workaround based on that.

The only way I can think of is based on gptel's version, which is not a runtime check.  I can add a `gptel-version` defconst, but it won't change between releases either.

> I hope there will be a way standardize the rg-edit format and fine tune the mainstream LLM for it to make it even less likely it eats separators, which is the main annoyance one can run into.

`gptel-request` takes a `:schema` argument that you can use to force LLMs to generate structured JSON output.  This is what I use when I want it to do something like your rg-edit package does.  `:schema` works with all the major LLM APIs as long as you don't also include tools in the request, and gptel's (incomplete) manual has [a detailed explanation](https://gptel.org/manual.html#output-schema) of how to use it.

aagit:
> The only way I can think of is based on gptel's version, which is not a runtime check. I can add a `gptel-version` defconst, but it won't change between releases either.

How would a not runtime check work? Like scanning the path or something? That looks a complication.

I don't get why a gptel-version wouldn't change between releases?

I think the workaround works well enough that I can keep it for a while and drop it eventually.

> 
> > I hope there will be a way standardize the rg-edit format and fine tune the mainstream LLM for it to make it even less likely it eats separators, which is the main annoyance one can run into.
> 
> `gptel-request` takes a `:schema` argument that you can use to force LLMs to generate structured JSON output. This is what I use when I want it to do something like your rg-edit package does. `:schema` works with all the major LLM APIs as long as you don't also include tools in the request, and gptel's (incomplete) manual has [a detailed explanation](https://gptel.org/manual.html#output-schema) of how to use it.

Say I need to remove "param" in a function signature because the field is not in the second parameter anymore and now it's reachable as args.param with args already passed to the function and I've 10 files using it.

What I'd do without emacs and without gptel, is rg-edit the function name, copy and paste the file content into any LLM interface including web, give the prompt at the top with the requested change, get the rewritten rg-edit format back and overwrite the whole buffer and close the file, done.

How do I do that with :schema? Do I start with a ripgrep, and then what?

karthink:
> How would a not runtime check work? Like scanning the path or something? That
> looks a complication.
> 
> I don't get why a gptel-version wouldn't change between releases?

I meant that it won't change between two releases.  The fix for this issue is in a commit in between gptel 0.9.9.3 and 0.9.9.5, and `gptel-version` will only change to 0.9.9.5 after that version is released.

> Say I need to remove "param" in a function signature because the field is not in
> the second parameter anymore and now it's reachable as args.param with args
> already passed to the function and I've 10 files using it.
> 
> What I'd do without emacs and without gptel, is rg-edit the function name, copy
> and paste the file content into any LLM interface including web, give the prompt
> at the top with the requested change, get the rewritten rg-edit format back and
> overwrite the whole buffer and close the file, done.
> 
> How do I do that with :schema? Do I start with a ripgrep, and then what?

I don't know the internal details of how rg-edit works, I've only looked at the demo videos on the project page that use gptel-rewrite.  But I imagine the way it would work is
1. The LLM input would be the same rg output text/buffer that you currently use.
2. The LLM output would be in a schema that makes it return an array of replacements (before/after strings).
3. You would parse the array with `json-read` and apply each replacement in turn.  This way you don't have to worry about the LLM preserving the separators/delimiters etc.

If the ordering of the array is stable, you can just loop through them and apply it without additional validation to find the right text to replace.  If it's not stable, you can include an index field in the schema that should match up with the input ordering, and use that.

aagit:
> I meant that it won't change between two releases. The fix for this issue is in a commit in between gptel 0.9.9.3 and 0.9.9.5, and `gptel-version` will only change to 0.9.9.5 after that version is released.
> 

Ah yes, no problem that's the way it'd be intended to work. So it's up to you if you want to deal with the extra work of updating the versioning.

> I don't know the internal details of how rg-edit works, I've only looked at the demo videos on the project page that use gptel-rewrite. But I imagine the way it would work is
> 
>     1. The LLM input would be the same rg output text/buffer that you currently use.
> 
>     2. The LLM output would be in a schema that makes it return an array of replacements (before/after strings).
> 
>     3. You would parse the array with `json-read` and apply each replacement in turn.  This way you don't have to worry about the LLM preserving the separators/delimiters etc.
> 
> 
> If the ordering of the array is stable, you can just loop through them and apply it without additional validation to find the right text to replace. If it's not stable, you can include an index field in the schema that should match up with the input ordering, and use that.

So I imagine the rg output would be encoded as a list of json strings to be sent to the LLM. That sounds like something that should be tested just to see what happens.

rg-edit is opening a virtual window in multiple snippet of multiple files, it's an hard window but you (or the LLM) is free to grow or shrink the code inside the window any way it wants or it can drop a snippet by deleting a separator and then there's not enough info left to guess where to write the remaining ones and which one was deleted. I'm not comparing code as if I was applying a patch, that would be possible too, but then you need to handle the conflicts too.

So if you get back 1 less entry in the json list in output than you gave in input, it's back to square one. That entry can't be filled as a empty string safely, I mean rg-edit could already write even if there's a missing snippet in the first place but it signals something is off so the chances this is going great even doing code comparisons handling conflicts aren't high. It's preferable to restart from scratch and get a clean refactoring in my view.

One thing I do often is to edit the rg-edit in the buffer before the gptel rewrite or after it (if there was just a small glitch that clearly break the formats, once for example it changed the file prefix from -- to // that's very easy to see in the diff and fix by hand to pass the format checks).

rg-edit doesn't need a date or a number, json facilitates the parsing for frontends that need to deal with structured data. If you know the next token must be a certain thing you improve accuracy too with json beam search.

Here there's no constraint that can be applied to improve accuracy of the softmax pick in output from the json parser, not even on when to end a sentence. So the challenge is to give a prompt that makes it stay constrained and not treat each snippnet (no matter if encoded as a json string or with a separator) as an entire file. It must not complete and it's hard for the LLM to know when to stop completing especially with the code also present in context.

While it's worth testing what happens, I don't expect a change of transport to change the fundamental challenge. As opposed fine tuning also with synthetic training can improve things as the LLM would understand what a virtual window into file snppnet is then. That's what I think it needs to get to fully leverage the ability to get freely expanding windows into the code files that rg-edit provides to the AI. rg-edit could be used as tool as well in theory although it's made for human input (regexp selection) and for human review.

There's also the possibility for you to code rg-edit in gptel native, so you don't have to start an external program, but it'd be more complicated than wgrep. vimini did that. I personally prefer the external tool holding the state and doing the file writing, plus it's usable for other use cases, not just LLM.

karthink:
> Ah yes, no problem that's the way it'd be intended to work. So it's up to you if
> you want to deal with the extra work of updating the versioning.

I've added `gptel-version` as a constant now.  Of course, it's not useful here until the next version of gptel is released.

> So I imagine the rg output would be encoded as a list of json strings to be sent
> to the LLM. That sounds like something that should be tested just to see what
> happens.

No, you send the same input that you're sending now.  It's the LLM _output_ schema that's specified.

> Here there's no constraint that can be applied to improve accuracy of the
> softmax pick in output from the json parser, not even on when to end a
> sentence. So the challenge is to give a prompt that makes it stay constrained
> and not treat each snippnet (no matter if encoded as a json string or with a
> separator) as an entire file. It must not complete and it's hard for the LLM to
> know when to stop completing especially with the code also present in context.
> 
> While it's worth testing what happens, I don't expect a change of transport to
> change the fundamental challenge. As opposed fine tuning also with synthetic
> training can improve things as the LLM would understand what a virtual window
> into file snppnet is then. That's what I think it needs to get to fully leverage
> the ability to get freely expanding windows into the code files that rg-edit
> provides to the AI. rg-edit could be used as tool as well in theory although
> it's made for human input (regexp selection) and for human review.

Sorry, I'm having trouble following what you're saying here.  The advantage of using JSON output is to keep each replacement separated from the rest.  Otherwise, there is no difference in the quality of the result.  I thought this was an issue since you mentioned that maintaining the delimiters between replacements was a problem.
 
> There's also the possibility for you to code rg-edit in gptel native, so you
> don't have to start an external program, but it'd be more complicated than
> wgrep. vimini did that. I personally prefer the external tool holding the state
> and doing the file writing, plus it's usable for other use cases, not just LLM.
 
Yes, I understand this part -- separating the LLM loop from producing the grep results and writing the files makes sense.

aagit:
> > Ah yes, no problem that's the way it'd be intended to work. So it's up to you if
> > you want to deal with the extra work of updating the versioning.
> 
> I've added `gptel-version` as a constant now. Of course, it's not useful here until the next version of gptel is released.

Perfect thanks.

> No, you send the same input that you're sending now. It's the LLM _output_ schema that's specified.

So it's not a plain rewrite anymore, it'd be a rewrite + format conversion, I thought it would give higher accuracy to feed json in input to avoid the conversion step.

> > Here there's no constraint that can be applied to improve accuracy of the
> > softmax pick in output from the json parser, not even on when to end a
> > sentence. So the challenge is to give a prompt that makes it stay constrained
> > and not treat each snippnet (no matter if encoded as a json string or with a
> > separator) as an entire file. It must not complete and it's hard for the LLM to
> > know when to stop completing especially with the code also present in context.
> > While it's worth testing what happens, I don't expect a change of transport to
> > change the fundamental challenge. As opposed fine tuning also with synthetic
> > training can improve things as the LLM would understand what a virtual window
> > into file snppnet is then. That's what I think it needs to get to fully leverage
> > the ability to get freely expanding windows into the code files that rg-edit
> > provides to the AI. rg-edit could be used as tool as well in theory although
> > it's made for human input (regexp selection) and for human review.
> 
> Sorry, I'm having trouble following what you're saying here. The advantage of using JSON output is to keep each replacement separated from the rest. Otherwise, there is no difference in the quality of the result. I thought this was an issue since you mentioned that maintaining the delimiters between replacements was a problem.

When you provide a schema with an int, the softmax in output of the transformer can be picked from the highest softmax percentage after excluding all tokens that can't fit into the json parser as an integer.

My output is not an int or a date or a list of cities, it's code in any language mixed as well as documentation that is changed at once in the rg-edit rewrite.

So the only way I can imagine the json output is a hash (or list) of files, and each file having a list of strings associated to it.

One problem is the LLM can decide to "unify" two snippet, that are relatively close in the file, by not having restraint and by doing what it does best (i.e. completing the next word it has already seen). That's when it eats the separator and then you started with 5 snippnets and the output has 4.

here's an example:

-- file1
A
BX
C
DX
E

I grep for X and I get this rg-edit buffer with 1 file and 2 snippets:

-- file1
BX
--
DX
--

After the rewrite this is the only format that is valid (otherwise the apply will fail gracefully logging the exact format failure in the rg-edit logging buffer):

-- file1
somestring
--
someother
--


So a json schema would need to become: {'file1': ['somestring','someother']}. 

However somestring and someother can be anything including // file2 which happened once. The softmax can't be constrained when it starts to write somestring.

So I believe json would prevent to corrupt the first "-- file1" but not a second "// file2" header and it can't prevent to eat the separators.

The challenge is to have the LLM constrained from doing this:

-- file1
BX
C
DX
--

And return 1 snippet.

The LLM really wants to complete and fixup stuff it doesn't expect.... unless told to stop at the separator with the prompt. The separator also needs to be clearly in non language compatible format (i.e. // DO NOT DELETE THIS SEPARATOR woudln't work and it would replace it with "C").

I found with many tries -- DO NOT DELETE THIS SEPARATOR which gives a directive within the separator itself appears the only way to make this work generally cross LLM and make it emit the separator instead of C, but it must be combined with a ad-hoc prompt. A tiny bit of fine tuning on the separator would go a long way....

So what would happen with the JSON schema? The challenge would become not to make the LLM emit this: {'file1': ['BX\nC\nDX\n','']}.

When the LLM goes off the rails it'll become undectable if the schema enforces the format to be compliant.

In other words, it may very well guarantee rg-edit doesn't reject an LLM edit of the rg-edit buffer, but by doing so it'll make it impossible for rg-edit to detect automatically when the LLM went off the rails. So the reviewer would need to more work to detect the LLM did a mistake and revert the code applied by mistake. Instead of making rg-edit LLM rewrites more reliable, the json schema would make them less reliable and more prone to hallucinated corruption landing in files requiring a git checkout -p to fixup.

It's not intuitive that a feature meant to increase relaibility and robustness would lead to the expect opposite result but that's because I don't have a integer, a date or a list of cities to filter the softmax against.

I'm thinking aloud so I may be missing something.

> > There's also the possibility for you to code rg-edit in gptel native, so you
> > don't have to start an external program, but it'd be more complicated than
> > wgrep. vimini did that. I personally prefer the external tool holding the state
> > and doing the file writing, plus it's usable for other use cases, not just LLM.
> 
> Yes, I understand this part -- separating the LLM loop from producing the grep results and writing the files makes sense.

Sounds great, thanks for sharing your view!

PR #1096, opened 2025-10-09
Revert "gptel-rewrite: Diff buffers, not files (#731)"

aagit:
This reverts commit 0ce917a7f091eee7d2148ba53536816106f0a7b0.

The above commit while not wrong per se, prevents to use diff-mode to apply hunks, which makes it not a good tradeoff.

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -354,7 +354,10 @@ BUF is the buffer to modify, defaults to the overlay buffer."
               ((buffer-live-p ov-buf)))
     (require 'diff)
     (let* ((newbuf (gptel--rewrite-prepare-buffer ovs))
-           (diff-buf (diff-no-select ov-buf newbuf switches)))
+           (diff-buf (diff-no-select
+                      (if-let* ((buf-file (buffer-file-name ov-buf)))
+                          (expand-file-name buf-file) ov-buf)
+                      newbuf switches)))
       (with-current-buffer diff-buf
         (setq-local diff-jump-to-old-file t))
       (display-buffer diff-buf))))


karthink:
If possible, I would rather find a way to get diff-mode to apply the hunks to buffers.  I'll look into it, but if you've studied this already you can let me know if this is possible.

aagit:
That I would prefer too indeed. It didn't appear possible, but I didn't spend much time trying.

pabl0:
Isn't this possible in ediff?

aagit:
I'm sorry I don't find ediff a good alternative. One thing is to review code one thing is to review what changed.  In this case I need to review what changed, akin to git log. When I do conflict resolution I prefer to see code not diffs as I need to review code as well not just what changed. ediff is not my favorite interface in general. diff mode is also quite nice in how it allows to split a diff and selectively apply or drop hunks.

pabl0:
How about two separate actions: **d**iff (buf), diff **f**iles, or something along those lines? Too confusing?

aagit:
a diff files option sounds great with me.... it's a workaround, but it's certainly simpler than a upstream diff-mode change. It should be simple to drop if diff-mode is improved later.

The biggest upstream shortfall is not diff-mode but the scrolling over overlays, I *really*  *really* would like the inference to keep showing the last character of the overlay in my window with auto scrolling, I don't even pretend to move the pointer inside the overlay if I could just show it automatically as it is extended by the inference.... Similarly to what you can achieve with gptel-auto-scroll. Alternatively if it's simpler it should be possible to move the pointer through the overlay with up/down pgup/pgdown. So I don't think it's worth spending time on diff-mode when there's a bigger usability problem in the core.

PR #1108, opened 2025-10-15
Fix context ordering for KV cache reuse

aagit:
With the current LIFO order of the gptel-context collection, any hooks that automatically add buffers to the context end up preventing the KV cache reuse at inference time, causing lots of extra unnecessary GPU computations.

New context should always be added at the bottom of the prompt to maintain cache efficiency.

Added `gptel-context-lifo` to control context ordering, restoring the previous behavior as an opt-in option.

### DIFF
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -480,7 +480,7 @@ Ignore overlays, buffers and files that are not live or readable."
                                  (list :mime (mailcap-file-name-to-mime-type entry))))
                  res)))
         ((pred buffer-live-p) (push (list entry) res)))) ;Just a buffer
-    (nreverse res)))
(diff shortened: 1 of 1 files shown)


karthink:
@aagit Thanks for the PR.  That makes sense.  I don't think we need `gptel-context-lifo` or any customization here -- gptel doesn't make any guarantees on the ordering of context chunks, so we can simply do what's most efficient.

The result is going to be a one line change, from `(nreverse res)` to `res`.  You can make the change and I can merge it, or I can make the change, whatever you'd prefer.

karthink:
@aagit Thank you for the fix!

PR #1110, opened 2025-10-17
gptel-context.el: move the context before the prompt

aagit:
Enhance the KV cache reuse further. This is particularly helpful with the gptel-rewrite-directives-hook that may change the prompt at every request depending on the language of the buffer.

The model seem to follow the prompt better when it's at the end (at least open weight qwen3 coder instruct), for example it starts to follow the instruction not add markdown fences around the rewritten code that is otherwise ignored on a long context.

Supposedly the fact the context format is actively using the markdown quotes, washes the directive not to use the markdown quotes if it's done at the top of the system message.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -55,6 +55,14 @@
 
 ** New features and UI changes
 
+- New minor-mode ~gptel-highlight-mode~ to highlight LLM responses and
+  more.  An oft-requested feature, gptel can now highlight responses by
+  decorating the (left) margin or fringe, and apply a face to the
+  response region.  To use it, just turn on ~gptel-highlight-mode~ in
+  any buffer (and not just dedicated chat buffers).  You can customize
+  the type of decoration performed via ~gptel-highlight-methods~, which
+  see.
+
 - Link annotations: When ~gptel-track-media~ is enabled in gptel chat
   buffers, gptel follows (Markdown/Org) links to files in the prompt and
   includes these files with queries.  However, it was not clear if a
--- a/README.org
+++ b/README.org
@@ -1252,6 +1252,12 @@ gptel provides a few powerful, general purpose and flexible commands.  You can d
 
 #+html: <img src="https://github.com/karthink/gptel/assets/8607532/3562a6e2-7a5c-4f7e-8e57-bf3c11589c73" align="center" alt="Image showing gptel's menu with some of the available query options.">
 
+-----
+
+You can use =gptel-highlight-mode= to highlight LLM responses in different ways:
+
+#+html: <img width="1226" height="764" alt="Image" src="https://github.com/user-attachments/assets/1a3d5411-3cbc-41cf-b9fa-5089a1ffa096" />
+
 You can also define a "preset" bundle of options that are applied together, see [[#option-presets][Option presets]] below.
 
 *** In a dedicated chat buffer:
@@ -1269,6 +1275,8 @@ You can also define a "preset" bundle of options that are applied together, see 
 That's it. You can go back and edit previous prompts and responses if you want.
 
 The default mode is =markdown-mode= if available, else =text-mode=.  You can set =gptel-default-mode= to =org-mode= if desired.
+
+You can use =gptel-highlight-mode= to highlight LLM responses in different ways.
 
 You can also define a "preset" bundle of options that are applied together, see [[#option-presets][Option presets]] below.
 
@@ -1588,9 +1596,11 @@ You can also call =gptel-end-of-response= as a command at any time.
 **** I want to change the formatting of the prompt and LLM response
 #+html: </summary>
 
-For dedicated chat buffers: customize =gptel-prompt-prefix-alist= and =gptel-response-prefix-alist=.  You can set a different pair for each major-mode.
+Anywhere in Emacs: Turn on =gptel-highlight-mode=.  See its documentation for customization options.
 
-Anywhere in Emacs: Use =gptel-pre-response-hook= and =gptel-post-response-functions=, which see.
+In dedicated chat buffers: you can additionally customize =gptel-prompt-prefix-alist= and =gptel-response-prefix-alist=, which are prefixes inserted before the prompt and response.  You can set a different pair for each major-mode.
+
+For more custom formatting: Use =gptel-pre-response-hook= and =gptel-post-response-functions=, which see.
 
 #+html: </details>
 #+html: <details><summary>
@@ -1750,6 +1760,7 @@ Other Emacs clients for LLMs prescribe the format of the interaction (a comint s
 | =gptel-temperature=   | Randomness in response text, 0 to 2.                    |
 | =gptel-cache=         | Cache prompts, system message or tools (Anthropic only) |
 | =gptel-use-context=   | How/whether to include additional context               |
+| =gptel-context=       | List of context sources (files/buffers) for queries     |
 | =gptel-use-tools=     | Disable, allow or force LLM tool-use                    |
 | =gptel-tools=         | List of tools to include with requests                  |
 |-----------------------+---------------------------------------------------------|
@@ -1758,6 +1769,7 @@ Other Emacs clients for LLMs prescribe the format of the interaction (a comint s
 | *Chat UI options*              |                                                                |
 |--------------------------------+----------------------------------------------------------------|
 | =gptel-default-mode=           | Major mode for dedicated chat buffers.                         |
+| =gptel-highlight-methods=      | Highlighting type(s) used by =gptel-highlight-mode=            |
 | =gptel-prompt-prefix-alist=    | Text inserted before queries.                                  |
 | =gptel-response-prefix-alist=  | Text inserted before responses.                                |
 | =gptel-track-response=         | Distinguish between user messages and LLM responses?           |
@@ -1820,7 +1832,7 @@ More generally, you can specify a bundle of options:
   :tools '("read_buffer" "modify_buffer")) ;gptel tools or tool names
 #+end_src
 
-Besides a couple of special keys (=:description=, =:parents= to inherit other presets), there is no predefined list of keys.  Instead, the key =:foo= corresponds to setting =gptel-foo= (preferred) or =gptel--foo=.  So the preset can include the value of any gptel option.  For example, the following preset sets =gptel-temperature= and =gptel-use-context=:
+Besides a couple of special keys (=:description=, =:parents= to inherit other presets), there is no predefined list of keys.  Instead, the key =:foo= corresponds to setting =gptel-foo= (preferred) or =gptel--foo=.  So the preset can include the value of any gptel option.  For example, the following preset sets =gptel-temperature=, =gptel-use-context= and =gptel-context=, a list of files to include as context:
 
 #+begin_src emacs-lisp
 (gptel-make-preset 'proofreader
@@ -1828,8 +1840,9 @@ Besides a couple of special keys (=:description=, =:parents= to inherit other pr
   :backend "ChatGPT"
   :model 'gpt-4.1-mini
   :tools '("read_buffer" "spell_check" "grammar_check")
-  :temperature 0.7                      ;sets gptel-temperature
-  :use-context 'system)                 ;sets gptel-use-context
+  :use-context 'system                                ;sets gptel-use-context
+  :context '("./.grammar_rules.md" "./jargonfile.md") ;sets gptel-context
+  :temperature 0.2)                                   ;sets gptel-temperature
 #+end_src
 
 Switching to a preset applies the specified settings without affecting other settings.  Depending on the scope option (~=~ in gptel's transient menu), presets can be applied globally, buffer-locally or for the next request only.
--- a/gptel-context.el
+++ b/gptel-context.el
@@ -346,12 +346,14 @@ afterwards."
   "Highlight the region from START to END.
 
 ADVANCE controls the overlay boundary behavior."
-  (let ((overlay (make-overlay start end nil (not advance) advance)))
+  (let ((overlay (make-overlay start end nil (not advance) advance))
+        (buf-entry (alist-get (current-buffer) gptel-context)))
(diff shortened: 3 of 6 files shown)


meain:
I don't personally think this is the right change to make and so wanted to write down my thoughts on this.

> Enhance the KV cache reuse further. This is particularly helpful with the gptel-rewrite-directives-hook that may change the prompt at every request depending on the language of the buffer.

Are you talking about the prompt caching? If so, I feel like you'll have more cases where the context is changed but the the same prompt than the other way round. For example it is like more often that you'll be refactoring different bits of code with the same system prompt.

> The model seem to follow the prompt better when it's at the end (at least open weight qwen3 coder instruct), for example it starts to follow the instruction not add markdown fences around the rewritten code that is otherwise ignored on a long context.

I am not so sure about that. I have found that the general pattern in other implementations to be the opposite (ie what gptel currently has). In fact, if you ask me, context should not have been part of the system prompt at all.

> Supposedly the fact the context format is actively using the markdown quotes, washes the directive not to use the markdown quotes if it's done at the top of the system message.

I don't think that is a strong reason to change the ordering. Even with this is true, I don't know if changing the order would affect this. Maybe we can consider changing the context use xml blocks instead of markdown quotes if that would make more sense.

aagit:
> I don't personally think this is the right change to make and so wanted to write down my thoughts on this.
> 
> > Enhance the KV cache reuse further. This is particularly helpful with the gptel-rewrite-directives-hook that may change the prompt at every request depending on the language of the buffer.
> 
> Are you talking about the prompt caching? If so, I feel like you'll have more cases where the context is changed but the the same prompt than the other way round. For example it is like more often that you'll be refactoring different bits of code with the same system prompt.
> 

I'm not sure I understand this. Looking at the code in question:

https://github.com/karthink/gptel/blob/master/gptel-rewrite.el#L159

The %s is the issue. It changes every time you switch between file types (.c, .py, .rs), and I don't want the entire KV cache to be invalidated each time I invoke gptel-rewrite on a different file extension within the same project.

So either:
1. Remove the %s entirely from the programming prompt, or
2. Reverse the order as shown in the provided commit

> > The model seem to follow the prompt better when it's at the end (at least open weight qwen3 coder instruct), for example it starts to follow the instruction not add markdown fences around the rewritten code that is otherwise ignored on a long context.
> 
> I am not so sure about that. I have found that the general pattern in other implementations to be the opposite (ie what gptel currently has). In fact, if you ask me, context should not have been part of the system prompt at all.
> 

If that's your view, you should set the variable to tell gptel to keep context out of the system prompt. This can already be done for such models by setting gptel-use-context to 'user. In that case, this pull request is supposed to have zero effect - you'd just need to set that variable.

Moving the context out of the system prompt is a feature already implemented in gptel, if this commits breaks it, it'd be a bug.

> > Supposedly the fact the context format is actively using the markdown quotes, washes the directive not to use the markdown quotes if it's done at the top of the system message.
> 
> I don't think that is a strong reason to change the ordering. Even with this is true, I don't know if changing the order would affect this. Maybe we can consider changing the context use xml blocks instead of markdown quotes if that would make more sense.

The only objective here is to avoid invalidating the KV cache when the dynamic prompt changes, so converting the markdown fences  to XML won't help with that.

The fact the prompt appears to be followed more closely when moved to the end, because the markdown fences finally disappeared from generation, is not the objective of this change.

I mentioned this detail in the header, because it's an hopefully useful indicator that there are no apparent downsides to moving the prompt to the end. If something it was washed out and ignored when kept at the top.

All of these are transformers, the nature of the transformer is the closest text is the more meaningful to generation of the one far in the past (all attention being equal), this is how it always answers to the last question and not to the second last by mistake. This being a kvcache behavioral concern, all LLM use  the same kvcache flash attention vulkan, rocm and cuda kernels, so it's hard to imagine my open source open weight model working "better" with this change, while others don't. Still, I can't rule it out, but if one model doesn't work well, maybe it should use the 'user setting in the first place.

The only really appealing alternative to this approach to me appears the use of a fixed non dynamic programming prompt that drops all %s from the above link, which is easier, but this appeared a win win compared to dropping the dynamic prompt.

Thanks a lot for the review.

meain:
Thanks for the detailed response. IIUC, what you are saying is that if we have a fixed context, but we are running rewrites in different language buffers, the prompt cache cannot be reused as the language will be different; and that if we were to flip the ordering, we could be caching the context information.

I agree with this, but my concern is that you are more likely to have a fixed system prompt with changing context - consider cases where you are using a fixed system prompt with non rewrite gptel commands or having a conversation in a gptel buffer. In these cases it would be more useful to have it the current way.

---

> All of these are transformers, the nature of the transformer is the closest text is the more meaningful to generation of the one far in the past (all attention being equal), this is how it always answers to the last question and not to the second last by mistake. This being a kvcache behavioral concern, all LLM use the same kvcache flash attention vulkan, rocm and cuda kernels, so it's hard to imagine my open source open weight model working "better" with this change, while others don't.

Ahh I see. I was not aware of the technical side of things. I was mostly going based on the patterns that I think I have internalized at some point 😅

aagit:
> Thanks for the detailed response. IIUC, what you are saying is that if we have a fixed context, but we are running rewrites in different language buffers, the prompt cache cannot be reused as the language will be different; and that if we were to flip the ordering, we could be caching the context information.
> 
> I agree with this, but my concern is that you are more likely to have a fixed system prompt with changing context - consider cases where you are using a fixed system prompt with non rewrite gptel commands or having a conversation in a gptel buffer. In these cases it would be more useful to have it the current way.

That's a good point - the context does change more often in many scenarios. However, the prompt is generally just a dozen lines at most (even if it's a hundred lines, it's negligible compared to the size of the context). So even if I'm going to recompute the prompt more frequently now, it's not really noticeable. The user and assistant messages also have to be recomputed, but those are also small compared to the context.

So while this change will cause the prompt to be cached less frequently, it's still a win overall, unless of course the prompt never changes. If the prompt never changes, I agree it would be more efficient to keep it before the context, but that's not how gptel-rewrite operates. Gptel changes the prompt dynamically depending on where you're invoking it: in gptel-mode it uses the default prompt, in gptel-rewrite it uses the rewrite prompt, etc.

On a side note, I noticed with this commit applied, if I turn off context completely with "-i No" in gptel-mode (in buffer mode), I can keep using the 100k token context cache for gptel-rewrite in other files. So the question is why the llama.cpp server can't cache all those different prompts + context too - maybe it hits some limits. I use --reuse-cache 256 if that matters.

There are lots of "ifs" about why it's not caching after a certain point that would be nice to dig deeper into, but this commit is inference server agnostic: here we simply try to generate the fewest amount of "different" KV cache payloads that the server will ever have to cache.

> > All of these are transformers, the nature of the transformer is the closest text is the more meaningful to generation of the one far in the past (all attention being equal), this is how it always answers to the last question and not to the second last by mistake. This being a kvcache behavioral concern, all LLM use the same kvcache flash attention vulkan, rocm and cuda kernels, so it's hard to imagine my open source open weight model working "better" with this change, while others don't.
> 
> Ahh I see. I was not aware of the technical side of things. I was mostly going based on the patterns that I think I have internalized at some point 😅

Well, I hope my understanding of how KV cache works isn't flawed... but if I'm wrong, I'd appreciate being corrected.

meain:
I think I'm starting to agree with you about moving it due to caching reasons. About what would yield better, I guess I'll play around and report if we do get any major change in the performance of responses.

I'm not yet fully convinced yet that having context before would be beneficial for actual outputs mostly because I've seem most articles mention adding context after system prompt for example with https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents .

faa261102e46c7f090a2402a49000ffae18c5dd6-2292x1290

karthink:
**From the perspective of retaining the KV cache**, it makes sense to place the context before the instructions because the context is usually much longer.  Even if you change the context more often than the instructions (as @meain says), the savings from placing the instructions first will be negligible.  So I agree with @aagit here.

**From the perspective of recency causing better adherence to instructions**, such as following the instruction to not add markdown fences: I am not sure about this.  Any mental model about how transformers work is not adequate for making this decision, because in practice there are many unknown variables.  We'll have to test both strategies (context before/after instructions) with different models, or find some data online about this.  So I'm looking for more information here.  This is not a blocker for merging this, unless this somehow makes adherence *worse*, which I doubt.

**As for the question of whether the context should go in the system prompt or user prompt**, there is no consensus as far as I know.  I've seen recommendations for both from Anthropic and OpenAI at different times in the past two years.  One question I have is whether the context should go at the top of or bottom of the user prompt.  I don't know.

karthink:
> after system prompt for example with
> https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents

My reading of this image is that documents (i.e. context) are not part of the system prompt at all, as seen in the "Context window" block on the right.

karthink:
> The model seem to follow the prompt better when it's at the end (at least open weight qwen3 coder instruct), for example it starts to follow the instruction not add markdown fences around the rewritten code that is otherwise ignored on a long context.

I just tested this with Sonnet-4-5 and gpt-4.1, and it did not make a difference where the context was placed relative to the instructions -- they behaved the same way in both cases with regards to the markdown fences.  Sonnet-4-5 produced backticks both times, gpt-4.1 did not.  Perhaps this matters more to smaller models.  (I understand that this is not the main reason for this PR.)

aagit:
On my side I tested qwen3 coder and gemini 2.5 pro. I didn't have a way to measure if it is followed more or less closely after this change, but it was easy to verify it still works after context too. And like reported earlier the markdown fences are finally gone from gptel-rewrite generation with context enabled so it's not ignored at least.

I tested it with this in gptel-mode:

```
*** explain the theory of relativity
:PROPERTIES:
:GPTEL_SYSTEM: you're a comic, make me laugh
:END:
```

(for the tests I double checked with Inspect there were >13k lines of code and a few compilation buffers before the system message)

```
Grep finished with 15 matches found at Sat Oct 18 15:27:42, duration 0.01 s

...
\```

you're a comic, make me laugh")
			   (:role "user" :content
				  "explain the theory of relativity")]
	:stream t)
```

So I try to be optimistic given how much computations it's saving me with >100k context loaded...

aagit:
Another data point, I did the simplest  test of [ripgrep-edit](https://gitlab.com/aarcange/ripgrep-edit) but with a >100k token context loaded w/ and w/o this patch.

The ripgrep-edit format requires a pretty specific prompt to be rewritten correctly. Usually a large context made the rewrite process fail. As expected with the 100k context it doesn't even get close to do the right thing.

With the same context but this commit applied, ripgrep-edit just works as with no context.

You can verify this by forcing this prompt in gptel-rewrite with "s s" (please double check with Inspect you got the write prompt setup).

```
You are a careful programmer. Rewrite cross-file snippnets:
- Rewrite everything exactly the same, except: the required change
- Keep the filename at the start of each file
- Keep the separator at the end of all snippnets
- Do not add markdown fences
- Do not ask clarification
```

This is the text to be rewritten with gptel-rewrite:

```
/home/andrea/devel/ai/ripgrep-edit/src/main.rs
    /// context_separator character
    #[clap(long, default_value = "\x7F")]
    context_separator: String,
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let args = Args::parse();

    // Run ripgrep with context lines
    let context_separator = &args.context_separator;

-- DO NOT DELETE THIS SEPARATOR --
                panic!("context_separator found after colon in line: {line}");
            }
            writeln!(file, "{after_colon}")?;
        } else {
            let mut line = line.chars().collect::<Vec<char>>();
	    line.dedup_by(|a, b| *a == '/' && *a == *b);
	    let line = line.iter().collect::<String>();
            writeln!(file, "{line}")?;
        }
    }
    file.flush()?;
-- DO NOT DELETE THIS SEPARATOR --
        if let Some(ranges) = file_ranges.get(file_path) {
            // Read the original file
            let content = std::fs::read_to_string(file_path)?;
            let mut lines: Vec<String> = content.lines().map(|s| s.to_string()).collect();

            // Process ranges in reverse order to maintain correct indices
            for (i, range) in ranges.iter().enumerate().rev() {
                let start = range.0 - 1; // Convert to 0-based indexing
                let end = range.1; // End is exclusive in our range

                // Replace the lines in the file
-- DO NOT DELETE THIS SEPARATOR --

        // Check if line is a file path
        if !line.chars().next().is_some_and(|c| c.is_ascii_digit()) {
            assert!(prev_line_empty);
            let mut line = line.chars().collect::<Vec<char>>();
	    line.dedup_by(|a, b| *a == '/' && *a == *b);
	    let line = line.iter().collect::<String>();
            if Path::new(&line).exists() {
                if let Some(ref current) = current_file {
                    if current != &line {
                        if let Some(range) = current_range {
-- DO NOT DELETE THIS SEPARATOR --

            prev_line_empty = false;
            continue;
        }
        let mut normalized_line = line.chars().collect::<Vec<char>>();
	normalized_line.dedup_by(|a, b| *a == '/' && *a == *b);
	let normalized_line = normalized_line.iter().collect::<String>();
        if file_ranges.contains_key(&normalized_line) {
            assert!(prev_line_empty);
            if !current_file.is_empty() {
                assert!(!current_lines.is_empty());
-- DO NOT DELETE THIS SEPARATOR --
            "Command failed with: {}",
            error_str
        );

        // Check that the output contains the expected lines
        assert!(output_str.contains("src/main.rs"));
        assert!(output_str.contains("context_separa"));
        assert!(error_str.contains("File was not modified"));
    }
}
-- DO NOT DELETE THIS SEPARATOR --
```

This the directive to past in the minibuffer:

```
Rewrite: create a function before main to do the deduplicating of '/' with dedup_by and use it instead of using dedup_by in the call sites using dedup_by
```

So at least for this case the prompt after context is much more effective and it appears a win win. ripgrep-edit format does not use markdown fences anywhere, so changing the context format shall have no effect.

karthink:
Merged on an experimental basis.  If we start seeing issues from users about models not using the context well, we'll revisit this.

Thanks for the PR @aagit!

PR #1115, opened 2025-10-22
Gh: update new models

kiennq:


karthink:
I added the changes but messed up a force push somewhere so the PR shows as rejected.  It's merged.  Thanks for the PR!

PR #1120, opened 2025-10-23
Support thinking responses from Ollama

### LINKED ISSUE #1011: Doesn't handle `thinking` key for ollama/gpt-oss responses

### Bug Description

While using

- ollama 0.11.2
- gptel 20250808.939
- model: gpt-oss:20b

`gptel` does not include any thinking output. 

The reason for this seems to be that the API responses for this model have this shape: `... "message":{"role":"assistant","content":"","thinking":"..."}...`,
unlike the others I tried which only use `content` and bracket their thinking with a special word. Now, every message either has a nonempty `content` or `thinking` field but not both.

### Backend

Ollama

### Steps to Reproduce

Using a `gpt-oss` model through ollama, attempt any request with "Include reasoning" enabled. No reasoning steps are included in output. 

### Additional Context

- ollama 0.11.2
- gptel 20250808.939
- model: gpt-oss:20b

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

nottwo:
Hi, this fixes #1011 for me.

I've used this with Ollama serving both qwen3:30b-a3b and gpt-oss:20b, with and without streaming enabled.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -1065,10 +1065,9 @@ The result is a string intended for display.  Newlines are
 replaced with REPLACEMENT."
   (cl-typecase directive
     (string
-     (concat
-      (string-replace "\n" (or replacement " ")
-                      (truncate-string-to-width
-                       directive width nil nil t))))
+     (string-replace
+      "\n" (or replacement " ")
+      (substring directive 0 (min width (length directive)))))
     (function
      (concat
       "λ: "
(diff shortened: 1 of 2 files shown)


karthink:
Thanks for the PR @nottwo!  I want to test it too, because there tend to be many edge cases when there are both reasoning and tools involved in a response.  I'll merge it after I have time to do that.

karthink:
Tested it out, it works well!  Pushed with some minor changes.  Thanks for the PR!

PR #1124, opened 2025-10-30
Support streaming tool calls with Ollama

nottwo:


karthink:
@nottwo Thanks for the PR.  I haven't been able to test this yet (no access to a GPU right now), I'll do it soon.

It looks like no special handling was required for streaming tool calls.  Is a tool call delivered in full inside a streaming chunk every time?

karthink:
@nottwo Merged with minor tweaks, thanks!

PR #1127, opened 2025-10-30
gptel--markdown-validate-link: dont increase 3 to file link extraction if file:// is not presented

kiennq:
The current logic where we always add `3` even when the `file://` is not there is not working for relative file path, ex: `./asset/a.jpg`


karthink:
@kiennq thanks for the fix!

PR #1131, opened 2025-11-02
Add latest anthropic models to bedrock, remove eol model

matthemsteger:
Add the latest anthropic models and remove sonnet 3, since it is EOL (see https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html).

I kept the latest naming scheme (version last)

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -697,92 +697,99 @@ Search between BEG and END."
         (and link-ovs (mapc #'delete-overlay link-ovs))))
     `(jit-lock-bounds ,beg . ,end)))
 
-(defun gptel-use-header-line ()
-  "Set up the header-line for a gptel buffer."
-  (setq
-   header-line-format
-   (list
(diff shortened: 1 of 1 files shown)


karthink:
@matthemsteger Thanks for the PR.  Please add a NEWS entry too, under "New Models and Backends", wrapped to 72 chars/row.  You can use the format of the existing entries there.

matthemsteger:
> @matthemsteger Thanks for the PR. Please add a NEWS entry too, under "New Models and Backends", wrapped to 72 chars/row. You can use the format of the existing entries there.

@karthink It looks like all the models I added are already represented by name in the NEWS file. I assume from direct support for Anthropic (not bedrock). I didn't see where we called out bedrock specifically in the past. Did you want me to add something for bedrock specifically? Opus 4.1 was added in August, but the other two are in the latest model section.

karthink:
@matthemsteger You're right, they're covered already.  Merged, thanks for the PR!

PR #1133, opened 2025-11-04
Refactor gptel-gh-login for SSH and terminal Emacs users

jsntn:
Remove automatic browser opening via browse-url and instead instruct users to manually visit https://github.com/login/device. This fixes authentication issues when:
- Running Emacs over SSH where browse-url opens on remote machine
- Using terminal Emacs where eww cannot complete OAuth flow

Also fix backend resolution to work when gptel-backend is not a GitHub Copilot backend, preventing "Wrong type argument" errors.

Changes:
- Remove browse-url call that fails in SSH/terminal contexts
- Update prompt to direct users to https://github.com/login/device
- Add backend resolution logic to find GitHub backend if not current
- Update docstring to mention SSH/terminal use case
- Add immediate message showing device code

### DIFF
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -243,40 +243,72 @@
 (defun gptel-gh-login ()
   "Login to GitHub Copilot API.
 
-This will prompt you to authorize in a browser and store the token."
+This will prompt you to authorize in a browser and store the token.
+
+In SSH sessions, the URL and code will be displayed for manual entry
+instead of attempting to open a browser automatically."
   (interactive)
-  (pcase-let (((map :device_code :user_code :verification_uri)
-               (gptel--url-retrieve
-                   "https://github.com/login/device/code"
-                 :method 'post
-                 :headers gptel--gh-auth-common-headers
-                 :data `( :client_id ,gptel--gh-client-id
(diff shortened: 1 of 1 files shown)


karthink:
@jsntn Thanks for the PR.  This looks good.

There is no reason to make Emacs users who are using a local, GUI version have to paste the URL manually into the browser.  

- For remote users, you can instead check if `(getenv "SSH_CLIENT")` or `(getenv "SSH_CONNECTION")` return non-nil.
- For terminal users, I don't understand the issue -- `browse-url` should still open the URL in the default web browser, not EWW.  I have verified this with `emacs -q -nw`.

jsntn:
Hi @karthink! Thank you for the feedback!

I've updated the pull request to only detect SSH sessions via the environment variables you suggested (SSH_CLIENT, SSH_CONNECTION, and also SSH_TTY). The logic now is:

- SSH sessions: Display the URL and code for manual entry
- Local sessions (both GUI and terminal): Preserve the original browse-url behavior

karthink:
Merged with minor formatting changes.  @jsntn Thanks for the PR!

PR #1147, opened 2025-11-21
gptel-context-optimizer: LRU Context Management, one-shot auto add *compilation* buffer, kvcache preloading

aagit:
This PR introduces a LRU (Least Recently Used) context management eliminating the need for manual implementation in `~/.emacs`.

Key features:

1. **Automatic LRU Context**: The LRU context management is now enabled automatically without requiring local configuration
2. **Compilation Buffer Integration**: Adds a configurable option to automatically include the *compilation* buffer at the bottom of the context. This ensures the LLM always has awareness of the most recent compilation issues. Note: this feature is disabled by default as it may not always beneficial.
3. **Relative Path Support**: Obsoletes issue #1113 by keeping relative paths enabled (they are preferable for the LLM too as less is more as far as context is concerned) while fixing the risk of duplication. gptel-context-file-add now does a expanded vs expanded comparison.

### DIFF
--- a/gptel-context-optimizer.el
+++ b/gptel-context-optimizer.el
@@ -2,7 +2,7 @@
 
 ;; Copyright (C) 2026  Red Hat Inc.
 
-;; Keywords: context, LRU, compilation
+;; Keywords: context, LRU, compilation, preload
 
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 
@@ -34,6 +34,22 @@
 ;; gptel-request can be optionally enabled with:
 ;;
 ;;    (setq-default gptel-context-enable-compilation-auto-add t)
+
+;; kvcache context preload will increase energy/cost of inference, but
+;; it may decrease the wait time before generation, so it's disabled
+;; by default.
+;;
+;; It can be toggled with:
+;;
+;;    (gptel-context-preload-mode):
+;;
+;; or enabled with:
+;;
+;;    (gptel-context-preload-mode 1)
+;;
+;; or disabled with:
+;;
+;;    (gptel-context-preload-mode 0)
 
 ;;; Code:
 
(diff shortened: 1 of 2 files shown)


karthink:
Thanks for the PR @aagit.

> 1. **Automatic LRU Context**: The LRU context management is now enabled
>    automatically without requiring local configuration

This feature looks good, but the implementation needs work.  gptel should not be adding anything to _global_ hooks (like `after-save-hook` and `after-revert-hook`) no matter what.  I would like to avoid adding anything to any buffer's local hooks as well, unless it's a gptel chat buffer.

> 2. **Compilation Buffer Integration**: Adds a configurable option to
>    automatically include the *compilation* buffer at the bottom of the
>    context. This ensures the LLM always has awareness of the most recent
>    compilation issues. Note: this feature is disabled by default as it may not
>    always beneficial.

This is a workflow-specific feature and doesn't fit with gptel's design, so please remove it from this PR.

> 3. **Relative Path Support**: Obsoletes issue #1113 by keeping relative paths
>    enabled (they are preferable for the LLM too as less is more as far as
>    context is concerned) while fixing the risk of
>    duplication. gptel-context-file-add now does a expanded vs expanded
>    comparison.
> 

This change is welcome, but I have some suggestions for the implementation:

1. `gptel-context--file-already-added` returns a matching file, so the name needs to reflect that.  Perhaps `gptel-context--get-path` or `gptel-context--get-file`.
2. This function will fail because `gptel-context` can have file paths and buffers that are not inside lists.  For example, this is a valid value for gptel-context: `(setq gptel-context (list "path1" "path2" (get-buffer "somebuf")))`.  See the documentation of `gptel-context`.
3. There is no need to create a duplicate of `gptel-context` in this function.  Something like the following works better:

   ``` emacs-lisp
     (cl-loop
      with expanded-path = (expand-file-name path)
      for entry in gptel-context
      for source = (or (car-safe entry) entry)
      when (and (stringp source)
                (string= expanded-path (expand-file-name source)))
      return entry)
   ```

   (Or an equivalent while loop.)
4. The message `(message "File \"%s\" added to context." path)` should be sent even if the file is already in the path.  Otherwise there is no feedback when running `gptel-add-file` on a file that is already part of the context.
5. Since we're already testing for membership in `--add-text-file` and `--add-binary-file`, there is no need to use `cl-pushnew`.  Regular `push` is enough.

karthink:
Also, I think implementing 2 is much simpler than what you're doing right now.  Just add the compilation buffer to `gptel-context` _once_, and you're set:

``` emacs-lisp
(push (get-buffer "*Compilation*") gptel-context)
```

You shouldn't need to update it again in the Emacs session, as long as the buffer name remains the same.

aagit:
> Thanks for the PR @aagit.

You're welcome.

> > 1. **Automatic LRU Context**: The LRU context management is now enabled
> >    automatically without requiring local configuration
> 
> This feature looks good, but the implementation needs work. gptel should not be adding anything to _global_ hooks (like `after-save-hook` and `after-revert-hook`) no matter what. I would like to avoid adding anything to any buffer's local hooks as well, unless it's a gptel chat buffer.

My after-revert-hook is set to quilt-hook. My auto-save-hook is set to gtags-update-hook.

> > 2. **Compilation Buffer Integration**: Adds a configurable option to
> >    automatically include the _compilation_ buffer at the bottom of the
> >    context. This ensures the LLM always has awareness of the most recent
> >    compilation issues. Note: this feature is disabled by default as it may not
> >    always beneficial.
> 
> This is a workflow-specific feature and doesn't fit with gptel's design, so please remove it from this PR.

I agree it cannot be enabled by default, but not adding an opt-in it means .emacs bloat for everyone who needs this. I thought this is common enough need.

> > 3. **Relative Path Support**: Obsoletes issue [gptel-context: add file path expansion to prevent duplicates #1113](https://github.com/karthink/gptel/pull/1113) by keeping relative paths
> >    enabled (they are preferable for the LLM too as less is more as far as
> >    context is concerned) while fixing the risk of
> >    duplication. gptel-context-file-add now does a expanded vs expanded
> >    comparison.
> 
> This change is welcome, but I have some suggestions for the implementation:
> 
>     1. `gptel-context--file-already-added` returns a matching file, so the name needs to reflect that.  Perhaps `gptel-context--get-path` or `gptel-context--get-file`.
> 
>     2. This function will fail because `gptel-context` can have file paths and buffers that are not inside lists.  For example, this is a valid value for gptel-context: `(setq gptel-context (list "path1" "path2" (get-buffer "somebuf")))`.  See the documentation of `gptel-context`.

Interesting, I didn't run into that, I've only had lists into it, but I don't normally set the gptel-context directly, I use the various function provided by gptel.  Maybe it's only happening when setting it by hand with setq? It looks a bit of a complication to support more than one format mixed though.

>     3. There is no need to create a duplicate of `gptel-context` in this function.  Something like the following works better:
>        ```emacs-lisp
>          (cl-loop
>           with expanded-path = (expand-file-name path)
>           for entry in gptel-context
>           for source = (or (car-safe entry) entry)
>           when (and (stringp source)
>                     (string= expanded-path (expand-file-name source)))
>           return entry)
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        (Or an equivalent while loop.)
> 
>     4. The message `(message "File \"%s\" added to context." path)` should be sent even if the file is already in the path.  Otherwise there is no feedback when running `gptel-add-file` on a file that is already part of the context.
> 
>     5. Since we're already testing for membership in `--add-text-file` and `--add-binary-file`, there is no need to use `cl-pushnew`.  Regular `push` is enough.

I'll look into the improvement suggested. Not sure how to handle the hooks, maybe gptel can implement the hook function and the user shall add the add-hook in .emacs. same for the compilation buffer? The compilation buffer was an incremental separate fix though, maybe it'd be better to handle it in a separate PR.

aagit:
> Also, I think implementing 2 is much simpler than what you're doing right now. Just add the compilation buffer to `gptel-context` _once_, and you're set:
> 
> ```emacs-lisp
> (push (get-buffer "*Compilation*") gptel-context)
> ```
> 
> You shouldn't need to update it again in the Emacs session, as long as the buffer name remains the same.

Yes if you want to just add it, but this is still in the context of the same LRU logic of the first patch: it guarantees at every new build the compilation buffer, if already in context, is removed and re-added at the end, where it's also going to be more effective. The first time you start a build it may invalidate 90% of your context but next time you rebuild it'll be close to zero.

karthink:
@aagit just checking if this PR is still required.  If it is, please let me know if the suggested changes have been made.

aagit:
Hi @karthink, yes it's pretty fundamental and I can't drop it from my gptel or it'd cause a noticeable slowdown. In fact I now combined with preloading, so any invalidate or save of any file in context also starts a new gptel-request immediately on a dedicated buffer to start crunching the new version of the kvcache well before I will need it. The LRU ordering is more important than the preloading, but with both combined I don't have to wait anymore for the context to be crunched. So I would be very happy to clean it so it can be merged. I actually started and I replaced the cl-pushnew with push but that caused a regression, I was getting dups for some reason... but cl-delete seem to still do something, or maybe that's the issue cl-delete just returns a new list and I need to setq gptel-context (cl-delete ... Unfortunately I didn't have time yet to figure why the first cleanups caused regressions so I kept running this version in the PR. I'll be offline next week, I will try to continue the cleanup the week after.

I suppose I need to drop the auto arming of the hooks and the compilation buffer too. btw the compilation buffer now I made it one shot in my daily driver gptel, so it's only available on the first inference after a compilation.

aagit:
Hi @karthink 

most of the feedback has been addressed, on top I added the context prelaoding feature that I'm using all the time for a while which is also a context optimization and  requires the same hooks and same gptel-context--get-path helper so it fits well in this series.

The way I addressed the request not to register into any hook by default is to add a new module. By adding a new module, unless somebody explicitly loads gptel-context-optimizer, nothing at all changes in emacs hooks. If you load it, then it means you need it and you should know what you're doing... If you don't change your ~/.emacs nothing will change so it respects the request not to register into any hook unless you ask for it explicitly.

I believe things should be as seamless as possible, at least with the gptel-context-optimizer opt-in loaded.

aagit:
@karthink ping

karthink:
@aagit Haven't lost track of this, just busy!  Will come back to this soon.

PR #1151, opened 2025-11-22
gptel-gh: update model list

kiennq:
Update new models and remove obsoleted one.
Also update input/output cost as per https://api.githubcopilot.com/models


karthink:
@kiennq thanks!

PR #1152, opened 2025-11-22
Add/remove context buffers from ibuffer

Stebalien:
This PR makes it possible add/remove buffers in bulk from Ibuffer as one can already do with files from Dired.


karthink:
Merged with documentation changes.  Thanks for the PR @Stebalien!

PR #1154, opened 2025-11-24
gptel-gemini: Support thoughtSignature in tool calls

### LINKED ISSUE #1146: support gemini thought_signature
Is it possible to support thought signatures for multiturn gemini conversations?

https://ai.google.dev/gemini-api/docs/thought-signatures

surenkov:
This PR updates the Gemini backend to preserve `thoughtSignature` fields, which is [mandatory for Gemini 3 Pro and recommended for Gemini 2.5](https://ai.google.dev/gemini-api/docs/thought-signatures#model-behavior).

Gemini reasoning models return an encrypted `thoughtSignature` alongside text or function calls to maintain reasoning context in their stateless API. [Per Gemini API specifications](https://ai.google.dev/gemini-api/docs/thought-signatures#:~:text=As%20a%20general,4xx%20status%20code):
> When using Gemini 3 Pro, you must pass back thought signatures during function calling, otherwise you will get a validation error (4xx status code).

Previously, `gptel-gemini` discarded this signature. This change ensures the signature is retained and sent back in the conversation history for subsequent turns, preventing API errors with reasoning models.

*Changes:*
-   **Response Parsing:**
    -   The `thoughtSignature` is extracted and attached to the tool-use plist `:id` property if a `functionCall` is present and the signature is supplied in the response.
    -   For text responses, the `thoughtSignature` is stored as a `'gptel-gemini-thought-signature` text property on the response content in the buffer.
-   **History Extraction:**
    -   The signature is retrieved from the text property for model text responses and from the tool call plist for model function calls.
    -   The retrieved `thoughtSignature` is then included in the message structure sent back to the Gemini API.

Closes #1146


surenkov:
Strictly speaking, I haven't got any idea better than to put the signature as an `:id` of the tool call. At least Gemini does not provide real call ids, and placing the signatures there doesn't break any existing behavior, allowing to carry them across multiple conversation turns. 

@karthink can you please take a look at these changes? Is there a better place of storing these signatures?

karthink:
> Strictly speaking, I haven't got any idea better than to put the signature as an `:id` of the tool call. At least Gemini does not provide real call ids, and placing the signatures there doesn't break any existing behavior, allowing to carry them across multiple conversation turns. 
>
> @karthink can you please take a look at these changes? Is there a better place of storing these signatures?

Thanks!  I'll take a look at the changes soon.  I hope it's clear that no changes are needed to send thoughtSignatures _between_ invocations of gptel-send -- after a tool call, I think they are being sent automatically with the tool result.

surenkov:
@karthink correct, within the single `gptel-send` invocation, the agent loop preserves signatures from the incoming messages.

These changes handle multi-turn conversations - each time I send a followup messages, and the conversation buffer contains tool calls, Gemini API raises HTTP 400, complaining on missing thought signatures in model response parts.

karthink:
> complaining on missing thought signatures in model response parts.

I'm not able to reproduce this with Gemini 3 pro.

surenkov:
@karthink interestingly now I'm also unable to reproduce this behaviour on the master branch. I'll play with this a bit more and try to catch something.

karthink:
I also don't think it's worth handling signatures in non-functionCall parts:

> Gemini may also return thought_signatures in the final part of the response in non-function-call parts.
>
>    Behavior: The final content part (text, inlineData…) returned by the model may contain a thought_signature.
>    Recommendation: Returning these signatures is recommended to ensure the model maintains high-quality reasoning, especially for complex instruction following or simulated agentic workflows.
>    Validation: The API does not strictly enforce validation. You won't receive a blocking error if you omit them, though performance may degrade.

This happens automatically in intra-request turns, and I'm guessing it's not going to affect the model performance between requests.  It also cannot be persisted reliably without adding it to the list of saved properties in the file, etc.

karthink:
> Strictly speaking, I haven't got any idea better than to put the signature as an :id of the tool call

For functionCalls I think storing it as the `:id` of the tool call is fine.

karthink:
Further, the current method of applying the `gptel-gemini-thought-signature` text-property to the response does not work for streaming responses, because the text that receives the thoughtSignature is typically the empty string.  Here is an example:

```json
    "candidates": [
      {
        "content": {
          "parts": [
            {
              "text": "",
              "thoughtSignature": "EocLCoQLAXLI2nz8CP1S2LQbKbm/+ppItrnk1GUToAQB8obIwfszXNJWp0CCqRb1BGY+Oid/88BCa03bHwMdTaMlyHgBUPP667xIuLHslGwUgx7u0TCOBMwrupy+dEb6ugX+BFqq/aB8yCAbOM0u2TyrfH75E8+RJYvwuxJ/7H+ENTuvkRw9nfnlawOVy80KSYHf4zdcW+fRNuDbQAHbITJWnZrht+DcUSE92k+FUuAjSWEnFwoEeT6V7qDeI3qWxoWzaq86wZwyFUQZ81fIwl/BvcHGjkMQFtyTxd4h2I77s21qsOflDhDI9L1ZbuQ/qYIuAKDcG1G5eTGCE22Rvrj6Ckf9g3GO0E7/369qloLoxH3vYjLKD3Qeqkn+CDPSrCP++haHFot1uoK7hJu24UAw6bm/r6i1UX2HB1XJcGODIGAmOg4mEPvWu9N4GWoy47ZA6kDh16GM3Q29cnnzEPQA6Wdw1VjSUykTtRd6bsDqOB1WWkMuNDPbWNZeOigH0/XjnfoXjbiAYEzrGqxN2+ZqLOvT+boudPsduwXnCl/4YgVfq9ERz346DN/7o1wFpcqFIJiqBElyZqJCIK5IAPHLMQaYJIEwnUjRK1iCbIxnD0H+vK4TIfXqcVWmBGqNYudDgx8DhzJiL1KR7IwqjergdcNX/r77sy/HRIKSweprGJapcRLgOqJfrWarffZulCpb3oQTpzVi2YWRlL1DiSQevXx604GlNyV7XXSF1wUfUJxYGiooH80jfLmOruNvHWlth+mhOEoNr8zuwzpv+n9UVdeE+L1nGI6oiliPJ3m3Xh6ftz1fo4+SAccfi09KmF6ZER+SIr4r3BCJitB/mFX3lpft0K2RiT6qQzmc9ugKM7UbWAj/a4mKP0RzziRFPmBZoA9jiCq0MTioAWiz9DIC1lBSmrXXebXrEgxl8TAwiLn0DnDxLKZ0yBtdCwHhGWhsq3maCSfctk+5aSdSmfmWM2NHw53TxcsENu/tfcuVMYv317i6S2AhoGt1sU3orn6eZ7AiBgiGCDyMx41K+bzA04u1ycU/2s5d25441lmlc2m9hOpSuL6i2BvBgz7yTiLmuO+HJkHHKeENrWvwbkGX1IiZBvLckYuZvyr39XwCQXfsSIqfQCqAcWDVLWZ0ISIraSyM7sdh1QnXCOVRucuR3lHCFGvQVGzQhhRn4QAuS0jflQZZL1kq/tU9ma+lNTzPwmaUmCZ6O44EeqsVhGb1WBX7kKrmYIA3W8Vt1lkE7BkZc8arqShxdJ7l1++uqmn3v4GLttaoffu8JjrR1+mOIK3EqqhS1cO6H/udJX8CLGb6sEQCHD6/RpgMFqY3f2fQJYa81V473klhczBatG5QVNgtwdghDusOV1BWVsGePUmLf+hScxKBaYr59fYRJ+wwHlIKxE05hFroEvJ/l2cZAUB4oW+jnKF1LetN642yjgomTe5pRy6j+GPCun/cG31i6wKXm/OB6IkNUPPZ09dji9/R1qmqBrs+BTSv0lgjxxQFmxw+lgydacGIFckfR4O6+d3qIOHno71nUzR1ZFHFAUULHN7+P9IfbQWL6tF8IDFaeAXuz++MS...Olfi8UslMaUONVCEuPdIou8Rqno+Sj21HRgQFv+r97EeAELwDqoJv2cn2L4XhKUV7ziWYTtxr4xSRPjUqW7v7o8eO8NkQiT5aFX3mQTWiNm2llMlPcYs9ouUfTlDPY1hpqKFZSqzsnfGEWJzC6VNdyZUy4YPpZJsgvb3wNkc29m1qRDuIudXnhxWkE5fkxsL3C0ktBfKTsn2EnmnHRVLRpIg3NbPdGfMaQdplJlKicsk2ytrjhBeuIT1nBMsjP1chAE189nO7X0YQjY7b8PEp8jthBHJeqT1oNlFq8upQ="
            }
          ],
          "role": "model"
        },
```

I also expect that the thoughtSignature being recorded as the tool call id is going to be too long for the other backends, so if you switch (in a buffer) from a Gemini model to Anthropic/OpenAI or one of the other providers, those backends will complain.  I haven't tested this yet, so please test if you can. 

Considering that the main issue being addressed by this PR is not reproducible, here's what I suggest:

1. Remove the change involving propertizing the response text with the thought signature, because this cannot be done reliably without a lot more effort, as indicated in my previous messages.
2. Retain the change of adding the thought signature as the tool call id.  If it creates issues when used with other backends, some other approach might be needed, this needs testing.

I can make these changes and merge the PR, or you can make them and let me know.  If you make the changes, please avoid touching the indentation of lines that you don't change.  I will reformat the contents updating the indentation as a separate commit after this PR is merged.

surenkov:
@karthink thank you for the feedback. I have tested the compatibility of using the thoughtSignature as a call ID with other backends. As suspected, it causes issues with both OpenAI and Anthropic integrations due to the signature length and format.

Given that reliable persistence of the thoughtSignature requires a deeper integration than this PR currently provides, I’m going to close this for now.

PR #1155, opened 2025-11-24
gptel-gh: use the host dynamically

kiennq:
This allows to use the correct host url that's set in the token, ex: `api.enterprise.githubcopilot.com`

### DIFF
--- a/gptel-gh.el
+++ b/gptel-gh.el
@@ -319,18 +319,21 @@ Then we need a session token."
     (name &key curl-args request-params
           (header (lambda ()
                     (gptel--gh-auth)
-                    `(("openai-intent" . "conversation-panel")
-                      ("authorization" . ,(concat "Bearer "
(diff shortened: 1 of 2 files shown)


karthink:
@kiennq thanks for the PR.  Does the host url in the token change all the time, or is it otherwise unpredictable?  I'm asking since you're repeatedly setting `(gptel-backend-host gptel-backend)` (and the protocol) with every request in this PR.

kiennq:
It's depending on the git hub chat token, which will be refreshed time to time.  I think it's not going to change but is there issue with always setting those values?

karthink:
> It's depending on the git hub chat token, which will be refreshed time to time.

I'm surprised the host can change between token refreshes.  Is this really the case?  I would have assumed it would be enough to set it once from the token.

> I think it's not going to change but is there issue with always setting those values?

The header function is setting the backend host and protocol by side effect -- it just seems confusing and the wrong place to do it.

Issue #1158, opened 2025-11-25
Add a synchronized LLM gptel call function


tninja:
I'd like a synchronized version of LLM gptel call. It will be very useful to automate simple work with fast model such as gemini-2.5-flash, or gptel-4.1-mini. I have an initial version using current gptel function but doesn't looks very stable, especially for long input. I wish there is a better version.

@karthink would you mind take a look? Thanks in advance.

```
(defun ai-code-call-gptel-sync (question)
  "Get an answer from gptel synchronously for a given QUESTION.
This function blocks until a response is received or a timeout occurs.
Only works when gptel package is installed, otherwise shows error message."
  (unless (featurep 'gptel)
    (user-error "GPTel package is required for AI command generation. Please install gptel package"))
  (let ((answer nil)
        (done nil)
        (error-info nil)
        (start-time (float-time))
        (temp-buffer (generate-new-buffer " *gptel-sync*")))
    (unwind-protect
        (progn
          (gptel-request question
                         :buffer temp-buffer
                         :stream nil
                         :callback (lambda (response info)
                                     (cond
                                      ((stringp response)
                                       (setq answer response))
                                      ((eq response 'abort)
                                       (setq error-info "Request aborted."))
                                      (t
                                       (setq error-info (or (plist-get info :status) "Unknown error"))))
                                     (setq done t)))
          ;; Block until 'done' is true or timeout is reached
          (while (not done)
            (when (> (- (float-time) start-time) 30) ;; timeout after 30 seconds
              ;; Try to abort any running processes
              (gptel-abort temp-buffer)
              (setq done t
                    error-info (format "Request timed out after %d seconds" 30)))
            ;; Use sit-for to process events and allow interruption
            (sit-for 0.1)))
      ;; Clean up temp buffer
      (when (buffer-live-p temp-buffer)
        (kill-buffer temp-buffer)))
    (if error-info
        (error "ai-code-call-gptel-sync failed: %s" error-info)
      answer)))
```

karthink:
@tninja Yes, this is a good feature to have.  I've been meaning to add a `:sync` argument to `gptel-request`, but I don't know how to do it.  I'm including some notes here for the benefit of other readers.

----

For anyone that wants to help, please note that making `gptel-request` synchronous is trickier than just running `(while (accept-process-output process))`, for a few reasons:

1. `gptel-request` returns a finite state machine object, not the response.  This is not what you would expect from a synchronous API.  Instead of
   ```elisp
[...]

PR #1159, opened 2025-11-26
Add :sync argument to gptel-request for synchronous LLM requests (#3)

tninja:
@karthink PTAL Thanks

Addressing https://github.com/karthink/gptel/issues/1158

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,27 +1,34 @@
 # -*- mode: org; -*-
 
-* 0.9.9.5-pre
+* 0.9.9.4
 
 ** Breaking changes
 
-- The models =o3=, =o3-mini=, =o4-mini=, =claude-3.5-sonnet=,
-  =claude-3.7-sonnet=, =claude-3.7-sonnet-thought=, =claude-opus-4= and
-  =gemini-2.0-flash-001= have been removed from the default list of GitHub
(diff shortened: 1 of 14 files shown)


karthink:
@tninja Thanks for the PR.  Was this generated by Copilot, BTW?  This is a good start, but there are several problems and edge cases in the code, since many more parts of the process of sending requests can be asynchronous, such as context collection and the prompt transform functions.

tninja:
@karthink Thanks for the reply. Do you mean that context collection and the prompt transform functions are still asynchronous and need to be converted to synchronous versions? I assumed these two functions run locally and wouldn’t take much time, so using either sync or async might not make a significant difference at the moment. I also imagine that most of the time the code is actually waiting for the LLM to generate the full response.

I agree this solution is not perfect yet. The code was generated by Copilot after several iterations. I have tested it and it works on my side, e.g. `(gptel-request "What is 2+2?" :sync t)`. One possible, more formal improvement could be to set `stream=False` for the ChatGPT API so that the API itself performs a synchronous call. However, this would require investigation and code changes for different LLM vendors, and we would need to verify the behavior for each of them, which may take time.

Perhaps we can add a TODO comment in the code to note a better future version?

tninja:
Close this PR, new thread start at https://github.com/karthink/gptel/pull/1264

Issue #697, opened 2025-03-12
Support for OpenAI model response


ahmed-shariff:
This reminds me of: https://xkcd.com/927

Jokes aside it's adding two interesting things:
1. Allowing to refer to the previous responses: gptel already handles part of this, where it constructs the history by parsing text properties. Could this be extended to additionally store the response_id and use that instead of the response itself?
2. Built-in/hosted tools: I am not sure about the computer use, but the other two (web-search and file-search) could be framed as special tools? if web-search/file-search is selected as a tool, an additional transients are added to allow the user to modify settings that are specific to them. I've been playing around with some of this as part of the [openai-assistant backend](https://github.com/ahmed-shariff/gptel-openai-assistant) (which is being deprecated now) to query and display/update options when necessary, which could be used here as well.

There are also some similarities to features in other backends ([Antrhopic citations](https://github.com/karthink/gptel/issues/588), [Gemini grounding](https://github.com/karthink/gptel/issues/542#issuecomment-2693955384)) but I am not familiar enough about how they work to comment on them.

Details from: 
- https://platform.openai.com/docs/api-reference/responses
- https://x.com/OpenAIDevs/status/1899531225468969240

karthink:
The moment I read this announcement yesterday I knew it was going to be a bit of a mess.  The responses API is essentially taking on more of the responsibilities of the LLM client to make the developer's job easier, but this means the client has to support both server-side and local state, which is messier.

You also have two sources of truth for the state now.  Storing the resp_id is possible but the user's impression is that the buffer contents represent the state of the conversation, so if they edit the buffer which state do you pick?

I generally target the common denominator of the available APIs to keep the maintenance load under control.  I'm content to wait and see if the responses API enables anything different from the conversation API.  The special tools are available under different model names in the chat completions API.

Supporting the web search tool should be easy, but the file search tool is going to be a lot of work as it requires a whole subsystem for sending and tracking uploads.

karthink:
> Thats fair. Also, the response API doesn't seem to have the list of messages option as part of a request anymore. It's just one input string now.

This is actually how I imagined the chat completions API would work before it was released in 2023 and I was thinking of writing a package for LLM interaction.  I found the idea of sending the entire conversation every turn quite strange and wasteful.  But gptel's buffer-based paradigm is designed around the strengths of the chat completions API, so here we are.

> Well, I started playing around to see if I can get this to work.

Thanks for looking into it.  Let me know if you get somewhere interesting with it!

[...]

PR #1168, opened 2025-12-03
Add OpenAI Responses API support  (`/v1/responses`)

CsBigDataHub:
@karthink please consider this a initial draft, 

This PR extends `gptel-openai.el` to support OpenAI's new **Responses API**, an agentic primitive designed for stateful, multi-turn reasoning and server-side tool execution.

The implementation automates the switch between the standard **Chat Completions API** (`/v1/chat/completions`) and the **Responses API** (`/v1/responses`) based solely on the configured endpoint URL.

### Motivation

The Responses API introduces a different schema compared to Chat Completions:

  * **Input:** Uses `input` items instead of `messages`.
  * **Output:** Returns `output` items instead of `choices`.
  * **Streaming:** Emits distinct Server-Sent Events (SSE) for agentic steps.
  * **Tools:** Leverages server-side capabilities (e.g., `apply_patch`, `shell`) rather than client-defined schemas.

This PR enables users to access these agentic workflows (e.g., using `gpt-5.1-codex`) without breaking existing functionality.

### Implementation Details

To ensure **zero regressions** for existing users, I utilized `cl-defmethod :around` advice on the core backend methods.

1.  **Endpoint Detection:** Added `gptel--is-responses-api-p` to detect if the user's endpoint contains `/responses`.
2.  **Dispatch Strategy:**
      * The `:around` methods check the endpoint.
      * If it is a standard endpoint, `cl-call-next-method` is invoked, running the original, stable code unchanged.
      * If it is a Responses endpoint, the request is routed to new dedicated handlers (`gptel--request-data-responses`, `gptel--parse-response-responses`, etc.).
3.  **Tooling:** Added `gptel--parse-responses-tools` to request server-side capabilities (`apply_patch`, `shell`) instead of serializing client-side function definitions.

### Usage Example

Users can opt-in to the new API simply by changing the endpoint. No new configuration variables are required.

```elisp
;; Standard Chat (Unchanged)
(gptel-make-openai "OpenAI"
  :models '(gpt-4o))

;; New Agentic Workflow
(gptel-make-openai "OpenAI-Agent"
  :endpoint "/v1/responses"  ;; <--- Triggers new behavior automatically
  :models '(gpt-5.1-codex))
```

### Impact

  * **Breaking Changes:** None. The original code paths are preserved via `cl-call-next-method`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,7 +4,7 @@
 
 ** Breaking changes
 
-- The models =o3=, =o3-mini=, =o4-mini=, =claude-3.5-sonnet=,
+- The models =gpt-5-codex=, =o3=, =o3-mini=, =o4-mini=, =claude-3.5-sonnet=,
   =claude-3.7-sonnet=, =claude-3.7-sonnet-thought=, =claude-opus-4= and
   =gemini-2.0-flash-001= have been removed from the default list of GitHub
   Copilot models.  These models are no longer available in the GitHub Copilot
@@ -18,10 +18,12 @@
 
 ** New models and backends
 
-- GitHub Copilot backend: Add support for =gpt-41-copilot= and
-  =gemini-3-pro-preview=
+- GitHub Copilot backend: Add support for =gpt-5.2=, =gpt-41-copilot=,
+  =claude-opus-4.5=, =gemini-3-pro-preview= and =gemini-3-flash-preview=
 
-- Add support for =gemini-3-pro-preview=.
+- Bedrock backend: Add support for =claude-opus-4.5= and =nova-2-lite=
+
+- Add support for =gemini-3-pro-preview=, =gemini-3-flash-preview=.
 
 - Add support for =gpt-5.1=.
 
@@ -38,6 +40,10 @@
   ring redirection now correctly captures the full response from the
   LLM, including pre- and post-tool-call text.
 
+- =gptel-rewrite= now supports tool calling.  If =gptel-tools= is
+  non-nil the LLM can, for instance, read files to fetch more context
+  for the rewrite action.
+
 ** Notable bug fixes
 
 - Function-valued system messages/directives are now evaluated in the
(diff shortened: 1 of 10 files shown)


karthink:
@CsBigDataHub thanks for the PR.  I'll take a look when I have time -- this one looks AI generated, which means it's a lot more work for me to vet.

But please note that the problem with adding the Responses API to gptel is not the implementation at the backend level, which this PR addresses.  I could have added that at any time.

The problem is that gptel's model is stateless but the Responses API is not, and it violates the fundamental interaction paradigm presented to the user.  See prior discussion in #697.

Inkbottle007:
> The problem is that gptel's model is stateless but the Responses API is not, and it violates the fundamental interaction paradigm presented to the user. See prior discussion in #697.

We can switch to the Responses API without changing gptel’s stateless model:

- We won’t rely on the server-side state provided by the Responses API (so no need to use or manage resp_* state).
- We’ll simply ignore the returned `"id": "resp_087eaexxx"` and keep sending the full context with each request.

Note: the Completions API seems deprecated; the Responses API can do everything the legacy API did. The documentation for the Completions API is no longer maintained on the OpenAI site.

karthink:
> - We won’t rely on the server-side state provided by the Responses API (so no need to use or manage resp_* state).
> - We’ll simply ignore the returned `"id": "resp_087eaexxx"` and keep sending the full context with each request.

The last time I checked, there was no way to specify user and model roles when sending text using the Responses API, so this doesn't work.  If I missed it, please provide a link.

karthink:
Ah, never mind, it looks like they added it at some point:

> Input message
> object
>
> A message input to the model with a role indicating instruction following hierarchy. Instructions given with the developer or system role take precedence over instructions given with the user role. Messages with the assistant role are presumed to have been generated by the model in previous interactions.

From https://platform.openai.com/docs/api-reference/responses/create, inside one of the dropdowns.

This wasn't there (or wasn't documented) earlier this year.

So we can support the responses API.  Thanks for bringing it to my attention.

karthink:
Responses API support added in PR #1292, closing.

API tools are not supported yet.  This is a gptel-wide effort that I'll address separately.

PR #1170, opened 2025-12-05
README: Add custom fn to read API keys from ENV Vars

vedang:
Adds an example function to the `Additional Configuration` section to show how users can reuse any API keys they may have already defined as ENV vars.

### DIFF
--- a/README.org
+++ b/README.org
@@ -68,6 +68,7 @@ gptel uses Curl if available, but falls back to the built-in url-retrieve to wor
     - [[#chatgpt][ChatGPT]]
     - [[#other-llm-backends][Other LLM backends]]
       - [[#optional-securing-api-keys-with-authinfo][(Optional) Securing API keys with =authinfo=]]
+      - [[#optional-reading-api-keys-from-environment-variables][(Optional) Reading API keys from environment variables]]
       - [[#azure][Azure]]
       - [[#gpt4all][GPT4All]]
       - [[#ollama][Ollama]]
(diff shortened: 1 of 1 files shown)


karthink:
@vedang Thanks for the PR.  I don't think most users will have more than 2-3 backends defined, so I think the instructions can be much simpler, and a function that tries to cover all cases (unsuccessfully, because it hard-codes the names) isn't required.

Instead, how about a section next to the `(Optional) Securing API keys with =authinfo=` section that explains the general idea briefly?

Something like:

``` text
**** (Optional) Reading API keys from environment variables

Alternatively, you can fetch API keys from variables in Emacs' environment (such
as =OPENAI_API_KEY=) using a function like this:

#+begin_src emacs-lisp
(defun gptel-api-key-from-environment (&optional var)
  (lambda ()
    (getenv (or var                     ;provided key
                (thread-first           ;or fall back to <TYPE>_API_KEY
                  (type-of gptel-backend)
                  (symbol-name)
                  (substring 6)
                  (upcase)
                  (concat "_API_KEY"))))))
#+end_src

Then set the =:key= when defining backend to the appropriate environment
variable:

#+begin_src emacs-lisp
(gptel-make-anthropic "My-Claude-backend"
  :key (gptel-api-key-from-environment "ANTHROPIC_API_KEY"))
#+end_src

If no environment variable is prvided:

#+begin_src emacs-lisp
(gptel-make-gemini "My-Gemini-backend"
  :key (gptel-api-key-from-environment))
#+end_src

it will try to guess the key from the backend type, /i.e./ try to use
=GEMINI_API_KEY= when using the Gemini backend, and so on.
```

vedang:
Perfect, this is a much cleaner approach. I'll update the PR

vedang:
I've updated the PR and modified my local setup to work with the new function. Everything looks good to me.

karthink:
Looks good except for a typo, could you update that?

vedang:
Sorry for the silly mistake. I've updated the PR.

karthink:
Thank you for the PR @vedang!

PR #1171, opened 2025-12-05
Fix context order when confirming changes in context buffer

aagit:
When confirming changes in the gptel context buffer, the context list was being reversed due to how it was collected and stored.


karthink:
@aagit Thanks for the PR!  Merged with reworded commit.

PR #1177, opened 2025-12-12
Move post-request hook before sentinel creation

aagit:
The gptel-post-request-hook is run after the process sentinel has been created. If this hook aborts a running background request (e.g., a kvcache preloading task) using gptel-abort, it can interfere with the new request's setup, causing a "Search failed" error when the new request tries to access the FSM state.

This change ensures the post-request hook runs before the process sentinel is created, preventing interference between concurrent requests and avoiding the "Search failed" error when aborting background tasks.

The error message observed was:

```
------  kvcache speculative prelaoding started by timer ------
Querying ryzen...
Wrote /tmp/gptel-curl-datadgs2S0.json
------  gptel-post-request-hook calls gptel-abort to stop preloading ------
Stopped gptel request in buffer "*gptel-context-preload*"
------  real request fails to start ------
Querying ryzen...
error in process sentinel: save-current-buffer: Search failed: "79f42ff48ff153e47dc3d3dba8020dc6"
error in process sentinel: Search failed: "79f42ff48ff153e47dc3d3dba8020dc6"
```

This is only observed if running a gptel-request in a gptel mode buffer, does not happen if it's gptel-rewrite issuing the gptel-request.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -1657,6 +1657,7 @@ MACHINE is an instance of `gptel-fsm'"
   ;; Reset some flags in info.  This is necessary when reusing fsm's context for
   ;; a second network request: gptel tests for the presence of these flags to
   ;; handle state transitions.  (NOTE: Don't add :token to this.)
+  (run-hooks 'gptel-post-request-hook)
   (let ((info (gptel-fsm-info fsm)))
     (dolist (key '(:tool-success :tool-use :error :http-status :reasoning))
       (when (plist-get info key)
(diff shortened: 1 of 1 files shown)


aagit:
After reading your answer about how to inject the per rg-edit grammar, I wondered if I could use the gptel-prompt-transform-functions instead of gptel-post-request-hook and it works. So I worked around this issue by switching hook.

karthink:
> So I worked around this issue by switching hook.

@aagit Do you still want to pursue this PR?  Note that there are some subtleties involved in the placement of this hook w.r.t. sentinel creation that will need to be investigated carefully.

aagit:
My workflow doesn't require this one anymore as I shifted hook successfully. It's up to you judje where's the best place to run those hooks. What I found is the place where it's run now breaks if gptel-abort is called by the hook.

PR #1185, opened 2025-12-16
Fix gptel-context-remove-all message

Sonnicon:
Stop `gptel-context-remove-all` success message from being displayed when confirmation is rejected.


karthink:
Thank you @Sonnicon!

PR #1188, opened 2025-12-18
Ensure `setter` propagates to parent presets

kmontag:
This fixes an issue where, when e.g. applying a preset buffer-locally, any `:parents` of the preset would still be applied using the default (non-buffer-local) setter.


karthink:
Thank you @kmontag!

PR #1198, opened 2025-12-30
gptel-bedrock: Add claude-opus-4-5-20251101

leezu:


karthink:
@leezu Thank you for the PR.  Could you also add a NEWS entry in the "New Models and Backends" section?  See the one that says `GitHub Copilot backend: ...` for a reference.

leezu:
@karthink of course. Updated the NEWS just now

PR #1201, opened 2026-01-04
Correctly parse presets with trailing newline.

lynaghk:
`:blank:` doesn't match newlines, so an invocation like:

```
@preset

my prompt
```

was parsing the potential preset as "@preset\n\nmy" rather than "@preset".


lynaghk:
p.s. I absolutely love gptel and it's one of the main reasons that after 15+ years of using Emacs I'm now happily coding up my own customizations. If you're ever around Amsterdam, drop me a line so I can buy you a coffee =D

karthink:
@lynaghk Thank you for catching that.  I've experienced this bug before, but only exactly when I'm on a deadline and can't stop to investigate!

Issue #870, opened 2025-05-23
Cursor position unexpectedly changes after gptel-rewrite when scroll-preserve-screen-position is set to 'always


jamescherti:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

When using GPTel in Emacs with the variable `scroll-preserve-screen-position` set to `'always`, invoking the `gptel-rewrite` refactor results in an unexpected change in the cursor (point) position. This contradicts the expected behavior of the `scroll-preserve-screen-position` setting, which is intended to maintain the cursor's screen-relative position after buffer changes.

**Expected Behavior:**

The cursor should retain its position when `scroll-preserve-screen-position` is set to `'always`.

**Impact:**

This behavior disrupts the editing workflow, especially in long buffers or when precise cursor control is required, and undermines the expected consistency of Emacs' screen positioning behavior.

### Backend

None

### Steps to Reproduce

1. Set `scroll-preserve-screen-position` as follows:
```emacs-lisp
(setq scroll-preserve-screen-position 'always)
```

2. Select text.
3. The user invokes the `M-x gptel-rewrite` then press `r`.
4. After the GPTel refactoring completes, the cursor (point) moves from its original position.

### Additional Context

Emacs 30.1.90
OS: Arch Linux

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

dschrempf:
I also observe the cursors jumping around when rewriting text. The rewritten text is also mangled and misses substantial parts. Could it be that the major mode is interfering with the pasting process of `gptel`?

dschrempf:
Maybe related to streaming (see https://github.com/karthink/gptel/pull/850).

Without streaming, the complete rewrite text is inserted, but the cursor is still jumping around.

jamescherti:
The `gptel-rewrite` package misplaces the cursor only during **linewise selection**. For other types of selection, such as point-to-point ranges, the cursor is positioned correctly.

jamescherti:
> I also observe the cursors jumping around when rewriting text. The rewritten text is also mangled and misses substantial parts. Could it be that the major mode is interfering with the pasting process of `gptel`?

It is not caused by the major mode; rather. I tested multiple major modes, including `fundamental-mode`, and the problem persists in all cases. The `gptel-rewrite` package misplaces the cursor during **linewise selection**.

dschrempf:
[...]

Issue #564, opened 2025-01-12
post response for gptel-rewrite gets the same position for beg and end


ping13:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [X] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

The bug occurs when attempting using `gptel-rewrite` with a defined hook in `gpterl-post-response-functions`: the "begin" and "end" positions for the functions are identical (and supposedly the position of the cursor). 

Expected outcome would be the beginning and end positions for the marked region. Note that the request has not failed, otherwise this behavior is expected as written in the documentation

### Backend

OpenAI/Azure

### Steps to Reproduce

1. add a function to the hook that writes parameters to the `*Messages*` buffer

```elisp
(add-hook 'gptel-post-response-functions '(lambda (beg end) 
                                              (message "Post response %s %s!" beg end)))
```

2. Mark some text and call `gptel-rewrite`
3. Observe a similar content in the `*Messages*` buffer with the beginning and end positions being the same.

```
[...]
Mark set
LLM rewrite output ready: <mouse-1>, RET or M-x gptel-rewrite to continue.
Post response 2964 2964!
Action:  (accept, k reject, iterate, merge, diff, ediff, ?): 
[....]
```

### Additional Context

Emacs version: GNU Emacs 29.4 (build 1, aarch64-apple-darwin21.6.0, NS appkit-2113.60 Version

### Backtrace

```emacs-lisp
n/a
```

karthink:
> The bug occurs when attempting using `gptel-rewrite` with a defined hook in
> `gpterl-post-response-functions`: the "begin" and "end" positions for the
> functions are identical (and supposedly the position of the cursor).
> 
> Expected outcome would be the beginning and end positions for the marked
> region. Note that the request has not failed, otherwise this behavior is
> expected as written in the documentation

Thanks for the report.  I'm not sure that the correct result would be the beginning and end positions of the marked region.  As stated in the documentation:

``` text
This hook is called in the buffer to which the LLM response is
sent, and after the full response has been inserted.  Each
function is called with two arguments: the response beginning and
end positions.
```

In this case the response has not been inserted into the buffer at all -- the buffer is not modified until you accept the rewrite.  So the hook is called with both beginning and end set to where the request was fired from.  This is similar to the behavior when the response fails: the post-response-functions are still called for possible clean-up, but the positions have no special meaning.

[...]

PR #1205, opened 2026-01-08
gptel-rewrite: Fix issues with evil's visual selection mode

### LINKED ISSUE #1062: Rewrite disfunctional when selecting only a single line

- [ ] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

In some circumstances rewrite does not work anymore in two ways:
- the transient buffer fails to show some parts
- while the API request corresponding to the execute gptel-rewrite command is being executed (e.g. curl is running in the background), the new content does not make it into the buffer in any way (it is ignored in some sense)

I observed this for anthropic and for openai backends.

### Backend

Anthropic

### Steps to Reproduce

1. Select/Mark multiple lines
3. `M-x gptel` -> Press 'r' for rewrite -> Press 'M-RET'
4. *Everything looks good* 

Image

5.  Select a single line (for org mode buffers, a folded heading (incorporating multiple lines) will also have this issue if selected as single line)
6. repeat `M-x gptel` -> Press 'r' for rewrite -> Press 'M-RET'
7. The lower part of the rewrite transient window is missing. 
Image

### Additional Context

```
Gptel version:
     Status: Installed in ‘gptel-20250831.846/’ (unsigned). Delete
    Version: 20250831.846
     Commit: a2fe4ec752f3859db7d665d03a4c874cfad1e7e2
    Summary: Interact with ChatGPT or other LLMs.
   Requires: emacs-27.1, transient-0.7.4, compat-30.1.0.0
    Website: https://github.com/karthink/gptel
```

OS: NixOS (up2date)
Emacs version: 30.2 (spacemacs 0.999.0)

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

real-or-random:
Fixes #1062. See commit messages for details.

There are more uses of `(run-at-time 0 ...)` in the code base, but my understanding is that the code they invoke (currently) doesn't call `(region-beginning)` or `(region-end)`, so there's no bad interaction with evil. But perhaps it's a good idea to fix these instances as well to avoid potential future issues, possibly also with other uses of `post-command-hook`.

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -743,7 +743,7 @@ generated from functions."
       ;; Move back so that the cursor is on the overlay when done.
       (unless (get-char-property (point) 'gptel-rewrite)
         (when (= (point) (region-end)) (run-at-time 0 nil #'backward-char 1)))
-      (deactivate-mark))))
+      (setq deactivate-mark t))))
 
 ;; Allow this to be called non-interactively for dry runs
 (put 'gptel--suffix-rewrite 'interactive-only nil)


karthink:
Thanks for the PR @real-or-random!  I'll have to look at the code carefully to figure out why I used `(run-at-time 0 ...`  instead of just calling `transient-setup` in the first place.  I don't usually schedule things on the event loop awkwardly like this unless it's avoiding some subtle problem, but it's possible I made a mistake or was overly conservative here.

real-or-random:
> Thanks for the PR @real-or-random! I'll have to look at the code carefully to figure out why I used `(run-at-time 0 ...` instead of just calling `transient-setup` in the first place. I don't usually schedule things on the event loop awkwardly like this unless it's avoiding some subtle problem, but it's possible I made a mistake or was overly conservative here.

My assumption is simply that it's used to execute code after `(exit-minibuffer)` which isn't possible without some tricks (https://emacs.stackexchange.com/a/20988 for those unaware of the reasons).

I believe the crucial part to fix the evil stuff here is to not use `run-at-time` for `(gptel--suffix-rewrite)`. That is, I believe we could keep `run-at-time` it for `transient-setup` if that's desired. (But I still think it's cleaner to replace it.)  

In any case, we cannot keep the entire user interaction with the transient menu inside one editor command.  In fact, that is the reason why this PR here doesn't fix #909 entirely. There's still one cosmetic issue left: When you use evil visual line selection to, say, lines 4-6, and then invoke `gptel-menu`, the status line will claim that you selected 4-7 because evil moves the point after the newline on line 6 and onto the first char of line 7. When you now interact with the transient menu, this is changed to 4-6 because evil moves the point back... This is purely cosmetic because `gptel-rewrite` doesn't really care whether the trailing newline is selected or not. It will work in either case. Apparently, evil has some functionality for always excluding the newline if desired (see variable `evil-visual-newline-commands`.) I see if this can be used to fix this cosmetic issue, but let's postpone this to a follow-up PR.

One more observation: I don't think we'll need the `(with-current-buffer (window-buffer (minibuffer-selected-window)) ...` stuff for `transient-setup`. This should really only be required for `(gptel--suffix-rewrite)`. I had just applied it mechanically without thinking about it. I'll revise.

karthink:
@real-or-random Thanks for the clarification.  After you're done revising, I'll test it with and without evil-mode and hopefully merge it.

real-or-random:
Okay, updated. Ready from my side.

Changes:
  - I use the existing `cb` variable for the current buffer instead of `(window-buffer (minibuffer-selected-window)`
  - This now removes `run-at-time` only for `gptel--suffix-rewrite` where it's required to fix the evil issues.
  - So while the entire `gptel--suffix-rewrite` is no longer scheduled in the event loop now, I needed to add/keep a `run-at-time` for the logic within `gptel--suffix-rewrite` that moves the point back one position to make sure it will end up on the overlay. Without this hack, there's something that always moves the point back (i.e., forward by 1 again) when *not* in evil-mode. I really have no idea what this something is (I tried to debug this further but gave up), but I suspect it could have been the real reason why you resorted to `run-at-time` here in the first place.

More remarks:
  - I sometimes see a "display bug" where the cursor appears to jump to a wrong position (last char on first line) when the overlay is inserted. But actually, the point is at the right position on the overlay (which can be verified by evaluating `(point)`) and `RET` works as intended. This drove me crazy when testing the changes here. (Now I see that you're aware of that one; you mentioned it [here](https://github.com/karthink/gptel/issues/564#issuecomment-2625002360).)
  - I don't claim any longer that this fixes #870. The second commit fixes the cursor position in evil after issuing the rewrite call on a "visual-line" selection. But this is not #870: That one is about the cursor placement after the overlay has been inserted (IIUC), and it is reproducible even without evil-mode according to the reporter. So this must be something else ultimately. Perhaps just the aforementioned display bug...?
  - I don't claim any longer that this fixes #909. That one is more complex. I'll post there.

karthink:
@real-or-random Thank you for the PR and for the careful testing!

The overlay display bug you ran into is in Emacs' display engine and not fixable in gptel, sorry about that!  If you can create a recipe to reproduce it without gptel, we can submit a bug upstream.  I haven't been able to.

PR #1208, opened 2026-01-09
Fix for github copilot authentication when behind a strict network proxy.

skittishdev:
This also aligns with [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.1-6) which strongly discourages sending content with a GET request.

```
A client SHOULD NOT generate content in a GET request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported. An origin server SHOULD NOT rely on private agreements to receive content, since participants in HTTP communication are often unaware of intermediaries along the request chain.
```

* How was this problem discovered? A strict network proxy was returning an error because the HTTP GET request has a body. This meant that we were unable to acquire a session token when trying to use Github Copilot.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -884,8 +884,8 @@ MODE-SYM is typically a major-mode symbol."
 (cl-defun gptel--url-retrieve (url &key method data headers)
   "Retrieve URL synchronously with METHOD, DATA and HEADERS."
   (declare (indent 1))
-  (let ((url-request-method (if (eq method'post) "POST" "GET"))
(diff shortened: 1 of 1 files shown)


karthink:
@skittishdev Thanks for the PR.  There might be endpoints `gptel--url-retrieve` accesses for auth that expect a GET request with some data, I haven't checked.  If that is the case we'll find out soon as something will break.

skittishdev:
Thanks. When I grepped, I saw that this method was only used in gptel-gh.el, and there was only 1 GET request.

PR #1218, opened 2026-01-18
gptel-openai: support responses API

kiennq:
I must say the AI generated this PR quite impressive.
Since the other PR #1168 doesn't work for me when trying with `gptel-gh`. I've asked opus model to create this PR. It works quite well.
Tried with `(setq gptel-backend (gptel-make-gh-copilot "Copilot" :responses-endpoint "/responses"))` and I can use with the `codex` model that only accepts `responses` endpoint.

For the summary:
This pull request adds support for OpenAI's Responses API as an alternative to the Chat Completions API in `gptel`, allowing access to server-side tools like web search and code interpreter. It introduces new backend configuration options and updates model capabilities to enable this functionality.

**OpenAI Responses API Integration:**

* Added documentation in `README.org` explaining how to use the OpenAI Responses API with `gptel`, including configuration examples and available built-in tools.
* Updated the `gptel-make-openai` and `gptel-make-gh-backend` functions in `gptel-gh.el` to support a new `:responses-endpoint` parameter for specifying the Responses API endpoint and a `:builtin-tools` parameter for enabling server-side tools. [[1]](diffhunk://#diff-f7b546d19260380655a444a8c2bd64627fab8812695f85a5952c2455b607a852R362-R363) [[2]](diffhunk://#diff-f7b546d19260380655a444a8c2bd64627fab8812695f85a5952c2455b607a852R405-R414)
* Modified backend URL resolution logic to route requests to the Responses API endpoint for models with the `responses-api` capability.

**Model Capabilities Update:**

* Added the `responses-api` capability to the `gpt-5.1-codex-max` model in `gptel-gh.el` to enable use of the Responses API when available.

### DIFF
--- a/README.org
+++ b/README.org
@@ -230,6 +230,34 @@ Optional: Set =gptel-api-key= to the key. Alternatively, you may choose a more s
 - Setting it to a custom function that returns the key.
 - Leaving it set to the default =gptel-api-key-from-auth-source= function which reads keys from =~/.authinfo=. (See [[#optional-securing-api-keys-with-authinfo][authinfo details]])
 
+**** (Optional) Using the OpenAI Responses API
+
+gptel supports OpenAI's [[https://platform.openai.com/docs/api-reference/responses][Responses API]] as an alternative to the Chat Completions API.  The Responses API provides access to server-side built-in tools like web search, code interpreter, and file search.
+
+To use the Responses API, create a custom OpenAI backend with the =:responses-endpoint= parameter and add the =responses-api= capability to models:
+
+#+begin_src emacs-lisp
+(gptel-make-openai "OpenAI-Responses"
+  :key #'gptel-api-key
+  :stream t
+  :responses-endpoint "/v1/responses"
+  :builtin-tools '((:type "web_search")       ;; Enable web search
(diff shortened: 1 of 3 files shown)


ahmed-shariff:
Hey, this is pretty cool. Very similar to the custom backend I had written (https://github.com/ahmed-shariff/gptel-openai-responses-backend) - I had taken a different approach to the built-in tools - where I can add/remove them with gptel-menu.

karthink:
@kiennq I looked at this implementation, and I don't like overloading the `gptel-openai` methods with the check for responses-endpoint/responses API check.  I think it makes the code messier and harder to maintain.

Ideally I would like to define a separate `gptel-openai-responses` backend, but as you point out this won't work if certain models in the same backend require the responses API and others require the completions API.

Can the Copilot API use the responses endpoint for all its models?  If so, it would be cleanest to derive `gptel--gh-copilot` from `gptel-openai-responses` instead of `gptel-openai`.

If it uses both depending on the model, I'll have to think more about how best to add responses API support to gptel.

karthink:
Is there a website with the official Copilot-chat REST API documentation?  I couldn't find one.

karthink:
Responses API support added in PR #1292, closing.

PR #1236, opened 2026-02-06
gptel-transient: Fix display issues with evil visual selections

### LINKED ISSUE #909: When using evil mode, "Send selection" goes away in transient menu on any menu choices

real-or-random:
Fixes #909 and a few other issues, e.g., missing "Rewrite r" choices in the transient menu when there's an evil visual selection.

I played around with multiple approaches, but I think this is by far the cleanest fix. First, it kills the problem entirely, i.e., we won't need to change every `(use-region-p)`, `(region-beginning)`, etc. responsible for displaying a transient, which would be very annoying to maintain. Second, this approach is not very hacky because it uses a mechanism offered by transient for setting up an "environment" in which the display code runs.

We'll need transient 0.7.8. I think it's fine to bump it -- the release dates differ just by a few months -- but we can also make the code conditional in the version if you prefer that approach. 

I couldn't notice any difference in performance, and performance should anyway be totally fine: 
 - The first thing `evil-visual-expand-region` does is to check if there's a visual selection, i.e., if there's anything to do at all. If not, it returns immediately. (Note that these evil functions are performance-critical in evil-mode itself because they run in the pre/post-command-hook.)
 - The code is skipped if the region is already expanded.
 - The code is skipped entirely if evil-mode is not loaded.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -4,7 +4,7 @@
 
 ;; Author: Karthik Chikmagalur <karthik.chikmagalur@gmail.com>
 ;; Version: 0.9.9.5
-;; Package-Requires: ((emacs "27.1") (transient "0.7.4") (compat "30.1.0.0"))
+;; Package-Requires: ((emacs "27.1") (transient "0.7.8") (compat "30.1.0.0"))
 ;; Keywords: convenience, tools
 ;; URL: https://github.com/karthink/gptel
 
--- a/NEWS
+++ b/NEWS
@@ -220,6 +220,9 @@
 * 0.9.9.4 2026-02-21
 
 ** Breaking changes
+- gptel now requires Transient 0.7.8 or higher. Transient is a built-in package
+  and Emacs does not update it by default.  Ensure that
+  =package-install-upgrade-built-in= is non-nil, or update Transient manually.
 
 - The models =gpt-5-codex=, =o3=, =o3-mini=, =o4-mini=,
   =claude-3.5-sonnet=, =claude-3.7-sonnet=, =claude-3.7-sonnet-thought=,
--- a/README.org
+++ b/README.org
@@ -135,7 +135,7 @@ gptel uses Curl if available, but falls back to the built-in url-retrieve to wor
 
 ** Installation
 
-Note: gptel requires Transient 0.7.4 or higher.  Transient is a built-in package and Emacs does not update it by default.  Ensure that =package-install-upgrade-built-in= is true, or update Transient manually.
+Note: gptel requires Transient 0.7.8 or higher.  Transient is a built-in package and Emacs does not update it by default.  Ensure that =package-install-upgrade-built-in= is true, or update Transient manually.
 
 - *Release version*: =M-x package-install= ⏎ =gptel= in Emacs.
 - *Development snapshot*: Add MELPA or NonGNU-devel ELPA to your list of package sources, then install with =M-x package-install= ⏎ =gptel=.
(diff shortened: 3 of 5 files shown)


karthink:
This looks great, thanks for the PR!  I'll merge it soon.

real-or-random:
Force-pushed to fix a typo in the docstring and to rebase on master

real-or-random:
@karthink Friendly ping.

> This looks great, thanks for the PR! I'll merge it soon.

As I said in my last comment in #1235, this PR here is independent of #1235 and should be merged in any case. And then we can see what to do with #1235.

karthink:
@real-or-random Could you bump the transient requirement in gptel's package header to 0.7.8?

gptel.el
```diff
@@ -4,7 +4,7 @@
 
 ;; Author: Karthik Chikmagalur <karthik.chikmagalur@gmail.com>
 ;; Version: 0.9.9.4
-;; Package-Requires: ((emacs "27.1") (transient "0.7.4") (compat "30.1.0.0"))
+;; Package-Requires: ((emacs "27.1") (transient "0.7.8") (compat "30.1.0.0"))
```
real-or-random, line 7: This PR does this already here. (Or are you referring to something else?)

real-or-random (commented):
Anyway, rebased on master. :)

karthink:
@real-or-random Could you update this branch?  I got some merge conflicts when I tried to rebase it on master.

real-or-random:
> @real-or-random Could you update this branch? I got some merge conflicts when I tried to rebase it on master.

I'm pretty sure that there must have been something wrong or unusual or your side. The branch was indeed 10 commits behind master, but it could be merged cleanly accordingly to GitHub, and my manual rebase on master worked without any conflicts now. Perhaps you were still looking at the very first version of this PR (https://github.com/karthink/gptel/commit/8ae1d51a9cab52ff8b6afb55d4207bf2a90790be)? This would also explain why you missed that I had already updated the package header for transient.

Anyway, I force-pushed to rebase on master (at d7c103ce8a66ecaee7c8fc9ecadb681dc4291408) again. Please check you have the right PR branch. It's currently at https://github.com/karthink/gptel/commit/35ef37870fe74466da9a32a2526417c77a5e5b6b.

karthink:
@real-or-random Merged, thank you for the PR!  Changed the docstring of the fix function to mention the `:environment` slot instead of `:entry` (assuming that was a typo).

Does this `:environment` fix also need to be applied to the `gptel-rewrite` transient?

real-or-random:
> @real-or-random Merged, thank you for the PR! Changed the docstring of the fix function to mention the `:environment` slot instead of `:entry` (assuming that was a typo).

Thanks. Yes, this was a typo, of course.

 
> 
> Does this `:environment` fix also need to be applied to the `gptel-rewrite` transient?

Yes, `gptel-rewrite` also needs it. But the second commit here already added it to `gptel-rewrite`, see https://github.com/karthink/gptel/pull/1236/changes#diff-fbd38c30d1c347f4f98f0a57f95bb118228ae68a90f3d646621a494736c9ac6a. I assumed you missed that line in the code review. So this has been taken of already in this PR.

karthink:
> Yes, `gptel-rewrite` also needs it. But the second commit here already added it to `gptel-rewrite`, see https://github.com/karthink/gptel/pull/1236/changes#diff-fbd38c30d1c347f4f98f0a57f95bb118228ae68a90f3d646621a494736c9ac6a. I assumed you missed that line in the code review. So this has been taken of already in this PR.

Thanks, I did indeed miss that!

PR #1238, opened 2026-02-07
gptel-transient: fix broken compatibility with Transient 0.12+

budevg:
Transient 0.12+ may pass EIEIO suffix/group objects to the description function,
which broke gptel-transient's assumptions.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -306,6 +306,9 @@ documention.  Return nil if user does not provide a number, for default."
 
 Handle formatting for system messages when the active
 `gptel-model' does not support system messages."
+  ;; Transient >= 0.12 may call description functions with the suffix/group
+  ;; object as an argument. Treat that as "no explicit message supplied".
+  (if (eieio-object-p message) (setq message nil))
   (setq message (or message gptel--system-message))
   (if (gptel--model-capable-p 'nosystem)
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR.

Can you direct me to the place in the transient documentation that talks about this?  I'm running an older transient version, I looked at some pages in https://docs.magit.vc/transient/ but couldn't find any mention of this.  I expected to find it at https://docs.magit.vc/transient/Group-Specifications.html.

BTW, `transient` `require`s `eieio`, which `require`s `eieio-core`, which includes `eieio-object-p` -- so there is no need for the `fboundp` check.

budevg:
Actually it was always passing the object as parameter. I guess this commit https://github.com/magit/transient/commit/6d1d865666c5333b4a342ebfe6de835718410478 introduced regression where the number of parameters in `gptel-system-prompt--format` is  deduced to 1 and then the extra object is passed. I also didn't find mention of this in transient docs just in the code.

budevg:
Also, I removed the `fboundp` check.

karthink:
I fixed it in a slightly simpler way in 8b9114d72088a3a3cfec825f00b76ad82e4d617e, so I'm closing this PR.  Thanks for bringing this problem to my attention!

PR #1242, opened 2026-02-12
deduplication of tools

### LINKED ISSUE #1097: Handling of duplicate mcp-derived tool names
Forgive me if this has already been discussed, I did not see it in my search of the repo. Please let me know if further details are required.

When working with third party MCP implementations, there are sometimes duplicates. `search` is a good example. It seems that the tool selection does not handle this properly - for example when selecting the `search` tool from the `foo` mcp, `search` will also show as active in the `bar` mcp - and the user is uanble to deactivate it.

I wonder if it makes sense to _always_ force a prefix for MCP derived tools? Perhaps this could be configurable.

otakup0pe:
This fixes #1097 through the following mechanisms. Ended up not splitting this into two PRs as the functionality seemed kinda intertwined, I hope that is :ok:

* (configurable) visual indicators in transient menu of duplicate tools
* runtime resolution of duplicate tools for presets and chat mode

This PR was created with LLM assistance. I have been running it for about six weeks and it's been behaving solidly during that time.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -37,50 +37,7 @@
 
 ;; * Helper functions and vars
 
-(defgroup gptel-transient-ui nil
-  "UI options for gptel's transient menus."
-  :group 'gptel)
 
-(defcustom gptel-transient-disambiguate-tools 'off
-  "Highlight duplicated tool names in `gptel-tools'.
-
-Values:
-  on  Highlight tools and categories involved in duplicate tool names.
-  off Do not highlight duplicates.
-
-When highlighting is on: - Category entries that contain one or more
-duplicated tool names display a warning symbol plus the count (⚠N) after
-the category name, where N is the number of duplicated tool names in
-that category.  - Each duplicated tool name within a category also
-displays a warning symbol (⚠) after the tool name.
-
-No other disambiguation text (tags, provider lists) is shown.
-
-Changing this requires reopening the transient to take effect."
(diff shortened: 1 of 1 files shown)


karthink:
@otakup0pe This is a much bigger change than I thought it would be, and it looks like the LLM has thrown away some of my comments.  If you can reduce the scope of the PR and not discard my comments -- which are essential for future understanding of the code (by me and others), I will have an easier time reviewing this.

otakup0pe:
Oh my gosh sorry. I must have misread reviewing the computer hallucination before getting on a flight - it seemed the comments no longer applied given the proposed implementation!

Do you have any preference on how I shrink the scope? if you are not tied to the "disambiguation" part of the PR (showing which tools are duplicates in the transient view) I might just drop that part completely. During my use of this PR it has been less useful than I thought, and the de-duplication could stand alone.

otakup0pe:
OK I removed some of the transient ux changes - just kept the bit that let's the user (de) select duplicate tools across categories. In the end, the warnings/counts ended up not really being needed outside of troubleshooting. Please let me know if you would like any other changes, or any [other tests](https://github.com/karthink/gptel-test/pull/12).

karthink:
Thanks.  This PR is still on my radar, just busy with some other features atm.

otakup0pe:
Oh hey thanks for the bump! I'll update from mainline branch, and make sure the tests are updated 💖

PR #1243, opened 2026-02-12
gptel-anthropic: gptel-bedrock: support for claude-opus-4.6

surenkov:
According to:
- https://platform.claude.com/docs/en/about-claude/models/overview#latest-models-comparison
- https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html


karthink:
@surenkov Thanks for the update!

PR #1250, opened 2026-02-16
README: Expand Doom Emacs section

mandos:
Add info about native Doom Emacs module, which use gptel as main driver.

If this is too much, and basic instalation is enough, don't bother and just reject PR.


karthink:
This is great, thanks for the PR @mandos!  Merged.

PR #1251, opened 2026-02-16
feat: add customizable AWS CLI command path

liangfu:
### Motivation
Users may need to:
- Use a specific AWS CLI installation (e.g., `/usr/local/bin/aws`, `/opt/homebrew/bin/aws`)
- Handle environments where the AWS CLI is not in the default `$PATH`
- Manage multiple AWS CLI versions for different projects

#### Default behavior (no configuration needed)
By default, gptel will continue to use the `aws` command from your `$PATH`:
```elisp
;; No configuration needed - uses "aws" from PATH
```

#### Custom AWS CLI path
To specify a custom AWS CLI installation path:
```elisp
(setq gptel-bedrock-aws-cli-command "/usr/local/bin/aws")
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -56,6 +56,9 @@
   results, block the tool call but continue the request, or stop the
   request entirely.
 
+- New variable ~gptel-bedrock-aws-cli-command~ to set the path to the
+  AWS CLI command for the Bedrock backend.  Defaults to "ews".
+
 ** Notable bug fixes
 
 - ~gptel-backend~ can now be set from customize buffers.  These are
--- a/gptel-bedrock.el
+++ b/gptel-bedrock.el
@@ -516,6 +516,12 @@ conversation."
 (defvar gptel-bedrock--aws-profile-cache nil
   "Cache for AWS profile credentials in the form of (PROFILE . CREDS).")
 
+(defvar gptel-bedrock-aws-cli-command (executable-find "aws")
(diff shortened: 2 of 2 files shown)


karthink:
@akssri and @felipeochoa, any feedback on this change?

It seems like `gptel-bedrock` should have been using `(executable-find "aws")` in the first place instead of `"aws"`?

liangfu:
Are you suggesting to use `(executable-find "aws")` instead of introducing new configurable `gptel-bedrock-aws-cli-command` ?

karthink:
> Are you suggesting to use `(executable-find "aws")` instead of introducing new
> configurable `gptel-bedrock-aws-cli-command` ?

No, `(executable-find "aws")` is equivalent to just calling "aws" in most respects.  Adding `gptel-bedrock-aws-cli-command`  seems reasonable to me, I'm just waiting on the authors' feedback.

akssri:
@karthink I think this was an oversight on our part; it probably should've been ```(executable-find "aws")```, but I'm not entirely sure what the ultimate difference would be (call-process probably does this internally).

@liangfu PR looks good; we may want to change the default to ```(executable-find "aws")``` based on the discussion here.

liangfu:
changed the default to `(executable-find "aws")`

karthink:
Thanks for the PR @liangfu!

PR #1252, opened 2026-02-17
Implement selective tools activation in gptel-mcp-connect

vspinu:
This MR allows registration of MCP servers with selective tool activation. Currently one can only activate *all* tools for a server.  My personal use case is to make all MCPs available to gptel, but don't activate any of them globally.

I am also adding a short line to the README as it take a bit of digging currently to figure this server initialization part.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -277,6 +277,9 @@
   Support for =gptel-send= in Term/Ansi-Term and Eat buffers is not
   yet available but planned.
 
+- =gptel-mcp-connect= now supports selective activation of tools for
+  each MCP server.
+
 ** Notable bug fixes
 
 - Function-valued system messages/directives are now evaluated in the
--- a/NEWS
+++ b/NEWS
@@ -141,6 +141,10 @@
   outputs simultaenously in a request.  Previously the two were
   incompatible.
 
+- You can now specify (programmatically) which MCP tools to add when
+  starting MCP servers.  Previously all tools provided by an MCP server
+  would be added automatically.  See ~gptel-mcp-connect~ for details.
+
 ** Notable bug fixes
 
 - Improved error messages: Curl request failures now include more
@@ -421,6 +425,9 @@
 
   Support for =gptel-send= in Term/Ansi-Term and Eat buffers is not
   yet available but planned.
+
+- =gptel-mcp-connect= now supports selective activation of tools for
+  each MCP server.
 
 ** Notable bug fixes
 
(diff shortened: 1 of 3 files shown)


karthink:
Thanks for the PR @vspinu!  Busy with some features, will take a look soon.

karthink:
@vspinu Reviewing this PR now.

1. I take it there is no change to the interactive behavior of gptel-mcp-connect?
2. Why the specification `("server-name" . t)`?  Aren't the cases covered by `"server-name"` and `("server-name" "tool1" "tool2" ...)`?

vspinu:
@karthink 

1. Yes, indeed. Only programatic usage is changed
2. It's for consistency ("server-name . nil) - no tools activated, ("server" . t) - all tools activated, ("server" 'a 'b) only tools. Such consistency is good for programatic usage, for instance a variable ("server" somevar), but it's a minor. I am fine either way. I can surely remove it if you don't like it.

karthink:
> 2. It's for consistency ("server-name . nil) - no tools activated, ("server" . t) - all tools activated, ("server" 'a 'b) only tools. Such consistency is good for programmatic usage, for instance a variable ("server" somevar), but it's a minor. I am fine either way. I can surely remove it if you don't like it.

Yeah, that makes sense.

@vspinu Could you update this PR?  I'd like to merge it but `gptel-mcp-connect` got some conflicting bug-fix changes in the meantime.

vspinu:
@karthink 

I have rewritten the code. After these months I no longer could recognize what I have done - not a good sign. Back then the priority was to get a minimal incision without changing the interface. 

Today, I think that's not the best approach, the code got too complex. So instead, I have moved the filtering code directly into `gptel-mcp--activate-tools` where it belongs. 

I took the liberty to change the behavior of this internal function by removing`&optional` from the `tools` argument. Basically removing `(unless tools (setq tools (gptel-mcp--get-tools))`. I think this is the right design for an internal utility function. It's used only once in the entire cocebase. If you want I can put that line back.

karthink:
> Today, I think that's not the best approach, the code got too complex. So instead, I have moved the filtering code directly into `gptel-mcp--activate-tools` where it belongs.

@vspinu I only see changes to `gptel-mcp-connect`.  Are you sure you've updated the branch?

vspinu:
@karthink Sorry for the delay. I don't know what went wrong, appologies.
 
I have re-pushed my local branch. Now it shows all the code correctly.

karthink:
Merged with minor tweaks.  Thanks @vspinu!

PR #1261, opened 2026-02-23
Add tool call editing and tool call hooks

karthink:
Draft PR to add support for fine-grained control over tool calls.  Planned features:

1. New hooks `gptel-pre-tool-call-functions` and `gptel-post-tool-call-functions`.
2. Editing tool call arguments at tool-call time (rewriting the conversation history in the process).  This will be possible both interactively (when using gptel-send) and from the hooks.
3. MAYBE: The hooks can cancel the tool call and continue the conversation with an explanation to the LLM for why the tool was canceled.
4. Editing tool call results from these hooks after the tool runs.
5. Specifying permission overrides (confirm/cancel) per tool at tool-call time from `gptel-pre-tool-call-functions`.  This determination can depend on the specific call arguments.
6. Specifying permission overrides (confirm/cancel) per tool at tool-call time with different scopes.  Interactively, this can be set
   - In the buffer ("Always run this tool in this (chat) buffer")
   - In this Emacs session ("Always run this tool until Emacs restarts.")
   - In this project or directory ("Always run this tool for requests in this project.")
     This one will require writing a file to the project directory, either to `.dir-locals`, `.gptel.eld` or `.gptel.json`.  Could also use a global registry stored in custom-vars, still undecided.
7. Better tool-call inspection buffer where the call arguments can be inspected in full (and edited).

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1499,20 +1499,20 @@ Perform UI updates and run post-response hooks."
                          (plist-put tool-call :result result))))))))))))))
 
 (defun gptel--update-wait (fsm)
-  "Update gptel's status after sending a request."
+  "Update gptel's status in FSM after sending a request."
   (with-current-buffer (plist-get (gptel-fsm-info fsm) :buffer)
     (when gptel-mode
       (gptel--update-status " Waiting..." 'warning))))
 
 (defun gptel--update-tool-call (fsm)
-  "Update gptel's status when calling a tool."
+  "Update gptel's status in FSM when calling a tool."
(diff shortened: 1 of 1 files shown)


karthink:
The scope of this PR has been reduced: the following two planned features will be covered by other PRs.

1. Permission overrides per tool at tool-call time with different scopes.
2. Editing tool call results interactively.  (Editing results via `gptel-post-tool-call-functions` has been implemented.)

All other planned features have been implemented.

Issue #1258, opened 2026-02-21
Idea: show the "context percentage" for the selected model


jdtsmith:
I tend to play with different models and they have wildly different context window sizes.  So I often end up with too much context.  It would be nice if the header in a `gptel-mode` buffer, and the `gptel-menu`, and the `*gptel-context*` buffer header all showed a small completion bar or color-coded _context budget percentage_ to indicate approximately how full your context buffer is, if gptel knows the limits.  A color-coded set of unicode box chars would be perfect for drawing this.

I can see a few ways to calculate "context buffer fullness":

1. just with the selected context you see in `*gptel-context*`.
2. including the full buffer or chain of org-tree nodes if in/invoked from a `gptel-mode` buffer.
3. including the current buffer up to point (unless you have "minibuffer instead" selected).

karthink:
See prior discussion in #580, and other issues/discussions I unfortunately can't find (GitHub issues search isn't great).

First the good news: gptel does know the limits if you're using a model that it includes the metadata for, which is most of them.  If you define a custom backend and specify a set of `:models` without associated metadata, you're out of luck.

Calculating the "context buffer fullness" exactly requires a network call and sending a bunch of data.  An approximate count is possible as `number of words / 2.5`, but I've been worried about performance.  It may be worth implementing something and testing if this really is an issue.  (I also try to avoid idle-timers in gptel, those tend to pile up with ones from other packages and slow down my Emacs)

I can try to explain why I expect performance to be an issue.  There are six sources of tokens in a gptel request: the system prompt, in-buffer text, Org/Markdown linked binary/text files, `gptel-context`, `gptel-tools` and `gptel-prompt-transform-functions`.

1. The system prompt: This could be a function, in which case we have to evaluate it to count the length.  The function could do anything, including file I/O -- users like to store long prompts in files.  (It can also be a list of strings, but that's not a problem.)  It might also be fetched from an Org property, which means running `org-entry-get`, possibly with property inheritance.  Note that we have to run `org-entry-get` in Org buffers no matter what, since we _don't know_ apriori if there is an Org heading with the property `GPTEL_SYSTEM` above the cursor.

2. In-buffer text: Some calculations here.  If `gptel-org-branching-context` is non-nil, we have to do a traversal of the buffer with `org-element-map` or `org-map-entries`, and can't just run `count-words-region`.

[...]

PR #1262, opened 2026-02-23
Add prompt token estimation and context window usage UI

karthink:
Following from #1258, a first attempt at token estimation without slowing down Emacs.

To try it out,

``` emacs-lisp
(require 'gptel-tokens)
```

and run `(gptel--estimate-tokens)` in a buffer.  The first run might take a bit (~10-100 ms), later runs should be faster because of caching.

1. It's not hooked up to the UI yet, but the idea is to run this on an idle timer via `gptel-mode-hook` to capture the result in a variable, and read from the variable to display it in gptel's header line, the context inspection menu, transient menu etc.

2. The heuristic is just `(number of words / 2.5)` right now, and something in between this and a full tokenizer can be implemented eventually.

3. Many edge cases!  The calculations should rightly be performed in the prompt buffer (created by `gptel--create-prompt-buffer`), but it's expensive to do this every time.

### DIFF
--- a/gptel-tokens.el
+++ b/gptel-tokens.el
@@ -53,11 +53,11 @@ Plist with keys:
       0
     (with-temp-buffer
       (insert str)
-      (/ (count-words (point-min) (point-max)) 2.5))))
+      (* (count-words (point-min) (point-max)) 2.5))))
 
 (defun gptel--estimate-tokens-from-words (num)
   "Estimate token count from NUM words."
-  (or (and (numberp num) (/ num 2.5)) 0))
+  (or (and (numberp num) (* num 2.5)) 0))
 
 (defun gptel--sha1 (obj)
   "Compute SHA1 hash of OBJ for caching.
--- a/gptel-tokens.el
+++ b/gptel-tokens.el
@@ -1,0 +1,330 @@
+;;; gptel-tokens.el --- Token estimation for gptel  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2025  Karthik Chikmagalur
+
+;; Author: Karthik Chikmagalur <karthikchikmagalur@gmail.com>
+;; Keywords: convenience
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Token estimation for gptel requests.  Provides a coarse estimate
+;; of how many tokens a gptel-send request would use, based on
+;; system prompt, buffer content, context, and tools.
+
+;;; Code:
+
+(require 'gptel-request)
+(require 'cl-lib)
+(require 'sha1)
+
+(defvar-local gptel--token-estimate-cache nil
+  "Cached token estimate data.
+
+Plist with keys:
+- :tokens - estimated token count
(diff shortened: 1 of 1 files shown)


jdtsmith:
I gave a quick try; it's a good start.  Once cache is warm, the `gptel--estimate-tokens` call takes about 50ms on a fairly fast machine for 35k tokens (against Copilot/Opus 4.6, 200k limit; though see below).  That would definitely slow down the interface on `post-command-hook` but I think is fine for an idle-timer.  Then again token count will continue to grow, so further optimization might be warranted.

The context estimate does appear to vary/miss in some unexpected ways.  For example, if I open a new empty `.org` file in gptel-mode, and include another long chat file as context (among several others), I get one value.  But if I request the same context estimate from the end of that long chat file included in the context, I get a longer estimate.  It's as if context inclusion status between files/buffers added and "current buffer to send" is not constant.  

But the bigger problem I noticed: all the estimates are too low, for the model combos I've been testing with.  E.g. here's an estimate just prior to sending a very small buffer, then the response after send:

```
(:tokens 26064 :context-window 200000 :percentage 13) [2 times]
Copilot error: (HTTP/2 400) prompt token count of 128886 exceeds the limit of 128000
```

If I make my own cheap estimate for this setup by calling `gptel-menu`, hitting `C`, and `M-x count-words` in the full context buffer, I find:

```
Buffer has 14175 lines, 3585 sentences, 66953 words, and 482176 characters
```

So with a 2.5 multiplier this would imply 193k tokens (as an approximation/upper-limit).  

BTW, this also reveals a "source error": via copilot, models are limited to 128k. 

Also, small thing, but I noticed that the `:percentage` estimate tops out at 100%.  For the interface, since this is an _estimate_ and not a certain calculation, I think going >100% is reasonable.

karthink:
> I gave a quick try; it's a good start.  Once cache is warm, the `gptel--estimate-tokens` call takes about 50ms on a fairly fast machine for 35k tokens (against Copilot/Opus 4.6, 200k limit; though see below).  That would definitely slow down the interface on `post-command-hook` but I think is fine for an idle-timer.  Then again token count will continue to grow, so further optimization might be warranted.

Right now the estimation is running `gptel--create-prompt-buffer`, which creates a buffer and copies over the text to be sent to the LLM (along with several buffer-local variables).  This handles the Org mode branching prompt and other subtleties, but wastes both CPU time and memory.  It's excessive just for counting words.  I suspect working around this should help a lot.

> The context estimate does appear to vary/miss in some unexpected ways.  For example, if I open a new empty `.org` file in gptel-mode, and include another long chat file as context (among several others), I get one value.  But if I request the same context estimate from the end of that long chat file included in the context, I get a longer estimate.  It's as if context inclusion status between files/buffers added and "current buffer to send" is not constant.

I don't follow.  In case 2, is the long chat file _also_ included in `gptel-context`?

> But the bigger problem I noticed: all the estimates are too low, for the model
> combos I've been testing with.  E.g. here's an estimate just prior to sending a
> very small buffer, then the response after send:
> 
> ```
> (:tokens 26064 :context-window 200000 :percentage 13) [2 times]
> Copilot error: (HTTP/2 400) prompt token count of 128886 exceeds the limit of 128000
> ```
> 
> If I make my own cheap estimate for this setup by calling `gptel-menu`, hitting
> `C`, and `M-x count-words` in the full context buffer, I find:
> 
> ```
> Buffer has 14175 lines, 3585 sentences, 66953 words, and 482176 characters
> ```
> 
> So with a 2.5 multiplier this would imply 193k tokens (as an approximation/upper-limit).  

Arithmetic error: I was dividing by 2.5 instead of multiplying.  Now fixed.

26064 * 2.5 * 2.5 = 162900, it's actually overestimating it by about 25%.

jdtsmith:
> I don't follow. In case 2, is the long chat file also included in gptel-context?

In either case, the long chat file is included.  When invoking `gptel-send` from a nearly empty chat buffer, the long file is included among the context.  When sending from the ed of the long chat file buffer itself, it's doubly included: by context, and by being the "implicit" context of everything in this buffer.  I'm not sure in which of the two directions the error goes.

> arithmetic error: I was dividing by 2.5 instead of multiplying. 

:)

BTW, I believe there's an API call to copilot that lists all the context limits; not sure if you are already hitting that.

PR #1264, opened 2026-02-25
Add :sync argument to gptel-request for synchronous LLM requests

tninja:
This PR is a follow up on another one: https://github.com/karthink/gptel/pull/1159

@karthink The feedback (https://github.com/karthink/gptel/pull/1159#issuecomment-3588309451) was addressed at this commit: https://github.com/karthink/gptel/commit/77ace42e9dda12b8b693d24baa606372f709a8e1, sorry for my slow turn around.

Would you mind help take a look? Let me know if there is any other feedback.

Regards

tninja

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -784,8 +784,7 @@ binary-encoded.")
 ;; Since we want this known at compile time, when markdown-mode is not
 ;; guaranteed to be available, we have to hardcode it.
 (defconst gptel-markdown--link-regex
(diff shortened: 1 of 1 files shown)


karthink:
@tninja I'll take a look.  BTW, why did you replace the `^M`s with newlines in the code?

tninja:
@karthink You're right — that change was accidental. I unintentionally normalized a couple of CR (`^M`) characters while editing (probably emacs editor normalized that during saving). I'll revert those hunks so the original `\r` handling stays intact, and push an updated commit. Sorry about the noise.

PR #1265, opened 2026-02-25
Fix to allow no-argument tools to work with Bedrock

jhoodsmith:
Without the fix `gptel--json-read-string` will throw an error when presented with empty string.

Aligns with what is done elsewhere in gptel


karthink:
Modified a bit and merged.  Thanks for the PR!

PR #1268, opened 2026-02-27
Fix header-line right-alignment for variable-pitch fonts

### LINKED ISSUE #763: gptel's `header-line` not right aligned well.
Image

the `header-lien-format` is:
```elisp
header-line-format is a buffer-local variable defined in buffer.c.

Value in #<buffer *ChatGPT*>
((:eval
   (concat (propertize " " 'display '(space :align-to 0))
     (format "%s" (gptel-backend-name gptel-backend))))
  #(" Ready" 0 6 (face success))
  (:eval
    (let*
      ((model (gptel--model-name gptel-model))
        (system
          (propertize
            (buttonize
              (format "[Prompt: %s]"
                (or (car-safe (rassoc gptel--system-message gptel-directives))
                  (gptel--describe-directive gptel--system-message 15)))
              (lambda (&rest _) (gptel-system-prompt)))
            'mouse-face 'highlight 'help-echo "System message for session"))
        (context
          (and gptel-context--alist
            (cl-loop for entry in gptel-context--alist if (bufferp (car entry))
              count it into bufs else count (stringp (car entry)) into files
              finally return
              (propertize
                (buttonize
                  (concat "[Context: " (and (> bufs 0) (format "%d buf" bufs))
                    (and (> bufs 1) "s") (and (> bufs 0) (> files 0) ", ")
                    (and (> files 0) (format "%d file" files))
                    (and (> files 1) "s") "]")
                  (lambda (&rest _) (require 'gptel-context)
                    (gptel-context--buffer-setup)))
                'mouse-face 'highlight 'help-echo "Active gptel context"))))
        (toggle-track-media
          (lambda (&rest _)
            (setq-local gptel-track-media (not gptel-track-media))
            (if gptel-track-media
              (message
                (concat
                  "Sending media from included links.  To include media, create "
                  "a \"standalone\" link in a paragraph by itself, separated from surrounding text."))
              (message "Ignoring image links.  Only link text will be sent."))
            (run-at-time 0 nil #'force-mode-line-update)))
        (track-media
          (and (gptel--model-capable-p 'media)
            (if gptel-track-media
              (propertize (buttonize "[Sending media]" toggle-track-media)
                'mouse-face 'highlight 'help-echo
                "Sending media from standalone links/urls when supported.\nClick to toggle")
              (propertize (buttonize "[Ignoring media]" toggle-track-media)
                'mouse-face 'highlight 'help-echo
                "Ignoring images from standalone links/urls.\nClick to toggle"))))
        (toggle-tools
          (lambda (&rest _) (interactive)
            (run-at-time 0 nil (lambda nil (call-interactively #'gptel-tools)))))
        (tools
          (when (and gptel-use-tools gptel-tools)
            (propertize
              (buttonize
                (pcase (length gptel-tools)
                  (0 "[No tools]") (1 "[1 tool]")
                  (len (format "[%d tools]" len)))
                toggle-tools)
              'mouse-face 'highlight 'help-echo "Select tools"))))
      (concat
        (propertize " " 'display
          `(space :align-to
             (- right
               ,(+ 5 (length model) (length system) (length track-media)
                  (length context) (length tools)))))
        tools (and track-media " ") track-media (and context " ") context " "
        system " "
        (propertize
          (buttonize (concat "[" model "]") (lambda (&rest _) (gptel-menu)))
          'mouse-face 'highlight 'help-echo "Model in use")))))

Buffer values Global value Set

Documentation
Analogous to mode-line-format, but controls the header line.

The header line appears, optionally, at the top of a window; the mode
line appears at the bottom.

Also see header-line-indent-mode if display-line-numbers-mode is
turned on and header-line text should be aligned with buffer text.

View in manual

References
Finding references in a .c file is not supported.

Find all references

Source Code
// Defined in ~/Projects/git.savannah.gnu.org/git/emacs/src/buffer.c
  DEFVAR_PER_BUFFER ("header-line-format",
		     &BVAR (current_buffer, header_line_format),
		     Qnil,
		     doc: /* Analogous to `mode-line-format', but controls the header line.
The header line appears, optionally, at the top of a window; the mode
line appears at the bottom.

Also see `header-line-indent-mode' if `display-line-numbers-mode' is
turned on and header-line text should be aligned with buffer text.  */);

Symbol Properties
risky-local-variable
  t
variable-documentation
  503966
watchers
  (#<subr set-buffer-redisplay>)
```

kovan:
Fixes #763 — header-line right-alignment off due to `length` counting characters instead of pixel width.

Builds the RHS string first, then uses `string-pixel-width` (Emacs 29+, falls back to `string-width`) for font-independent alignment. Eliminates hardcoded offset.

**Verified** on Emacs 30.2: old spec used `(- right 43)` (char-based + hardcoded 5), new spec uses `(- right (38))` (pixel-wrapped form).

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -828,19 +828,22 @@ Search between BEG and END."
                                 toggle-tools)
                      'mouse-face 'highlight
                      'help-echo "Select tools"))))
-      (concat
-       (propertize
-        " " 'display
-        `(space :align-to (- right
-                             ,(+ 5 (length model) (length system)
-                                 (length track-media) (length context) (length tools)))))
-       tools (and track-media " ") track-media (and context " ") context " " system " "
-       (propertize
-        (buttonize (concat "[" model "]")
-                   (lambda (&rest _) (gptel-menu)))
-        'mouse-face 'highlight
-        'help-echo "Model in use"))))
-  "Information segment for the header-line in gptel-mode.")
+      (let ((rhs (concat
+                  tools (and track-media " ") track-media
+                  (and context " ") context " " system " "
+                  (propertize
+                   (buttonize (concat "[" model "]")
+                              (lambda (&rest _) (gptel-menu)))
+                   'mouse-face 'highlight
+                   'help-echo "Model in use"))))
+        (concat
+         (propertize
+          " " 'display
+          (if (fboundp 'string-pixel-width)
+              `(space :align-to (- right (,(string-pixel-width rhs))))
+            `(space :align-to (- right ,(+ 5 (string-width rhs))))))
+         rhs))))
+  "Information segment for the header-line in `gptel-mode'.")
 
 (defun gptel-use-header-line ()
   "Set up the header-line for a gptel buffer.


karthink:
Thank you @kovan!  I'll merge this soon, but I have a couple of notes:

1. I couldn't find `string-pixel-width` in `compat-29` or `compat-30`, or anywhere in the compat directory.  Where is it?

2. If falling back to `string-width`, we should retain the hardcoded offset of 5 chars.

kovan:
Thanks for the review!

1. You're right — `string-pixel-width` was [dropped from compat](https://github.com/emacs-compat/compat/issues/8) due to performance issues. The `fboundp` check now effectively gates on native Emacs 29+ only.

2. Fixed — the `string-width` fallback now includes the original `(+ 5 ...)` offset to match the previous behavior.

Pushed both fixes in 22b2e7d.

kovan:
Also added 2 unit tests in the test submodule (kovan/gptel-test@test/header-line-alignment):

- `gptel-test-header-line-pixel-alignment` — verifies pixel spec wraps offset in a list `(N)` on Emacs 29+
- `gptel-test-header-line-char-fallback-offset` — verifies char fallback includes the `+5` padding offset

The submodule ref is updated in 2f96598.

karthink:
The changes look good now, thanks.

Updating the submodule ref in this repo without updating gptel-test doesn't make sense, does it?  You need to raise a PR on gptel-test to add the tests there first.  (I just checked and didn't see any PR, let me know if I missed something.)

kovan:
Good catch — raised the PR on gptel-test: https://github.com/karthink/gptel-test/pull/13

I've dropped the submodule update commit from this branch for now. Once the test PR is merged, I can add it back (or you can update the ref separately).

karthink:
I just took a look at the definition of `string-pixel-width` in subr-x.el.  Considering that this is going to run on every redisplay, this does not look good!

It makes me think there must be some more primitive way to right-align a string in Emacs without hacks like this.  This is such a common requirement.

I'll merge the PR but I'm going to be profiling gptel for the next few days.  If `string-pixel-width` shows up on the profile I might have to revert this commit.

FWIW, it looks like it shouldn't be an issue:

``` emacs-lisp
(benchmark-run-compiled (string-pixel-width (make-string (random 45) ?a)) 10000)
;; => (7.1169e-05 0 0.0)
```

kovan:
Good to see the benchmark confirms it's negligible — the \`with-temp-buffer\` in \`string-pixel-width\` looks heavy on paper but the numbers say otherwise.

Unfortunately there isn't a more primitive way to do this in Emacs. The \`string-pixel-width\` → \`:align-to (- right (N))\` pattern is the canonical approach for pixel-accurate right-alignment (doom-modeline and others do the same). The mode-line's built-in \`%=\` only works character-by-character so it can't handle variable-pitch fonts, which is the whole point of this fix.

If profiling does flag it, the straightforward optimization would be to cache the pixel width and only recompute when the header content actually changes (model/context/tools/media toggles) — those are all infrequent events. Happy to send a follow-up patch if that's needed.

karthink:
> If profiling does flag it, the straightforward optimization would be to cache the pixel width and only recompute when the header content actually changes (model/context/tools/media toggles) — those are all infrequent events. Happy to send a follow-up patch if that's needed.

Hopefully caching won't be needed.  I'll merge it soon.

PR #1269, opened 2026-02-27
xAI: `grok-4{,-1}-fast-{,-non}-reasoning`

endgame:
Add support for some missing Grok models, and re-sort the backend config to match the list at https://docs.x.ai/developers/models for easy future syncing.


karthink:
@endgame Merged, thanks for the PR!

PR #1270, opened 2026-02-27
gptel-openai-extras: Correct context window limits on xAI models

endgame:
Sorry to send rapid-fire PRs but a) I forgot to fix the context limits on the new models, and b) some of the old models had missed that `:context-window` is measured in thousands of tokens.


karthink:
No worries, thanks for the PR!

PR #1271, opened 2026-02-27
gptel-gh: Update models list

gessen:
Cleaned up old models and added missing ones:
- `gpt-41-copilot` is a code completion model and cannot be used by `gptel`
- both `gpt-5` and `claude-opus-41` are no longer supported

However, some exisiting, as well as newly added, OpenAi models (starting from `gpt-5.1-codex`) are not supported by `/chat/completions` endpoint so at this point it may be pointless to have them.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -20,6 +20,9 @@
   =grok-4-1-fast-non-reasoning=, =grok-4-fast-reasoning=, and
   =grok-4-fast-non-reasoning=.
 
+- - GitHub Copilot backend: Add support for =gpt-5.1-codex-,
+  =gpt-5.1-codex-mini, =claude-sonnet-4.6= and =gemini-3.1-pro-preview=.
+
 ** New features and UI changes
 
 - When using ~setopt~ or the customize interface, ~gptel-backend~ can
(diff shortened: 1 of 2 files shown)


karthink:
> However, some exisiting, as well as newly added, OpenAi models (starting from
> `gpt-5.1-codex`) are not supported by `/chat/completions` endpoint so at this
> point it may be pointless to have them.

Are you saying that `gpt-5.1-codex` and others are not supported _by GitHub Copilot_ using the `/chat/completions` endpoint, or not supported by OpenAI, or both?

gessen:
The models are not suported by GitHub Copilot using the `/chat/completions` endpoint. This is the error I see:
```
Copilot error: (HTTP/2 400) model "gpt-5.1-codex" is not accessible via the /chat/completions endpoint
```
With endpoint `/responses` I have:
```
Copilot error: (HTTP/2 400) Invalid type for 'input': expected a string, but got an object instead.
```
Just changing the endpoint isn't enough.

karthink:
> ```
> Copilot error: (HTTP/2 400) Invalid type for 'input': expected a string, but got an object instead.
> ```
> Just changing the endpoint isn't enough.

This is expected, the responses API format is different.  I'll merge this for now, thanks.  It will probably need to be updated again once I add support for the responses API (PR #1168, PR #1218).

karthink:
Merged, thank you for the PR!

PR #1275, opened 2026-03-02
gptel-gh: update models

luismbo:
I've updated the GitHub Copilot models using this list from VS Code as reference.

image

Some questions/comments:
1. What is the `:cutoff-date` and where can I find that information?
2. Where can I find reference info for `:capabilities` and `:description`?
3. I've assumed `:context-window` is input only.
4. Is there some better source for this information that the VS Code listing?


luismbo:
D'oh. Just noticed that the list was updated two days ago. Closing.

Issue #1035, opened 2025-08-19
Magistral with thinking mode: wrong type argument: stringp


zauster:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

Hey, 

I am using gptel with the Magistral model from mistral.ai. This is a thinking/reasoning model, it seems that the reasoning and response json-content is not correctly parsed:

```
error in process sentinel: apply: Wrong type argument: stringp, [(:type "thinking" :thinking [(:type "text" :text "The user is asking for the model name. Since I am a large language model living in Emacs, I should respond with the name of my model. However, the problem statement does not explicitly mention the model name. I recall that my model name is typically referred to in documentation or configuration files, but since I don't have access to those right now, I should respond with a general description of myself.")]) (:type "text" :text "I am a large language model living in Emacs, designed to assist with a variety of tasks.")]
error in process sentinel: Wrong type argument: stringp, [(:type "thinking" :thinking [(:type "text" :text "The user is asking for the model name. Since I am a large language model living in Emacs, I should respond with the name of my model. However, the problem statement does not explicitly mention the model name. I recall that my model name is typically referred to in documentation or configuration files, but since I don't have access to those right now, I should respond with a general description of myself.")]) (:type "text" :text "I am a large language model living in Emacs, designed to assist with a variety of tasks.")]
``` 

Include reasoning with response is set to "yes".

The gptel-debug-log has this:

```
{
  "gptel": "request headers",
  "timestamp": "2025-08-19 12:54:09"
}
{
  "Content-Type": "application/json",
  "Authorization": "Bearer XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
}
{
  "gptel": "request body",
  "timestamp": "2025-08-19 12:54:09"
}
{
  "model": "magistral-medium-latest",
  "messages": [
    {
      "role": "system",
      "content": "You are a large language model living in Emacs and a helpful assistant. Respond concisely."
    },
    {
      "role": "user",
      "content": "Hey, state your model name."
    }
  ],
  "stream": false,
  "temperature": 1.0
}
{"gptel": "request Curl command", "timestamp": "2025-08-19 12:54:09"}
curl \
--disable \
--location \
--silent \
--compressed \
-XPOST \
-y7200 \
-Y1 \
-D- \
-w\(e133abaea80225712705774d39fb9c7e\ .\ \%\{size_header\}\) \
[...]

Issue #848, opened 2025-05-17
wrong-type-argument sequencep :null: gptel--parse-response throws an error when tool-calls returns :null  in Mistral


JahQuellin:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

When I set up gptel with MistralLeChat with a very basic configuration straight from the repo I got an error wrong-type-argument sequencep :null. The error wrong-type-argument sequencep :null indicates that a function expected a sequence (like a list or vector) but received the symbol :null instead. This issue occurs in the gptel--parse-response function when processing the response from an API call. Specifically, it seems to be related to the handling of tool_calls in the response data, which is :null by MistralLeChat and is being treated as if it were a sequence.

Without really understanding what I am doing I have managed to get it working by inserting the following line into the gptel--parse-response function in  gptel-openai.el:

`
       (unless (eq tool-calls :null)`

The resulting function then looks like this: 

```
(cl-defmethod gptel--parse-response ((_backend gptel-openai) response info)
  "Parse an OpenAI (non-streaming) RESPONSE and return response text.

Mutate state INFO with response metadata."
  (let* ((choice0 (map-nested-elt response '(:choices 0)))
         (message (plist-get choice0 :message))
         (content (plist-get message :content)))
    (plist-put info :stop-reason
               (plist-get choice0 :finish_reason))
    (plist-put info :output-tokens
               (map-nested-elt response '(:usage :completion_tokens)))
    ;; OpenAI returns either non-blank text content or a tool call, not both.
    ;; However OpenAI-compatible APIs like llama.cpp can include both (#819), so
    ;; we check for both tool calls and responses independently.
    (when-let* ((tool-calls (plist-get message :tool_calls)))
       (unless (eq tool-calls :null)
      (gptel--inject-prompt        ; First add the tool call to the prompts list
       (plist-get info :backend) (plist-get info :data) message)
      (cl-loop             ;Then capture the tool call data for running the tool
       for tool-call across tool-calls  ;replace ":arguments" with ":args"
       for call-spec = (copy-sequence (plist-get tool-call :function))
       do (ignore-errors (plist-put call-spec :args
                                    (gptel--json-read-string
                                     (plist-get call-spec :arguments))))
       (plist-put call-spec :arguments nil)
       (plist-put call-spec :id (plist-get tool-call :id))
       collect call-spec into tool-use
       finally (plist-put info :tool-use tool-use))))
[...]

Issue #819, opened 2025-05-01
Failed to call function if response contains content


longlene:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

Actually, I created an issue for llama.cpp before to report the issue: https://github.com/ggml-org/llama.cpp/issues/13224

When I was testing Qwen3 + llama-server + gptel, when the tool is called by LLM, the following response is returned to gptel:

```json
{                                                                                                                                                           
  "choices": [                                                                                                                                              
    {                                                                                                                                                       
      "finish_reason": "tool_calls",                                                                                                                        
      "index": 0,                                                                                                                                           
      "message": {                                                                                                                                          
        "role": "assistant",                                                                                                                                
        "reasoning_content": "***",                                                                                                                                   
        "content": "\n\n",                                                                                                                                  
        "tool_calls": [                                                                                                                                     
          {                                                                                                                                                 
            "type": "function",                                                                                                                             
            "function": {                                                                                                                                   
              "name": "tool_name",                                                                                                                  
              "arguments": "{\"query\":\"Qwen3\"}"                                                                                        
            },                                                                                                                                                                                                                                                                                                                                                                                                                                                    
...                                                                                                      
  }                                                                                                                                                         
}
```
We can see that the response contains both `content` and `tool_calls` field.
However, after checking the source code, https://github.com/karthink/gptel/blob/db4e5c7ea39f7d58d8ca4856c98a6df71b1e18bf/gptel-openai.el#L267 this non-null `content` would cause gptel print empty without any tool calling.

### Backend

None

### Steps to Reproduce

1. llama-server -m Qwen3-32B.gguf to setup a server that listens on 8080 port
2. configure it to work with gptel
```emacs-lisp
  (gptel-backend                                                     
   (gptel-make-openai "llama-cpp"                                    
     ;;:stream t                                                     
     :protocol "http"                                                
     :host "localhost:8080"                                          
     :models '(local))) 
```
3. add some tool in emacs side
4. enter gptel and ask LLM to call this tool
5. then nonthing would happen, the tool call in response is just ignored

### Additional Context

https://github.com/karthink/gptel/blob/db4e5c7ea39f7d58d8ca4856c98a6df71b1e18bf/gptel-openai.el#L267

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

longlene:
[...]

PR #1284, opened 2026-03-11
gptel-openai: Merge tool calls from all choices in non-streaming responses

phrb:
## Problem

`gptel--parse-response` for `gptel-openai` hardcodes `(map-nested-elt response '(:choices 0))`, only reading the first entry in the `choices` array. This silently drops tool calls when an OpenAI-compatible API returns them in separate `choices` entries.

### Affected setup

GitHub Copilot backend (`gptel--gh`) proxying Claude models (e.g. `claude-opus-4.6`). Non-streaming requests only — sub-agents (researcher, introspector, executor via `gptel-agent`) always use `stream: false` and are completely broken by this.

### What happens

The Copilot API proxy reformats Claude's non-streaming responses, placing text content and each tool call in separate `choices` entries:

**Standard OpenAI format** (what gptel expects):
```json
{
  "choices": [
    {
      "finish_reason": "tool_calls",
      "message": {
        "content": "I'll search the source code...",
        "tool_calls": [
          {"function": {"name": "Glob", "arguments": "..."}, "id": "..."},
          {"function": {"name": "Grep", "arguments": "..."}, "id": "..."}
        ]
      }
    }
  ]
}
```

**Copilot format** (Claude via proxy, `stream: false`):
```json
{
  "choices": [
    {
      "finish_reason": "tool_calls",
      "message": {
        "content": "I'll search the source code...",
        "role": "assistant"
      }
    },
    {
      "finish_reason": "tool_calls",
      "message": {
        "role": "assistant",
        "tool_calls": [
          {"function": {"name": "Glob", "arguments": "..."}, "id": "...", "type": "function"}
        ]
      }
    },
    {
      "finish_reason": "tool_calls",
      "message": {
        "role": "assistant",
        "tool_calls": [
          {"function": {"name": "Grep", "arguments": "..."}, "id": "...", "type": "function"}
        ]
      }
    }
  ]
}
```

Since the parser only reads `choices[0]`, it finds the text content but no `tool_calls` → `:tool-use` is never set in info → the FSM transitions to DONE instead of TOOL → the tool call loop never executes.

For sub-agents this means every researcher/introspector/executor returns only its preamble text (e.g. "I'll systematically search the gptel source code...") without ever running any tools.

### Streaming is not affected

The streaming parser (`gptel-curl--parse-stream`) works correctly because:
1. It uses a different accumulation strategy via `choices[0].delta.tool_calls` with `index` fields
2. Copilot's streaming format already keeps everything in `choices[0]`

## Fix

Iterate over all `choices` entries, collecting `tool_calls` and `content` from whichever entry has them, then merge into a single message structure before the existing processing code runs.

When there is only one choice (standard OpenAI format), the merge block is guarded by `(when (> (length choices) 1) ...)` and skipped entirely — zero overhead for the common case.

## Testing

Tested with GitHub Copilot backend + `claude-opus-4.6`:
- **Before fix**: All sub-agent tool call loops fail silently (3/3 agents return preamble only)
- **After fix**: All sub-agent tool call loops complete successfully (3/3 agents execute tools and return substantive results across multiple TOOL→WAIT→TYPE cycles)
- **Streaming requests**: Unaffected, work as before

## Related issues

- #819 — content + tool_calls coexistence in non-streaming responses (same parser area)
- #848 — `:null` tool_calls handling in `gptel--parse-response`
- #1035 — non-string content in `choices[0].message`

### DIFF
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -158,7 +158,7 @@ information if the stream contains it."
                           (plist-put info :partial_json (list (plist-get func :arguments)))
                           ;; NOTE: Do NOT use `push' for this, it prepends and we lose the reference
                           (plist-put info :tool-use (cons tool-call (plist-get info :tool-use))))
-                      ;; old tool block continues, so continue collecting arguments in :partial_json 
+                      ;; old tool block continues, so continue collecting arguments in :partial_json
                       (push (plist-get func :arguments) (plist-get info :partial_json)))))
                 ;; Check for reasoning blocks, currently only used by Openrouter
                 (unless (eq (plist-get info :reasoning-block) 'done)
@@ -184,12 +184,31 @@ information if the stream contains it."
   "Parse an OpenAI (non-streaming) RESPONSE and return response text.
 
 Mutate state INFO with response metadata."
-  (let* ((choice0 (map-nested-elt response '(:choices 0)))
+  (let* ((choices (plist-get response :choices))
+         (choice0 (and (> (length choices) 0) (aref choices 0)))
          (message (plist-get choice0 :message))
          (content (plist-get message :content)))
     (plist-put info :stop-reason
                (plist-get choice0 :finish_reason))
     (gptel--openai-update-tokens (map-nested-elt response '(:usage)) info)
+    ;; Some OpenAI-compatible APIs (e.g. GitHub Copilot proxying Claude) return
+    ;; tool calls in separate `choices' entries rather than combining them in
+    ;; choices[0].message.tool_calls.  Merge them here.
+    (when-let* ((tool-calls
+                 (cl-loop
+                  for choice across choices
+                  for msg = (plist-get choice :message)
+                  for tc = (plist-get msg :tool_calls)
+                  when (and tc (not (eq tc :null)))
+                  vconcat tc
+                  when (and (not content)
+                            (let ((c (plist-get msg :content)))
+                              (and c (not (eq c :null))
+                                   (not (string-empty-p c)))))
+                  do (setq content (plist-get msg :content))))
+                ((> (length tool-calls) 0)))
+      (plist-put message :tool_calls tool-calls)
+      (plist-put message :content (or content :null)))
     ;; OpenAI returns either non-blank text content or a tool call, not both.
     ;; However OpenAI-compatible APIs like llama.cpp can include both (#819), so
     ;; we check for both tool calls and responses independently.
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -1902,7 +1902,7 @@ MACHINE is an instance of `gptel-fsm'"
   ;; a second network request: gptel tests for the presence of these flags to
   ;; handle state transitions.  (NOTE: Don't add :uuid to this.)
   (let ((info (gptel-fsm-info fsm)))
-    (dolist (key '(:tool-result :tool-use :error :http-status :reasoning :tokens))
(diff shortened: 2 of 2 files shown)


phrb:
This fixes the sub-agent failure reported in karthink/gptel-agent#66 for users on the GitHub Copilot backend with Claude models.

karthink:
I'm assuming that no backend returns a "contents" field outside of `choices[0].messages`.  Why not just:

``` emacs-lisp
;; OpenAI returns either non-blank text content or a tool call, not both.
;; However OpenAI-compatible APIs like llama.cpp can include both (#819), so
;; we check for both tool calls and responses independently.
;; Some OpenAI-compatible APIs return tool calls in separate `choices'
;; instead of in choices[0].messages.tool_calls, we merge them
(when-let* ((tool-calls
             (cl-loop
              for choice across choices
              vconcat (map-nested-elt choice '(:message :tool_calls))))
            ((not (eq tool-calls :null))))
  (when (> (length choices) 1)
    (plist-put message :tool_calls tool-calls))
  (gptel--inject-prompt        ; First add the tool call to the prompts list
   (plist-get info :backend) (plist-get info :data) message)
```

phrb:
@karthink Do you want your suggestion to be added to this branch, or did you already implement it elsewhere?

karthink:
I meant it as a more concise alternative to your implementation in this PR, as part of this PR.  I also wanted to know if you see any problem with this implementation.

carl-reverb:
Thank you, @phrb for digging into this! I just tested your branch since I'm on the same API/model stack and was frustrated by subagents constantly popping up and failing. This works like a charm!

phrb:
I've been rebasing it to the latest on main, I couldn't yet revisit it with the suggestions from karthink, but plan to do it.

karthink:
I can do it if you're busy, but I need to be sure that my version isn't 
missing something important.

judaew:
This fix resolves the DeepSeek issue where it returns HTTP/400 errors because the request's "assistant" block lacks the "tool_calls" field. Excellent improvement!

karthink:
@phrb gentle ping

phrb:
Hi @karthink, sorry for the delay, you do have the latest changes, I've been using the version on this branch for a while now, and it does still work.

karthink:
>  I've been using the version on this branch for a while now, and it does still work.

My question is if the code I suggested in this comment

https://github.com/karthink/gptel/pull/1284#issuecomment-4052089080

will work, because it is simpler than your suggested changes.

karthink:
@phrb gentle ping

phrb:
@karthink, just updated the code with a slightly modified version of your suggestion.
I just tested on my setup and it does work.

karthink:
@phrb This looks the same as before to me, and is difficult to follow.  Do you foresee any issue with the much more concise

``` emacs-lisp
(when-let* ((tool-calls
             (cl-loop
              for choice across choices
              vconcat (map-nested-elt choice '(:message :tool_calls))))
            ((not (eq tool-calls :null))))
  (when (> (length choices) 1)
    (plist-put message :tool_calls tool-calls))
  (gptel--inject-prompt        ; First add the tool call to the prompts list
   (plist-get info :backend) (plist-get info :data) message)
```

The extra thing your code does is look for `:content` in the other choices blocks.  Is non-null/non-blank `:content` in the other choices blocks something you've observed in the responses?

karthink:
@phrb gentle ping

karthink:
@phrb gentle ping

PR #1286, opened 2026-03-13
gptel-openai: Add gpt-5.3-chat-latest and gpt-5.4 models

berenddeboer:
This PR adds support for two new OpenAI models:

- gpt-5.3-chat-latest: Answers right away
- gpt-5.4: The best model for coding and agentic tasks

Both models support media, tool-use, json, and URL capabilities with a 400K context window.


karthink:
Merged, thanks for the PR!

PR #1288, opened 2026-03-14
Add spaces

aot-todorov:
Adding spaces to reasoning org blocks allows them to be expanded/collapsed by tabbing on them.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -1686,7 +1686,7 @@ Optional RAW disables text properties and transformation."
                          (not (string-suffix-p "\n" (gptel-response-prefix-string)))
                          "\n"))
                    (blocks (if (derived-mode-p 'org-mode)
-                               `("#+begin_reasoning\n" . ,(concat "\n#+end_reasoning"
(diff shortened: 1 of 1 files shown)


karthink:
@aot-todorov Can you provide some more context for this change?  I can expand/collapse reasoning blocks with TAB right now without issues.

aot-todorov:
On Emacs installed from apt in both Ubuntu and Debian tab expand/collapse does not work without a space after the name of the tag. Confirmed the behaviour on Debian old-stable, testing and Ubuntu 2024.04 LTS.

On Emacs build from official sources this isn't a problem.

I guess this isn't a gptel issue, but a Debian one. I will look into it more. Leaving it up to you to close or accept a hack fix that's distro based.

pabl0:
Works fine with Debian trixie (current stable; emacs 1:30.1+1-6), with built-in org.

karthink:
@pabl0 Thanks for checking.  I'm closing this issue.  In the future gptel might switch to using a drawer `:RESPONSES:...:END:` instead of an Org block so will probably become a non-issue anyway.

Issue #1289, opened 2026-03-14
Groq example config creates ‘invalid_request_error’


andreas-roehler:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

Running Groq example config from README.org

Got

Groq error: ((HTTP/2 400) invalid_request_error) The model `llama-3.1-70b-versatile` has been decommissioned and is no longer supported. Please refer to https://console.groq.com/docs/deprecations for a recommendation on which model to use instead.

---

The page mentioned above suggests ‘llama-3.1-70b-versatile’

The patch below should fix it:

```
diff --git a/README.org b/README.org
index 04870a9..45f820c 100644
--- a/README.org
+++ b/README.org
@@ -723,7 +723,7 @@ Register a backend with
       :endpoint "/openai/v1/chat/completions"
       :stream t
       :key "your-api-key"                   ;can be a function that returns the key
-  :models '(llama-3.1-70b-versatile
+      :models '(llama-3.3-70b-versatile
                 llama-3.1-8b-instant
                 llama3-70b-8192
                 llama3-8b-8192
```

### Backend

Other (please specify in Additional Context)

### Steps to Reproduce

Select Groq and llama-3.1-70b-versatile

### Additional Context

GNU Emacs 31.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.50, cairo version 1.18.4) of 2026-03-01

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

karthink:
Fixed by #1290.

PR #1290, opened 2026-03-14
#1289, Groq example config creates ‘invalid_request_error’

andreas-roehler:
Deprecated model replaced according to https://console.groq.com/docs/deprecations


karthink:
Thanks for the PR!  Merged.

Issue #1234, opened 2026-02-04
curl fails silently when missconfigured


igmacs:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

I had CA certificates missconfigured (for reasons orthogonal to this issue), and as a result I had the following error when running curl:

```
$ curl -v https://api.openai.com
* Host api.openai.com:443 was resolved.
* IPv6: (none)
* IPv4: 162.159.140.245, 172.66.0.243
*   Trying 162.159.140.245:443...
* Connected to api.openai.com (162.159.140.245) port 443
* GnuTLS ciphers: NORMAL:-ARCFOUR-128:-CTYPE-ALL:+CTYPE-X509:-VERS-SSL3.0
* ALPN: curl offers h2,http/1.1
* SSL connection using TLS1.3 / ECDHE_RSA_AES_256_GCM_SHA384
* server certificate verification failed. CAfile: none CRLfile: none
* Closing connection
curl: (60) server certificate verification failed. CAfile: none CRLfile: none
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

But when I tried to use gptel, the message "Querying ChatGPT..." was printed and nothing ever happened afterwards. Instead, I would have expected some kind of message telling me that curl failed.

### Backend

None

### Steps to Reproduce

In my case curl was failing because I was experimenting with the Guix package manager, but of course it doesn't make sense to give those reproduction steps here. But the issue can be reproduced with a similar error like this:

1) evaluate (setenv "SSL_CERT_FILE" "~/bad-path" t)
2) Try running curl directly in a new Emacs shell. You should see an error like this one:
```
$ curl -v https://api.openai.com
* Host api.openai.com:443 was resolved.
* IPv6: (none)
* IPv4: 172.66.0.243, 162.159.140.245
*   Trying 172.66.0.243:443...
* Connected to api.openai.com (172.66.0.243) port 443
* error reading ca cert file ~/bad-path (Error while reading file.)
* Closing connection
curl: (77) Problem with the SSL CA cert (path? access rights?)
```
3) Try using gptel. You should get a "Querying ChatGPT..." message, but nothing after that, neither an output nor an error 

### Additional Context

Emacs version: "30.2"
gptel version: latest
[...]

PR #1293, opened 2026-03-16
Fix search-failed error when curl exits without completing transfer

benthamite:
## Summary

- `gptel-curl--parse-response` and `gptel-curl--stream-cleanup` call `search-backward` to find the UUID marker in the curl process buffer without passing NOERROR.
- The UUID is written by curl's `-w` flag, which only executes after a completed transfer. When curl exits due to a connection-level failure (network error, DNS failure, timeout, etc.), it never writes the `-w` output, so the UUID is absent and `search-backward` signals `search-failed`.
- This patch passes NOERROR (`nil t`) to both `search-backward` call sites and returns a meaningful error when the marker is missing, consistent with how `gptel-curl--log-response` already handles this case via `(search-forward ... nil t)`.

## Reproduction

1. Start a non-streaming gptel request (e.g., via a programmatic `gptel-request` call to a Gemini backend).
2. Kill the buffer before the response arrives, or trigger a connection-level curl failure (e.g., network drop).
3. Observe `Debugger entered--Lisp error: (search-failed "<uuid>")` in `gptel-curl--parse-response`.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -2798,35 +2798,47 @@ PROC-INFO is the plist containing process metadata."
   "Process sentinel for gptel curl requests.
 
 PROCESS and _STATUS are process parameters."
-  (let ((proc-buf (process-buffer process)))
+  (let ((proc-buf (process-buffer process))
+        (exit-status (process-exit-status process)))
     (let* ((fsm (car (alist-get process gptel--request-alist)))
            (info (gptel-fsm-info fsm))
            (http-status (plist-get info :http-status)))
       (when gptel-log-level (gptel-curl--log-response proc-buf info)) ;logging
-      (if (member http-status '("200" "100")) ;Finish handling response
-          ;; Run the callback one last time to signal that the process has ended
-          (with-demoted-errors "gptel callback error: %S"
-            (funcall (plist-get info :callback) t info))
-        (with-current-buffer proc-buf   ; Or Capture error message
-          (goto-char (point-max))
-          (search-backward (plist-get info :uuid))
-          (backward-char)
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR @benthamite.

> 2. Kill the buffer before the response arrives, or trigger a connection-level
>    curl failure (e.g., network drop).

Which buffer do you mean here?

karthink:
I think this might be a good opportunity to fix #1234 while we're at it.  If Curl exits without completing the transfer, it will exit with a process code different from 0.  The process sentinels can check `process-exit-status` -- if it is different from 0, it can be reported as an error.  The safe check for the uuid doesn't hurt, but we will probably not trip if we check Curl's status.  What do you think?

benthamite:
> Which buffer do you mean here?

The buffer where the response would be inserted, i.e. the user's chat buffer. If it's killed mid-request, the curl process still runs to completion (or fails), and the sentinel fires against a process buffer whose contents depend on how far the transfer got.

> I think this might be a good opportunity to fix #1234 while we're at it.

Agreed: I've revised the PR. Both sentinels (`gptel-curl--stream-cleanup` and `gptel-curl--sentinel`) now check `process-exit-status` early. When curl exits with a non-zero status, they short-circuit with `"Curl failed with exit code %d. See curl manpage for details."` and skip response parsing entirely.

The `search-backward` NOERROR fix is kept as defense-in-depth (also in `gptel-curl--parse-response`), in case the UUID is ever absent despite a zero exit status.

karthink:
@benthamite As usual, thank you for the high quality PR!  Tweaked slightly (changed ifs it if-nots) and merged.

Issue #1294, opened 2026-03-16
Deprecated model llama3-70b-8192


andreas-roehler:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

README.md, Groq,  sugggests deprecated model llama3-70b-8192 

### Backend

None

### Steps to Reproduce

Checked for possibly other deprecated models at Groq

Found Deprecated: llama3-70b-8192

Groq error: ((HTTP/2 400) invalid_request_error) The model `llama3-70b-8192` has been decommissioned and is no longer supported. Please refer to https://console.groq.com/docs/deprecations for a recommendation on which model to use instead.

PR should follow

### Additional Context

GNU Emacs 31.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.50, cairo version 1.18.4) of `2026-03-15```

### Backtrace

```emacs-lisp
My checks: 

(defvar groq-deprecated-vs-alternative-models
  '((meta-llama/llama-4-maverick-17b-128e-instruct . openai/gpt-oss-120b)
    (meta-llama/llama-guard-4-12b . openai/gpt-oss-safeguard-20b)
    (playai-tts . canopylabs/orpheus-v1-english)
    (moonshotai/kimi-k2-instruct . moonshotai/kimi-k2-instruct-0905)
    (gemma2-9b-it . llama-3.1-8b-instant)
    (llama3-70b-8192 . llama-3.3-70b-versatile)
    (distil-whisper-large-v3-en . whisper-large-v3-turbo)
    (mistral-saba-24b . qwen/qwen3-32b)
    (qwen-qwq-32b . qwen/qwen3-32b)
    (llama-guard-3-8b . meta-llama/llama-guard-4-12b)
    (llama-3.2-1b-preview . llama-3.1-8b-instant))
  "List models deprecated at Groq alongside with their alternavtives.
See https://console.groq.com/docs/deprecations")

;; (car groq-deprecated-vs-alternative-models) ==> (meta-llama/llama-4-maverick-17b-128e-instruct . openai/gpt-oss-120b)

(defvar groq-current-deprecated-models
  '(meta-llama/llama-4-maverick-17b-128e-instruct
        meta-llama/llama-guard-4-12b
        playai-tts
        moonshotai/kimi-k2-instruct
        gemma2-9b-it
        llama3-70b-8192
        distil-whisper-large-v3-en
[...]

PR #1295, opened 2026-03-16
#1294, fix: Deprecated model ‘llama3-70b-8192’ removed

andreas-roehler:
Removes deprecated model ‘llama3-70b-8192’ - alternative model already in README.


karthink:
Thank you @andreas-roehler!

PR #1317, opened 2026-03-26
Update project homepage

pabl0:
The project has a rather attractive homepage at gptel.org, which I believe serves as a more appealing landing page for end users compared to GitHub. This is presented in Emacs when using `describe-package`. Additionally, the GNU project, FSF, and RMS disapprove of GitHub due to concerns about computing freedom, so this is better for non-gnu ELPA listing.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -6,7 +6,7 @@
 ;; Version: 0.9.9.4
 ;; Package-Requires: ((emacs "27.1") (transient "0.7.4") (compat "30.1.0.0"))
 ;; Keywords: convenience, tools
-;; URL: https://github.com/karthink/gptel
+;; URL: https://gptel.org
 
 ;; SPDX-License-Identifier: GPL-3.0-or-later
 


karthink:
Yes, the plan was to move the project "landing page" to gptel.org
eventually.  It's just that the GitHub page is more immediately relevant
for most users.  If you visit gptel.org, the README with instructions and
the issue tracker are both one more click away now, and you also need to
scroll down to find them.

There is no creature more impatient than an Emacs user who has installed a
package and wants to start using it.  Or one who has discovered a
problem and is looking for a quick fix.  I have seen new gptel users figure
out the package in real time and their frustration is palpable. ("I don't
want to read words, just give me the use-package block to copy!")

GNU guidelines notwithstanding, my goal is to get users the resource they
need with as little indirection as possible.

karthink:
Closing this for now, but will come back to it before gptel v1.0.  A few more things need to be in place on the website:

- Updated description and videos
- Completed gptel manual (also included as an Info manual in gptel)
- Links moved up to the top

PR #1327, opened 2026-04-02
Treat gpt-5.4-pro and o3-pro as reasoning models in gptel--request-data

djr7C4:
My PR neglected to add these models to the list in gptel--request-data. This
corrects that oversight.

### DIFF
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -225,15 +225,10 @@ Mutate state INFO with response metadata."
   (let ((prompts-plist
          (list :model (gptel--model-name gptel-model)
                :messages (vconcat prompts)
-               :stream (or gptel-stream :json-false)))
-        (reasoning-model-p ; TODO: Embed this capability in the model's properties
-         (memq gptel-model '(o1 o1-preview o1-mini o3-mini o3 o4-mini
-                                gpt-5 gpt-5-mini gpt-5-nano gpt-5.1 gpt-5.2
-                                gpt-5.3-chat-latest gpt-5.4 gpt-5.4-mini
-                                gpt-5.4-nano))))
+               :stream (or gptel-stream :json-false))))
     (when gptel-stream
       (plist-put prompts-plist :stream_options '(:include_usage t)))
-    (when (and gptel-temperature (not reasoning-model-p))
+    (when gptel-temperature
       (plist-put prompts-plist :temperature gptel-temperature))
     (when gptel-use-tools
       (when (eq gptel-use-tools 'force)
@@ -241,14 +236,11 @@ Mutate state INFO with response metadata."
       (when gptel-tools
         (plist-put prompts-plist :tools
                    (gptel--parse-tools backend gptel-tools))
-        (unless reasoning-model-p
-          (plist-put prompts-plist :parallel_tool_calls t))))
+        (plist-put prompts-plist :parallel_tool_calls t)))
     (when gptel-max-tokens
       ;; HACK: The OpenAI API has deprecated max_tokens, but we still need it
       ;; for OpenAI-compatible APIs like GPT4All (#485)
-      (plist-put prompts-plist
-                 (if reasoning-model-p :max_completion_tokens :max_tokens)
-                 gptel-max-tokens))
+      (plist-put prompts-plist :max_tokens gptel-max-tokens))
     (when gptel--schema
       (plist-put prompts-plist
                  :response_format (gptel--parse-schema backend gptel--schema)))


karthink:
This reasoning-model check should no longer be necessary in `gptel-openai`, as the OpenAI models should be using `gptel-openai-responses`.  The flag `reasoning-model-p` is a misnomer, what it's actually checking is something like `responses-api-p`.  

1. That said, you can still access these models using proxies like Openrouter using `gptel-openai`, so I don't know what happens when you access (say) `o3` using Openrouter.  Does it fully support the completions API fields?

2. Actually, it should no longer be necessary in `gptel-openai-responses` either, but we'll have to check how the Responses API handles `temperature` and `parallel_tool_calls`.

djr7C4:
Should those models be added in gptel-openai-responses.el instead then? There's a similar memq form there. 

I'm not sure exactly how openrouter handles it as I haven't used it for a while.

djr7C4:
The memq form in responses is also missing some values such as gpt-5.4-mini and nano.

karthink:
#### In gptel-openai.el

The reasoning-model-p check in gptel-openai.el is not  required any more when accessing `api.openai.com`.  But we should check accessing `gpt-5.*` and `o*` models with a couple of proxies (like Openrouter) before removing or fixing it.

#### In gptel-openai-responses.el

The `reasoning-model-p` check in gptel-openai-responses.el may also not be required any more, or it may need to be tweaked.  This (wrongly named) flag is checking three things:

1.  Whether `gptel-temperature` is applied: I think that in the Responses API, the temperature can be set for all `gpt-*` models, but not the `o*` models.  So this should be a check specifically for `o*`.

2. Whether `parallel_tool_calls` can be set: we'll have to look into how this works for `gpt-*` and `o*` models in the Responses API.

3. Whether max tokens should be specified as `max_completion_tokens` or `max_tokens`: In the Responses API, it should always be specified as `max_completion_tokens`, so no check is needed.

karthink:
I updated gptel-openai-responses.el.  A special flag is still required, but only for handling `gptel-temperature` with the o-series models.  You can continue to focus on gptel-openai.el for this PR.

djr7C4:
Testing shows that

    (gptel-make-openai "OpenRouter"
        :host "openrouter.ai"
        :endpoint "/api/v1/chat/completions"
        :stream t
        :key #'djr-openrouter-get-key
        :models '(gpt-5.4 gpt-5.4-pro gpt-5.4-mini gpt-5.4-nano gpt-5 gpt-5-mini gpt-5-nano o4-mini o1-preview o3-mini o3 o1))

works with the models listed (except for o1-preview which is no longer available on OpenRouter) and the reasoning-model-p check bound to nil. It looks like the check can be removed at least for OpenRouter. I don't have accounts with the other proxies so I didn't test those.

karthink:
Did you try (i) removing the `reasoning-model-p` check and (ii) specifying `gptel-temperature` when using an `o*` model in Openrouter?  I expect that this might be a problem.

djr7C4:
Yes, `C-u C-c RET I` showed that the temperature was included in the the request which was

    (:model "o3" :messages
            [(:role "system" :content
                    "You are an assistant in Emacs.")
             (:role "user" :content [(:type "text" :text "test")])]
            :stream t :stream_options (:include_usage t) :temperature 1.0)

djr7C4:
It should be checked with `gptel-tools` enabled as well though.

djr7C4:
I tested again with tools enabled and `reasoning-model-p` bound to `nil`. The models tested on OpenRouter were gpt-5.4, gpt-5, gpt-5-nano, o3 and o1. I also tested DeepSeek models deepseek-chat and deepseek-reasoner as I have an account there and it also uses the gptel-openai.el implementation. All of the queries completed correctly. I also tested  tools calls for some of the models (gpt-5.4, o3, deepseek-chat and deepseek-reasoner). This also worked correctly.

This is an example query (created with C-u C-c RET I) for reference. The tool enabled was WebSearch from gptel-agent.

    (:model "o3" :messages
            [(:role "system" :content
                    "You are an assistant in Emacs. Use markdown liberally to structure responses. When the language is not specified and isn't clear from the context, assume that it should be Emacs Lisp.")
             (:role "user" :content [(:type "text" :text "testing...")])]
            :stream t :stream_options (:include_usage t) :temperature 1.0 :tools
            [(:type "function" :function (:name "WebSearch" :description "Search the web for the first five results to a query.  The query can be an arbitrary string.  Returns the top five results from the search engine as a list of plists.  Each object has the keys `:url` and `:excerpt` for the corresponding search result.
    
    This tool uses the Emacs web browser (eww) with its default search engine (typically DuckDuckGo) to perform searches. No API key is required.
    
    If required, consider using the url as the input to the `Read` tool to get the contents of the url.  Note that this might not work as the `Read` tool does not handle javascript-enabled pages."
                                                :parameters
                                                (:type "object" :properties
                                                       (:query
                                                        (:type "string" :description
                                                               "The natural language search query, can be multiple words.")
                                                        :count
                                                        (:type "integer"
                                                               :description
                                                               "Number of results to return (default 5)"))
                                                       :required ["query"]
                                                       :additionalProperties
                                                       :json-false)))]
            :parallel_tool_calls t)

This was a good excuse to start exploring LLM tool use in Emacs which I hadn't had a chance to dive into before. The gptel implementation for tools seems quite slick.

Anyway, it looks to me like we can safely remove the check at least for these proxies. I probably won't check any others as I don't have accounts on them.

karthink:
@djr7C4 You can edit the PR or can I close it and remove the check myself.

djr7C4:
It's updated now.

karthink:
Merged, thanks again for the PR!

We'll hear soon if the reasoning-model-p check was required for other proxies, I can add it back then (probably in a more limited/targeted form).

djr7C4:
Sounds good!

PR #1332, opened 2026-04-07
gptel: allow anonymous plist parents in gptel--apply-preset

kmontag:
`gptel--apply-preset` previously always tried to resolve parents using `gptel-get-preset` (via the `cl-flet`'ed `get-preset`), which throws a `user-error` for anonymous (i.e. plist) parent presets.

This appears to have been introduced in https://github.com/karthink/gptel/commit/c3d5d7f758b8fbb071abbd18d599d1eacb0f2fbd.

The `get-preset` wrapper is still needed for parent resolution (to avoid overwriting `gptel--preset` with a parent symbol), so this fix adds a type check rather than removing the call.

Here's a simple repro:

```emacs-lisp
;; -*- lexical-binding: t; -\_-
;;
(require 'use-package)
(use-package
  gptel
  :demand t
  ;; Or wherever.
  :load-path "elpaca/sources/gptel")

(let\* ((base '(:system "base "))
        (main (list :parents (list base) :system '(:append "main"))))
       ;; Without fix: (user-error "gptel preset \"(:system base )\": Cannot find preset")
       ;; With fix: "base main"
       (gptel--apply-preset main)
       (message "gptel--system-message: %S" gptel--system-message))
```

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -2572,73 +2572,73 @@ two arguments, the symbol being set and the value to set it to.  It
 defaults to `set', and can be set to a different function to (for
 example) apply the preset buffer-locally."
   (unless setter (setq setter #'set))
-  (cl-flet ((get-preset (preset)
-              (or (gptel-get-preset preset)
-                  (user-error "gptel preset \"%s\": Cannot find preset"
-                              preset))))
+  (let* ((spec (if (memq (type-of preset) '(string symbol))
+                   (or (gptel-get-preset preset)
+                       (user-error "gptel preset \"%s\": Cannot find preset"
+                                   preset))
+                 preset)))
+    (when-let* ((func (plist-get spec :pre))) (funcall func))
+    (when-let* ((parents (plist-get spec :parents)))
+      (mapc (lambda (parent) (gptel--apply-preset parent setter))
+            (ensure-list parents)))
+    ;; Record preset name for persistence and UI display
     (when (memq (type-of preset) '(string symbol))
-      (let ((spec (get-preset preset)))
-        (funcall setter 'gptel--preset preset)
-        (setq preset spec)))
-    (when-let* ((func (plist-get preset :pre))) (funcall func))
-    (when-let* ((parents (plist-get preset :parents)))
-      (mapc (lambda (parent) (gptel--apply-preset (get-preset parent) setter))
-            (ensure-list parents))))
-  (map-do
-   (lambda (key val)
-     (pcase key
-       ((or :parents :description :pre :post) nil)
-       ((or :system :system-message :rewrite-directive)
-        (let ((sym (if (eq key :rewrite-directive)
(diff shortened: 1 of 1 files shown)


karthink:
@kmontag thanks for the PR.  I feel like this has gotten complicated and difficult to understand.

I tried to simplify it a little bit here.  Do you think we can do better?
``` emacs-lisp
(cl-flet ((preset-spec (preset)
            (if (memq (type-of preset) '(string symbol))
                (or (gptel-get-preset preset)
                    (user-error "gptel preset \"%s\": Cannot find preset"
                                preset))
              preset)))
  ;; Record preset name for persistence and UI display
  (when (memq (type-of preset) '(string symbol))
    (funcall setter 'gptel--preset preset))
  ;; Ensure that preset is a plist spec
  (setq preset (preset-spec preset))
  (when-let* ((func (plist-get preset :pre))) (funcall func))
  (when-let* ((parents (plist-get preset :parents)))
    (mapc (lambda (parent) (gptel--apply-preset (preset-spec parent) setter))
          (ensure-list parents))))
```

karthink:
Another option is to record the preset name into `gptel--preset` _after_ applying the parent presets:

``` emacs-lisp
(setq spec (if (memq (type-of preset) '(string symbol))
               (or (gptel-get-preset preset)
                   (user-error "gptel preset \"%s\": Cannot find preset"
                               preset))
             preset))
(when-let* ((func (plist-get spec :pre))) (funcall func))
(when-let* ((parents (plist-get spec :parents)))
  (mapc (lambda (parent) (gptel--apply-preset parent setter))
        (ensure-list parents)))
;; Record preset name for persistence and UI display
(when (memq (type-of preset) '(string symbol))
  (funcall setter 'gptel--preset preset))
;; Work with `spec' instead of preset in the rest of this function
;; ...
```
No `cl-flet` needed.  Now `gptel--preset` is set to each parent in turn, but eventually to the original preset as desired.

kmontag:
Nice, yeah I think the latter option seems cleaner.

I guess there's a slight wrinkle in that during preset application, and/or after a preset application that throws an error, `gptel--preset` would potentially be set to an interim value from one of the parents. Then again, maybe this is actually the most self-consistent behavior, i.e. the user probably expects the parents list to be applied literally before the child gets applied.

karthink:
Okay.  Can you modify your PR to the latter implementation?

kmontag:
done - I put `spec` in a let block, so the function is also another level indented.

karthink:
LGTM, merged!  Thanks for catching that.

karthink:
I had to revert this change in 8d6411b5 because in setting `gptel--preset` multiple times, oneshot preset application was broken again.  This is what PR #1325 was intended to fix in the first place.

Anonymous parent presets should be handled correctly now.

kmontag:
Nice one, thanks!

PR #1336, opened 2026-04-08
Fix non-streaming parsing for backends that split tool_calls across choices

agwblack:
Some OpenAI-compatible backends (e.g. GitHub Copilot) return non-streaming responses where text content and tool_calls are spread across separate choices entries rather than combined in a single choices[0].message.

gptel--parse-response for gptel-openai only examines choices[0], so it finds the text content but misses the tool_calls entirely.  :tool-use is never set in info, and the response is treated as a plain text completion -- the tools never run.

This is particularly damaging for sub-agent (task) tool calls: the agent's introductory text is returned as the final result, and the caller sees an empty/useless response.

Fix: Collect tool_calls from all choices and merge them into a single injected prompt message.  For standard backends where everything is in choices[0], behavior is unchanged.

### DIFF
--- a/gptel-openai.el
+++ b/gptel-openai.el
@@ -276,7 +276,8 @@ information if the stream contains it."
   "Parse an OpenAI (non-streaming) RESPONSE and return response text.
 
 Mutate state INFO with response metadata."
-  (let* ((choice0 (map-nested-elt response '(:choices 0)))
+  (let* ((choices (plist-get response :choices))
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR.  This looks like a duplicate of #1284, so please focus there instead.  I will be closing this PR unless you think this isn't a duplicate.

agwblack:
Thanks, sorry I missed that one. We can close this

PR #1338, opened 2026-04-09
Add gptel-reasoning-effort and add it to the transient menu

djr7C4:
This PR implements adds a gptel-reasoning-effort variable and adds it to the
transient menu. It currently only supports OpenAI's responses API. I haven't
tested the gptel-org.el code yet and suspect that it is broken but I thought I'd
send what I have to see if you think the approach is reasonable before I invest
more effort.

Selecting reasoning effort from the transient menu is working for OpenAI models
through the responses API. If the approach seems good, I'll fix any issues with
the implementation in gptel-org.el and implement it for Gemini and Anthropic as
well. Adding it to gptel-openai.el is also possible but I'd have to investigate
how reasoning effort works for non-OpenAI models on platforms like OpenRouter.


djr7C4:
Also, the `:reasoning-effort` forms in the model definitions are valid types for `cl-typep`. That's why the first value is `member`. `cl-typep` also supports integer intervals which will be helpful for implementing reasoning effort for Anthropic and Gemini. When I do that, I'll also update `gptel--transient-read-reasoning-effort` to enforce the number read to the valid range (currently it accepts anything).

karthink:
Thanks for the PR @djr7C4, I'll take a look soon.

djr7C4:
@karthink No worries if you're just busy but I wanted to make sure this hadn't fallen through the cracks.

karthink:
> @karthink No worries if you're just busy but I wanted to make sure this hadn't fallen through the cracks.

Not at all, just busy.  On average I can only focus on one new feature -- self-coded or PR review -- per week.  This week that slot went to the now merged token tracking system and token usage UI

(gptel development is steady but slower than the pace people usually have patience for. I've merged PRs and implemented requested features as long as 14 months after they were submitted, long after the interested parties had moved on!)

karthink:
It's not clear what the type of `gptel-reasoning-effort` is.  I see various sets of values -- high/low in Deepseek, none/low/high/medium/high/xhigh in OpenAI Responses, numeric values(?) in Anthropic, maybe other types too?

If this is the case we'll have to enforce a consistent type in gptel and translate its value to what various APIs expect internally.  Otherwise this is not a worthwhile upgrade over using `:request-params` in the backend or model, and things will break if you switch backends.  It also makes using `gptel-reasoning-effort` confusing for users.

We should decide on the spec for `gptel-reasoning-effort` before implementing more backend-dependent bespoke functionality.

djr7C4:
> Not at all, just busy. On average I can only focus on one new feature -- self-coded or PR review -- per week. This week that slot went to the now merged token tracking system and token usage UI

Fair enough. The number of issues and PRs for this project is kind of crazy. I appreciate you taking the time to look at this. The token tracking feature is something I've been wanting for a while.

> It's not clear what the type of gptel-reasoning-effort is. I see various sets of values -- high/low in Deepseek, none/low/high/medium/high/xhigh in OpenAI Responses, numeric values(?) in Anthropic, maybe other types too?

Yes. For Anthropic and Google it could be either a symbol or an integer. Older Google and Anthropic models used to use an integer to specify the number of thinking tokens allowed but they have switched to different schemas for their newer models. The different types reflect the API choices of the providers. 

> If this is the case we'll have to enforce a consistent type in gptel and translate its value to what various APIs expect internally.

My main use case for this feature was to be able to interactively experiment with different reasoning effort values interactively using `completing-read`. This feature would be significantly less useful to me if I were only able to access a subset of the reasoning effort levels exposed by the API. I've often found significant differences in reasoning quality between `high`and `xhigh` for example but `xhigh` is also much slower.

> Otherwise this is not a worthwhile upgrade over using :request-params in the backend or model, 

I disagree. Without this there isn't a nice way to set `:request-params` interactively or to discover the allowed values for reasoning effort interactively. If I completed the UI as I originally intended, all allowed values would be fully discoverable interactively by the user. To me that is quite valuable. 

> and things will break if you switch backends.

Yes. This is an issue but I was willing to accept it to get full interactive control over the reasoning effort. See the proposal below for a way around it though.

> We should decide on the spec for gptel-reasoning-effort before implementing more backend-dependent bespoke functionality.

Would it be acceptable to allow generic values for the reasoning effort that are translated while also allowing model-specific values? I could add a configuration variable to disable model specific values by default but allow the user to optionally enable them (perhaps even allowing them to toggle this from the transient menu).

One way to do this would be to set the reasoning effort separately for each model. In other words, if you changed the model-specific reasoning effort for gpt-5.5 but then switched to opus-4.7, the reasoning effort would revert to the generic general value assuming you hadn't set a level specifically for opus-4.7.

karthink:
> Would it be acceptable to allow generic values for the reasoning effort that are translated while also allowing model-specific values? I could add a configuration variable to disable model specific values by default but allow the user to optionally enable them (perhaps even allowing them to toggle this from the transient menu).
> 
> One way to do this would be to set the reasoning effort separately for each model. In other words, if you changed the model-specific reasoning effort for gpt-5.5 but then switched to opus-4.7, the reasoning effort would revert to the generic general value assuming you hadn't set a level specifically for opus-4.7.

I think this is going to cause confusion for anyone who uses more than one gptel-backend type.  For now I suggest something like this:

- `gptel-reasoning-effort` is an enum, taking one of the values `none` (or `nil`), `low`, `medium`, `high`, `xhigh`.
- Setting `gptel-reasoning-effort` to `nil` turns off thinking for the request (if possible).
- Add a defcustom `gptel-reasoning-effort-alist` that maps these enum values to specific token counts:

``` emacs-lisp
(defcustom gptel-reasoning-effort-alist
  '((low    . 256)
    (medium . 1024)
    (high   . 4096)
    (xhigh  . 16384))
  "Mapping from reasoning levels to reasoning token budgets.

gptel supports specifying reasoning effort (via
`gptel-reasoning-effort') as discrete levels (low, medium etc).  If the
LLM expects reasoning effort to be provided as token budgets instead,
this mapping is used to specify the corresponding token counts."
  :type '(alist :key-type symbol :value-type natnum)
  :group 'gptel)
```

- The only other fallback behavior needed is for models that don't support one of these levels, like `xhigh`.  In that case gptel should switch to the next available (lower) level and issue a warning when running `gptel-request`.

Does this design cover most of the cases you're thinking of?

djr7C4:
> Does this design cover most of the cases you're thinking of?

Unfortunately not. For my use case, I need the provider specific values. The values allowed are also constantly changing and vary widely for different models and providers even when they are enums. For instance, some models now support `max` as well.

I'm also not sure how it would work for the DeepSeek models which also either high or max reasoning effort but also allow it to be disabled. I suppose it's equivalent to none, high and max for other models but there would need to be some translation.

karthink:
> Unfortunately not. For my use case, I need the provider specific values

I don't understand.  Are these values levels (like low/high), token counts or something else?

djr7C4:
Not all providers support setting a numeric token limit. Many including OpenAI and DeepSeek only allow enum values for reasoning effort but not necessarily only the ones in the list. The values supported also change frequently.

karthink:
> Not all providers support setting a numeric token limit. 

This is not an issue in the proposed design.  The numeric limits from `gptel-reasoning-effort-alist` are only used if they _require_ numeric limits and don't support the enum values.

> Many including OpanAI and DeepSeek only allow enum values for reasoning effort but not necessarily only the ones in the list. The values supported also change frequently.

Maybe I shouldn't have called it an enum -- you are free to set `gptel-reasoning-effort` to whatever symbol you want.  If the model/backend 
- supports the symbol you set, send it as is
- supports other symbols but not the one you set, ignore it and issue a warning.
- supports only numeric limits, you try to find the limit from `gptel-reasoning-effort-alist`.
  - If there is a match, you send that number.
  - If there isn't one, you ignore it and issue a warning.

The user is also free to add whatever associations they want to `gptel-reasoning-effort-alist`, including new symbol keys.

karthink:
Actually, I guess you can go further and allow `gptel-reasoning-effort` to be a natnum too.  Then:

- If `gptel-reasoning-effort` is a `natnum` but the model expects a symbol/level, you find the closest (reverse) match for that number in `gptel-reasoning-effort-alist` and send that symbol.

djr7C4:
@karthink 

Yes, this new system you suggest seems quite powerful and would cover my use cases. I have a few ideas to improve it further. 

1. Natural numbers might mean fairly different things for different models since the maximum number of thinking tokens can vary a lot. To solve this floats could be allowed as well. These would be interpreted as a percentage of the maximum number of reasoning effort tokens. When a float is not found, the closest value in the alist would be used like for natural numbers.

2. Symbols and lists of symbols could also be allowed in the alist. This would allow the user to define associations like gptel-max --> (max xhigh 1.0 high) or gptel-min --> (none min minimal 0.0 low) which for a particular model should use the first value in the list that is valid for the current model. In this way, the user could define generic reasoning effort values that would work for just about any model.

djr7C4:
> Setting gptel-reasoning-effort to nil turns off thinking for the request (if possible).

IMO, `nil` should use the API default instead. This matches the current behavior in master without support for reasoning effort. If you really want `nil` to disable it instead, perhaps there should be a special symbol such as `default`.

karthink:
> IMO, `nil` should use the API default instead. This matches the current behavior in master without support for reasoning effort. If you really want `nil` to disable it instead, perhaps there should be a special symbol such as `default`.

Yes, that makes sense.  There should also be a value that turns off reasoning (assuming the API supports it).  Maybe `none`, but I can see `nil` vs `none` being confusing.

karthink:
>     1. Natural numbers might mean fairly different things for different models since the maximum number of thinking tokens can vary a lot. To solve this floats could be allowed as well. These would be interpreted as a percentage of the maximum number of reasoning effort tokens. When a float is not found, the closest value in the alist would be used like for natural numbers.

Sounds good, but what happens if it's a float and the API does not have a maximum or if the maximum is not known?  For example, I can imagine users running `(gptel-make-openai "Openrouter" ...)` and not specifying the max reasoning effort model/backend metadata.

>     2. Symbols and lists of symbols could also be allowed in the alist. This would allow the user to define associations like gptel-max --> (max xhigh 1.0 high) or gptel-min --> (none min minimal 0.0 low) which for a particular model should use the first value in the list that is valid for the current model. In this way, the user could define generic reasoning effort values that would work for just about any model.

I think this sounds reasonable.  The default value of `gptel-reasoning-effort-alist` should cover all the common cases.

But the viability of this approach really comes down to having accurate per-model metadata about the (actual) reasoning effort enums supported by each model or backend.  Keeping this updated is going to be work.

djr7C4:
> Sounds good, but what happens if it's a float and the API does not have a maximum or if the maximum is not known? For example, I can imagine users running (gptel-make-openai "Openrouter" ...) and not specifying the max reasoning effort model/backend metadata.

In that case, I would emit a warning and use the API default instead as for the other cases when something isn't supported.

djr7C4:
> I think this sounds reasonable. The default value of gptel-reasoning-effort-alist should cover all the common cases.

Agreed.

> But the viability of this approach really comes down to having accurate per-model metadata about the (actual) reasoning effort enums supported by each model or backend. Keeping this updated is going to be work.

Yes, it will be. FWIW it's worth I've already been maintaining it in my personal fork for the providers I use (OpenAI, Anthropic, Gemini and DeepSeek).

djr7C4:
> Yes, that makes sense. There should also be a value that turns off reasoning (assuming the API supports it). Maybe none, but I can see nil vs none being confusing.

I'd suggest `disabled` instead. Using a `default` symbol instead also seems reasonable.

FrauH0lle:
A bit of feedback: I tested this PR/branch the last week with models from OpenAI, Deepseek and Anthropic and did not run into any issues.

djr7C4:
It's good to know that it worked for you.

djr7C4:
@karthink I've added support for `gptel-reasoning-effort-alist` as discussed above. The symbol `disabled` is used to disable reasoning effort when supported by the API. I've done some testing and it seems to work well. I'm using this branch locally so I'll fix any problems that arise in my daily usage. There are a number of fallbacks that issue warnings but continue when something is not directly supported.

PR #1339, opened 2026-04-09
Fix streaming responses

Azkae:
LLM responses are no longer streamed for me. I bisected the issue to commit 85126d3.

It seems that curl buffers the response when stdout is a pipe. Adding `-N` (`--no-buffer`) to `gptel-curl--common-args` fixes the issue for me.

I'm not sure whether this option should be added on Windows, I cannot test it.

Thank you for your work on gptel!

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -724,11 +724,16 @@ See `gptel-backend'."
   :type '(repeat string))
 
 (defconst gptel-curl--common-args
-  (if (memq system-type '(windows-nt ms-dos))
-      '("--disable" "--location" "--silent" "-XPOST"
-        "-y7200" "-Y1" "-D-")
+  (cond
+   ((memq system-type '(windows-nt ms-dos))
+    '("--disable" "--location" "--silent" "-XPOST"
+      "-y7200" "-Y1" "-D-"))
+   ((eq system-type 'darwin)
     '("--disable" "--location" "--silent" "--compressed"
       "-XPOST" "-y7200" "-Y1" "-N" "-D-"))
+   (t
+    '("--disable" "--location" "--silent" "--compressed"
+      "-XPOST" "-y7200" "-Y1" "-D-")))
   "Arguments always passed to Curl for gptel queries.")
 
 (defvar gptel--link-type-cache nil
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -728,7 +728,7 @@ See `gptel-backend'."
       '("--disable" "--location" "--silent" "-XPOST"
         "-y7200" "-Y1" "-D-")
     '("--disable" "--location" "--silent" "--compressed"
-      "-XPOST" "-y7200" "-Y1" "-D-"))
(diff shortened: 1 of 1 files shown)


karthink:
That's odd, streaming is still working here.  Could you let me know the OS and version of Curl you're using?

ryankask:
I can confirm that setting  `(setopt gptel-curl-extra-args '("-N"))` fixed the issue for me.

```
❯ curl -V
curl 8.7.1 (x86_64-apple-darwin25.0) libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.68.0
Release-Date: 2024-03-27
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL threadsafe UnixSockets
```

I also tried with homebrew's curl but it also didn't work.

Azkae:
I'm also on macOS

```
curl 8.7.1 (x86_64-apple-darwin24.0) libcurl/8.7.1 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.12 nghttp2/1.64.0
Release-Date: 2024-03-27
Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM SPNEGO SSL threadsafe UnixSockets
```

karthink:
I can't reproduce this on Linux, so maybe we can set this in `gptel-curl--common-args` but only on MacOS?

Azkae:
Sure, added a check to apply `-N` only on darwin

pabl0:
Curl 8.7.1 is somewhat outdated; typical Linux systems often feature more recent versions (Debian Trixie includes 8.14.1), which may behave differently. 

Is there any downside to adding `-N` across all platforms?

I (still) wonder about Windows.

karthink:
> Is there any downside to adding -N across all platforms?

I'm not sure, so I want to limit the fix to the platform from where the problem is reported.

@pabl0, any idea why this behavior showed up when we started reading the request headers via STDIN?

karthink:
@Azkae Thanks for the fix!

Azkae:
I may be wrong, but I believe that's because before commit 85126d398c800dfed80a91aac990a5dc0d15f9a6 `start-process` defaults to a pty as connection type and now we provide `:connection-type 'pipe` to `make-process`.

Perhaps curl is checking stdout with isatty(1) to enable or disable buffering by default.

karthink:
Oh right, I forgot we changed the connection type. That's almost certainly
the reason.

Issue #1342, opened 2026-04-11
Use Codex subscription


berenddeboer:
Is it possible for gptel to use my Codex subscription? This is explicitly allowed by Codex (but not by Anthropic, so that's why I switched).

Does it need a PR or are there ways to do this? For example opencode supports this, and many other tools.

karthink:
@berenddeboer Not yet, although there is PR #1260 which folks report is working for them.

It's not merged yet because (i) I need to review it carefully, as it's vibe-coded, and (ii) There are too many oauth implementations in gptel and I want to unify them first.

bjodah:
Just in case it would be of any value: based on #1260, I put together a branch which simply added Codex support:
- https://github.com/bjodah/gptel/tree/openai-responses-api-chatgpt-codex3

I then tasked these models (individually) to attempt to deduplicate the OAuth parts:

- https://github.com/bjodah/gptel/pull/3
- https://github.com/bjodah/gptel/pull/4
- https://github.com/bjodah/gptel/pull/5

To my novice elisp eyes Gemini 3.1 Pro did the better job (as you can see in the PR, number of lines actually was reduced by about 20 lines when you account for comments), albeit I'm really not a qualified judge the quality/maintainability of the final code. Anyhow, I've been using the branch by Gemini 3.1 Pro daily for three weeks now.

karthink:
@berenddeboer Added preliminary support for Codex in 56e5b06d107287890ace32a81ddbf5df13493e58.  Very early, so it's not documented in the README/NEWS yet.  You can use it as `(gptel-make-openai-oauth "MyOpenAISub")`.

Closing as completed but feel free to report bugs and suggestions in new issues.

Issue #10, opened 2023-03-10
Response error nil


minad:
After some conversation chatgpt just stopped abruptly.

> Moreover, the probabilistic nature of quantum mechanics is often at odds with our classical intuition, and the emergence of a classical world from the underlying quantum mechanics is still not fully understood. The decoherence approach provides a valuable framework for understanding this emergence, but it

Now I am getting `Response error nil` errors in the header line and it does not seem possible to recover. Did you observe this issue too? If we get #7 it should be possible to just restart.

karthink:
No, I've not encountered this error. It would be good to know the actual response/error from ChatGPT in this case. Assuming you're using curl, could you evaluate this function and try sending a query?

```elisp
(defun gptel-curl--sentinel (process status)
  "Process sentinel for GPTel curl requests.

PROCESS and STATUS are process parameters."
  (let ((proc-buf (process-buffer process)))
    (with-current-buffer proc-buf
        (clone-buffer "*gptel-error*" 'show))
    (if-let* ((ok-p (equal status "finished\n"))
              (proc-info (alist-get process gptel-curl--process-alist))
              (proc-token (plist-get proc-info :token))
              (content (gptel-curl--parse-response proc-buf proc-token)))
        (funcall (plist-get proc-info :callback) content)
      ;; Failed
      (funcall (plist-get proc-info :callback) nil))
    (kill-buffer proc-buf)))
```

> If we get #7 it should be possible to just restart.

The API is completely stateless so there is no concept of restarting. Every query sent is as fresh as any other.

minad:
> Assuming you're using curl, could you evaluate this function and try sending a query?

Yes, I have curl but I don't know if it used by gptel. Is this automatic? Anyway, I closed the session, but I will evaluate your code the next time the issue occurs.

> The API is completely stateless so there is no concept of restarting. Every query sent is as fresh as any other.

But what about the prompts? Do you sent them all the time to gpt, such that it has access to the context of the conversation?

karthink:
Curl is chosen automatically if `executable-find` finds it. (It works more reliably than `url-retrieve`, see #4.)

> But what about the prompts? Do you sent them all the time to gpt, such that it has access to the context of the conversation?

[Unfortunately yes](https://platform.openai.com/docs/guides/chat/introduction).

> Because the models have no memory of past requests, all relevant information must be supplied via the conversation. If a conversation cannot fit within the model’s token limit, it will need to be shortened in some way.

[...]

PR #1349, opened 2026-04-15
Extract gptel-oauth.el from gptel-gh.el

bjodah:
First step towards #1260 / #1342.

AI Disclosure: This is the `gptel-oauth.el` from bjodah/gptel#5 which means that the elisp that's new here has primarily been written by LLMs. I have read through the implementation, made some minor edits, but I'm still an elisp novice.

### DIFF
--- a/gptel-oauth.el
+++ b/gptel-oauth.el
@@ -1,0 +1,128 @@
+;;; gptel-oauth.el --- OAuth utilities for gptel  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2023-2026  Karthik Chikmagalur
+
+;; SPDX-License-Identifier: GPL-3.0-or-later
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Provides common OAuth 2.0 utilities for gptel backends: device flow
+;; prompting, PKCE (RFC 7636), base64url encoding, URL parameter encoding,
+;; JWT payload parsing, browser authorization-code flow, and token
+;; persistence.
+
+;;; Code:
+
+(require 'browse-url)
+(require 'url-http)
+(require 'gptel-request)
+
+;;; Token Storage
+
+(defun gptel-oauth--write-token (file token)
+  "Write TOKEN plist to FILE."
+  (let ((print-length nil)
+        (print-level nil)
+        (coding-system-for-write 'utf-8-unix))
+    (make-directory (file-name-directory file) t)
+    (write-region (prin1-to-string token) nil file nil :silent)
+    token))
+
+(defun gptel-oauth--read-token (file)
+  "Read token plist from FILE."
+  (when (file-exists-p file)
+    ;; Set coding system to auto-dos so files with CR EOL read properly.
+    (let ((coding-system-for-read 'utf-8-auto-dos))
+      (with-temp-buffer
+        (set-buffer-multibyte nil)
+        (insert-file-contents-literally file)
+        (goto-char (point-min))
+        (condition-case nil
+            (read (current-buffer))
+          (error nil))))))
+
+;;; PKCE Implementation
+
+(defun gptel-oauth--base64url-encode (str)
+  "Base64url-encode raw string STR (no padding)."
+  (let ((b64 (base64-encode-string str t)))
+    (setq b64 (replace-regexp-in-string "+" "-" b64))
+    (setq b64 (replace-regexp-in-string "/" "_" b64))
+    (replace-regexp-in-string "=+$" "" b64)))
+
+(defun gptel-oauth--base64url-decode (str)
+  "Decode Base64URL string STR, adding padding if necessary."
+  (let* ((s (replace-regexp-in-string "-" "+" str))
+         (s (replace-regexp-in-string "_" "/" s))
+         (pad (% (length s) 4)))
+    (when (> pad 0)
+      (setq s (concat s (make-string (- 4 pad) ?=))))
+    (decode-coding-string (base64-decode-string s) 'utf-8 t)))
+
+(defun gptel-oauth--generate-code-verifier ()
+  "Generate PKCE code verifier using `random'."
+  (let ((chars "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~"))
+    (apply #'string
+           (cl-loop repeat 128
+                    collect (aref chars (random (length chars)))))))
+
+(defun gptel-oauth--generate-code-challenge (verifier)
+  "Generate PKCE code challenge from VERIFIER."
+  (gptel-oauth--base64url-encode
+   (secure-hash 'sha256 verifier nil nil t)))
+
+;;; OAuth Flow
+
+(defun gptel-oauth--device-auth-prompt (user-code verification-uri)
+  "Prompt the user to authorize device using USER-CODE and VERIFICATION-URI.
+
+Copies USER-CODE to the clipboard and conditionally opens a browser."
+  (let ((in-ssh-session (or (getenv "SSH_CLIENT")
+                            (getenv "SSH_CONNECTION")
+                            (getenv "SSH_TTY"))))
+    (ignore-errors (gui-set-selection 'CLIPBOARD user-code))
+    (if in-ssh-session
+        (progn
+          (message "Device Code: %s (copied to clipboard)" user-code)
+          (read-from-minibuffer
+           (format "(One-time code %s copied) Visit %s in your local browser, \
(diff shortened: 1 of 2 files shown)


gptel-oauth.el
```diff
@@ -0,0 +1,118 @@
...
+(defun gptel-oauth-restore-token (file)
+  "Restore token plist from FILE."
+  (when (file-exists-p file)
+    ;; Set coding system to auto-dos so files with CR EOL read properly.
+    (let ((coding-system-for-read 'utf-8-auto-dos))
+      (with-temp-buffer
+        (set-buffer-multibyte nil)
+        (insert-file-contents-literally file)
+        (goto-char (point-min))
+        (condition-case nil
```
bjodah, line 53: The condition-case here is new when comparing with `gptel-gh.el`, if my understanding is correct: it will silence an error from e.g. the file being empty.

gptel-oauth.el
```diff
@@ -0,0 +1,118 @@
...
+          (when data
+            (encode-coding-string
+             (cond
+              ((string-prefix-p "application/json" content-type)
+               (gptel--json-encode data))
+              (t data))
+             'utf-8)))
+         (url-request-extra-headers
+          `(("Content-Type" . ,content-type)
+            ("Accept" . "application/json")
```
bjodah: When asking Gemini 3.1 Pro for a code review of the PR it pointed out that "url.el already sets `Accept`", but having multiple "Accept" entries is "probably fine". I guess that analysis is apt?

djr7C4:
Writing the token to disk seems like a security risk and also isn't consistent with how API authentication is currently done (:key in arguments to gptel-make-* functions and gptel-api-key). IMO it would be better and more consistent to extend API key handling to allow oath tokens as well.

karthink:
> Writing the token to disk seems like a security risk and also isn't consistent with how API authentication is currently done (:key in arguments to gptel-make-* functions and gptel-api-key). IMO it would be better and more consistent to extend API key handling to allow oath tokens as well.

See #1058.

I would like to encrypt the token at rest, of course.  The reason I haven't done it yet is that unlike decrypting API keys, it should happen without user intervention, and I haven't found a cross-platform way to do that.

When the user stores their API key (non-oauth) in `auth-source`, say, they expect to be asked for authentication when running `gptel-send`.  The API keys are as secure as the user wanted them to be, and it's up to them to make gpg or some other program available.

If the user authenticates with oauth, then the authentication process is "finished" after the browser round trip, and it did not require gpg or a keyring.  If the user is then asked to enter a passphrase whenever they run `gptel-send` with a Copilot backend, they will be very confused (and annoyed).  And this is assuming gpg is present on their system.

I did some searching for possible solutions, here's what I remember finding:

- `plstore` is built into emacs, but requires either a gpg key set up or a symmetric encryption passphrase to be entered semi-frequently.
- `auth-source` is (of course) built-in, but
  - the gpg backend has the same issues as plstore.
  - Requirements for other backends like age and sops are not commonly present on user systems
  - password-store and `auth-source-pass` are not suited to transient data like tokens (and also not guaranteed to be present).
  - On Gnome/KDE and MacOS, auth-source can use the session keychain (Secret Service API or equivalent) for transparent token en/decryption.  But there is no solution on Windows, or on window managers/compositors without a keyring daemon.

The only solution I could think of was to store the token using (say) auth-source, but cache the decrypted token until Emacs exits.  But that still leaves the problem of no encryption method (and no applicable auth-source backend) being available.

djr7C4:
Calling GPG every time isn't so bad if you have a gpg-agent instance running. That way you don't have to enter your password every time but only periodically. I use that method already for API keys (via auth-source).

The way I like to handle tokens in CLI programs is to store them in an encrypted password manager. Most CLI programs I use that require them allow you to set them via an environment variable and I use a wrapper to do that securely when I run them (in a way that doesn't leak them to every running process). 

CLI agents like Claude and Codex seem to also store tokens insecurely. I suppose it's not quite as bad when it's a subscription since you won't be charged extra if the token is stolen. It's a bigger issue for cloud services like AWS and Digital Ocean because then you could get a nasty bill.

djr7C4:
What about allowing the user to provide a function like for API keys? That way they can control authentication if they want and bypass storing tokens in plaintext on disk. I guess it would be a kind of a pain since you'd still have to authenticate with the browser and manually copy the token into encrypted storage.

karthink:
> Calling GPG every time isn't so bad if you have a gpg-agent instance running.  That way you don't have to enter your password every time but only periodically

You shouldn't have to enter a password at all if you're using oauth.  No browser asks you for a password to retrieve some authentication token periodically after you log into an account.

> The way I like to handle tokens in CLI programs is to store them in an encrypted password manager.

Yes, there are many options if you're sufficiently technically inclined.  But these are different for every user, and it's not clear what a safe default should be that is guaranteed to work for every gptel user on every platform.

> What about allowing the user to provide a function like for API keys? That way they can control authentication if they want and bypass storing tokens in plaintext on disk.

This is sort of what #1058 does.  But the default is still to store it as plaintext, because there is no encryption solution that will work uniformly.  The kind of user who will scan gptel's more advanced options and set up a custom function to cache oauth tokens securely probably also doesn't need the extra protection because they're generally careful in their computer use.

djr7C4:
I suppose it's already exposed anyway if one is using Claude Code or Codex (which is likely with a subscription) so perhaps it's somewhat moot. Those are much more likely to be targeted by an attacker than gptel.

karthink:
@djr7C4 I took a brief look at `gptel-oauth.el`.  It looks like a good start.  A couple of notes:

1. There is already a helper function to retrieve JSON in gptel: it's `gptel--url-retrieve`, which the code in gptel-gh was using.  Not sure why you removed it, but it should be used unless there's some reason not to.
2. There are more steps in OAuth.  For example, connecting to Anthropic's Claude subscription -- while it was possible -- requires [PKCE authentication, implemented here](https://share.karthinks.com/gptel-anthropic-oauth.el).  I haven't looked into how the Codex or Gemini plan authentication works yet.

djr7C4:
@karthink Thanks for taking a look but this actually isn't my PR. I am interested in the feature and may use it though.

bjodah:
Thank you for the feedback. I've pushed a gpt-5.4 + opus-4.6 commit to address this. I want to review this myself carefully later today.

<details>
<summary>EDIT: about ~40 lines can be shaved off gptel-anthropic-oauth.el it seems</summary>

```
$ wc -l gptel-anthropic-oauth.el
301 gptel-anthropic-oauth.el
$ wc -l gptel-anthropic-oauth-new.el
264 gptel-anthropic-oauth-new.el
```

</details>

EDIT2:
I'm trying something new for my own learning here, I've put up two draft PRs that tries to address the feedback here, one based on Gemini 3.1 Pro, and one on GPT-5.4 + Opus-4.6. The parts which are similar enough I'll backport into this PR, what's left can then be seen in the diff views of:
- https://github.com/bjodah/gptel/pull/9
- https://github.com/bjodah/gptel/pull/10

EDIT3: ~~I still have some work left to do with manually pruning the two approaches so I can get a better view in how they differ. I'll write a new comment in this thread once I've finished that work.~~ ✅

karthink:
> @karthink Thanks for taking a look but this actually isn't my PR. I am
> interested in the feature and may use it though.
> 

@djr7C4 Sorry for the ping!  I had multiple PR buffers open, including yours, and must have gotten confused.

bjodah:
@karthink I'm a bit uncertain on what the best approach is with respect to modifications in `gptel--url-retrieve` (in `gptel-request.el`).

If you look in the diff view of the PRs `#9` and `#10` of my fork (the links in my previous post) you will find one solution involving adding optional arguments, and another which inspects content-type in the headers. Would you prefer one over the other (or neither?).

In the main comment of each PR, I've pasted `grep` output showing how this extended functionality is utilized in the new chatgpt-codex-oauth backend.
(openai codex's authorization make use of "application/x-www-form-urlencoded", see e.g. [this](https://github.com/anomalyco/opencode/blob/5eaef6b758cbfb683deadb8d440a420c3c1ee1f8/packages/opencode/src/plugin/codex.ts#L116))

karthink:
As far as the calling convention for `gptel--url-retrieve` goes, PR 10 looks more general and better.

But looking at the diffs, both LLM-generated approaches look messy to me.

karthink:
@bjodah Do you plan to add more to this PR?  I'm trying to see if I can merge this now and work on adding oauth support for codex/OpenAI subscriptions

bjodah:
@karthink I also thought PR 10 looked like the better approach if we don't want a parallel alternative implementation of `gptel--url-retrieve` where `content-type` is e.g. an explicit argument. I agree that the code looks messy, I'm afraid my Elisp-fu is not up to the task of writing something beautiful (...one day, hopefully). So no, I don't think I'm planning on adding anything more to this PR (feel free to close it if you like). I could merge PR 10 into this branch if you'd like, but I think you or someone more experienced with elisp would need to push some clean-up commits to give a cleaner implementation?

Getting codex into gptel master's branch would be wonderful. The LLM-deduplicated versions of your anthropic connector and the vibe coded codex backend are attached as txt-file attachments in respective PR 9 & 10 (done so in the interest of keeping the diffs small, adding the actual codex implementation could then be made in a subsequent PR after a `gptel-oauth.el` one such as this one). The `wc` output indicated that deduplicating into a common `gptel-oauth.el` indeed is a net benefit as you suggested.

karthink:
I merged this PR.  It's a good enough base for me to work with, thank you!

karthink:
> Getting codex into gptel master's branch would be wonderful.

Done, see 56e5b06d107287890ace32a81ddbf5df13493e58.

Took a few hours but I daresay gptel-openai-oauth.el is simpler and more maintainable than what the LLMs generated.

bjodah:
Really nice @karthink, your solution was indeed a lot more elegant, inspiring to see!

PR #1353, opened 2026-04-16
ensure newline between text and tool

Azkae:
Ensure that a newline is inserted between a text & a tool. 

Sometimes claude returns a text between two tool invocation. Currently the second tool would be on the same line as the text without a newline in between:

```
#+begin_tool (...)
#+end_tool
A text describing the tool use: #+begin_tool (...)
#+end_tool
```

This renders the second tool block invalid in org mode since it's not at the beginning of a line.

This PR ensure that we insert a separator if there is no carriage return before a tool invocation.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -2084,10 +2084,7 @@ for tool call results.  INFO contains the state of the request."
                                  "\n")) ;start of response
                            ((not tool-marker) gptel-response-separator)
                            ((and (not (= tracking-marker tool-marker))
-                                 (save-excursion ;not consecutive tool result blocks
-                                   (goto-char tool-marker)
-                                   (skip-chars-forward " \r\t\n")
-                                   (>= (point) tracking-marker)))
+                                 (not (eq (char-before tracking-marker) ?\n)))
                             gptel-response-separator)))
                     (tool-use
                      ;; TODO(tool) also check args since there may be more than


karthink:
@Azkae thanks for the PR.  This is actually a regression caused by 72dd69453887f96c72688102e08565c2f94c8931, which was addressing the same problem but failed to do it.

The new cond bock and the one immediately before it need to be compared with each other carefully (they do closely related things).  It is possible that we don't need two cond blocks here.

karthink:
Basically, the behavior we want is the following:
1. If a tool bock is directly following another tool block, don't insert `gptel-response-separator`.  The single required newline at the end of a tool block is part of the tool block itself.
2. If a tool block is following response text, if there is already a newline at the end of that text, don't insert `gptel-response-separator` (This is what the faulty commit was trying to enforce).
3. Otherwise insert `gptel-response-separator`.

Azkae:
From my understanding:

```
#+begin_tool (...)
#+end_tool  ← tool_marker
              tracking_marker ↴ 
A text describing the tool use:#+begin_tool (...)
#+end_tool
```

```
((and (not (= tracking-marker tool-marker))
      (save-excursion ;not consecutive tool result blocks
        (goto-char tool-marker)
        (skip-chars-forward " \r\t\n")
        (>= (point) tracking-marker)))
 gptel-response-separator)
```

This cond bock checks if there is only separator/whitespace between the previous tool marker and the new tool, in which case it adds another separator. That doesn't make sense to me, maybe I'm missing something.

I removed the previous cond, tried a bunch of responses and everything seems to be correct.

Azkae:
Maybe we could update the cond like so:

```
((and (not (= tracking-marker tool-marker))
      (not (eq (char-before tracking-marker) ?\n))
      (not (save-excursion
             (goto-char tool-marker)
             (skip-chars-forward " \r\t\n")
             (>= (point) tracking-marker))))
 gptel-response-separator)
```

To handle the case where the llm prints only whitespace/newline between tool blocks, but i'm not sure this really happens.

EDIT: actually, after thinking about this, this wouldn't work if the llm output "\n " between tools, we still need a newline before the tool. We could remove the space but I don't think we want to do that.

karthink:
> EDIT: actually, after thinking about this, this wouldn't work if the llm output
> "\n " between tools, we still need a newline before the tool. We could remove
> the space but I don't think we want to do that.

We actually can't remove the space under the current design.

The current state of the `cond` looks good to me, I'll merge this.  The only other change to think about is if we want to insert `gptel-response-separator` between responses and tool calls, or just a single `\n`.  Right now the spacing between responses and tool calls can be inconsistent:

1. Response part ends in one-or-more `\n`: one-or-more `\n` between the response and tool block
2. Response part ends without a `\n`: `gptel-response-separator` newlines between response and tool block, which defaults to 2.

karthink:
Merged, thanks for the PR @Azkae!

PR #1357, opened 2026-04-17
fix: move preset capf description to company-compatible location

### LINKED ISSUE #1356: Use company-compatible attributes for `gptel-preset-capf`.
`gptel-preset-capf` puts the agent `:description` in the `:annotation-function` slot, but since many description strings are long, this causes many backends to look really bad.

Image

The most common convention is to use `:company-doc-buffer` instead. `company-mode` is common, and most other buffer completion packages are compatible with it. Emacs built-ins supply these attributes (e.g., see `elisp-completion-at-point`).

I experimented with attempting to make `gptel-agent-capf` sensitive to completion packages, but that gets complicated and I think it's simpler to use the convention.

Cerebus:
Closes #1356 

Example w/ corfu:

image

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -2798,11 +2798,18 @@ Add this to `completion-at-point-functions'."
              (list (point) (+ (point) num)
                    gptel--known-presets
                    :exclusive 'no
-                   :annotation-function
-                   #'(lambda (c) (thread-first
-                              (intern-soft c)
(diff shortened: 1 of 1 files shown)


karthink:
As discussed in #1356, please retain the `:annotation-function` with truncated descriptions in addition to `:company-doc-buffer`.

karthink:
@Cerebus gentle ping

PR #1362, opened 2026-04-19
Add Opus 4.7 as an Anthropic model.

djr7C4:
This PR also corrects Sonnet 4.6's context size to 1M as per the current
Anthropic documentation.


karthink:
Merged, thank you!

PR #1363, opened 2026-04-19
gptel-request: Add -N flag to curl on all platforms to fix regression

aagit:
The -N flag (no-buffering) was previously added for macOS in 4747ffa251af84dbb6fe181f5d2b484ec2e442a7, but
85126d398c800dfed80a91aac990a5dc0d15f9a6 caused a regression on Linux as well. I assume it's safer to add it on all platforms, not just Linux and Mac, just in case.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -727,13 +727,13 @@ See `gptel-backend'."
   (cond
    ((memq system-type '(windows-nt ms-dos))
     '("--disable" "--location" "--silent" "-XPOST"
-      "-y7200" "-Y1" "-D-"))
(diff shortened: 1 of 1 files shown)


Stebalien:
I can confirm this fixes "choppy" output on my Linux box (curl 8.18.0).

karthink:
Merged, thanks @aagit.

There is no difference between the MacOS and Linux Curl command lines after this patch, but I retained the three-way `cond` anyway, just in case we need to revert a change.

aagit:
Hi @karthink, yes I didn't write it, but that's exactly why I left it there too. Thanks.

PR #1364, opened 2026-04-19
gptel-request: make it possible to abort a single request

Stebalien:
This change makes it possible to abort a single request by passing an FSM.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -2258,33 +2258,33 @@ Initiate the request when done."
     (unless (plist-get info :dry-run) (gptel--fsm-transition fsm))
     fsm))
 
-(defun gptel-abort (buf)
-  "Stop any active gptel process associated with buffer BUF.
+(defun gptel-abort (buf-or-fsm)
+  "Abort a gptel request associated with BUF-OR-FSM.
 
-BUF defaults to the current buffer."
+When BUF-OR-FSM is a buffer, abort the most recent gptel requests
+associated with that buffer.
+
+When BUF-OR-FSM is a `gptel-fsm' (e.g., returned by `gptel-request'),
+abort that request.
+
+BUF-OR-FSM defaults to the current buffer."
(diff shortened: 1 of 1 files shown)


karthink:
Thanks for the PR.  I haven't looked at the code yet, but the idea of aborting a specific request by FSM makes sense.

2 is incorrect, `gptel-abort` should abort only the latest active request in the buffer.  So calling it three times should abort the three latest active requests, in turn.  If that's not happening it's a separate bug.

Stebalien:
> 2 is incorrect, gptel-abort should abort only the latest active request in the buffer. So calling it three times should abort the three latest active requests, in turn. If that's not happening it's a separate bug.

That's what it did, but the docs said:

> Stop any active gptel process associated with buffer BUF.

TBH, I think "stop everything" makes more sense but I'm also happy to just fix the docs.

Stebalien:
Eh, nvm. I can see it both ways. I've updated it to stop the first request it finds.

However, I do have to ask: is it possible to have multiple requests outstanding associated with a single FSM? If so, this logic isn't quite correct.

(I aslo won't be offended if you want to do this your own way, I ended up refactoring the logic a bit and I don't want to impose my opinions on your code)

karthink:
> However, I do have to ask: is it possible to have multiple requests outstanding
> associated with a single FSM? If so, this logic isn't quite correct.

No, one FSM == one request at a time.  It can be a multi-turn request if there are tool calls.

Stebalien:
Ah, got it. Then the PR should be correct as-is.

karthink:
I'm not sure I like the idea of a union type (buf-or-fsm) for abort here, it feels like a hack.  I'll need to think some more about this.  But it seems like splitting this into two functions, an interactive command `gptel-abort` that takes a buffer as an argument, that in turn calls a non-interactive function `gptel--abort-fsm` that takes an fsm as an action, might be cleaner.

PR #1367, opened 2026-04-20
gptel: fix gptel-with-preset when :parents is the first keyword

Stebalien:
When :parents comes first in the preset definition, `syms` is nil here so `(nconc syms ...)` doesn't update `syms`.

To reproduce:

```elisp
(gptel-make-preset 'parent
  :system "parent system prompt")
(gptel-make-preset 'child
  :parents '(parent)
  :use-tools t)
(gptel--preset-syms 'child)
```

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -2658,7 +2658,7 @@ PRESET is the name of a preset, or a spec (plist) of the form
         ((or :description :pre :post))
         (:parents
          (mapc (lambda (parent-preset)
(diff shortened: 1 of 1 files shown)


karthink:
I don't think this is an issue anymore (#1355)

Stebalien:
Ah... Elpaca doesn't update when you have a branch checked out...

sorry about that.

PR #1368, opened 2026-04-20
gptel-request: escape carriage returns in string literals

Stebalien:
Instead of including literal CR characters, use \r. Git has a tendency to mess with carriage return characters in text files on unix (i.e., every time I switch branches).

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -750,7 +750,7 @@ binary-encoded.")
 ;; Since we want this known at compile time, when markdown-mode is not
 ;; guaranteed to be available, we have to hardcode it.
 (defconst gptel-markdown--link-regex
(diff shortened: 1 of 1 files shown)


karthink:
Sorry, I don't understand why this is needed -- this looks like an encoding issue unrelated to gptel?

Stebalien:
I thought it was git's built-in CRLF line conversion, but it looks like it's a bug in the mergiraf merge driver. IMO, it's still better to use the escaped characters but I'll close this for now and let you decide.

Stebalien:
Lol, it looks like codeberg itself has issues displaying standalone CR characters:

https://codeberg.org/winterschon/emacs-llm-gptel/src/commit/76a0c1686192bc43e56b8f5e4be11736bb9806da/gptel-curl.el#L208-L210

(that's a gptel fork on codeberg, just as an example)

PR #1369, opened 2026-04-20
gptel-preset: allow system directives to be nil

Stebalien:
Without this, emacs signals `gptel--apply-preset` signals if the preset's `system-message` (or `system`) is nil because `nil` is a valid symbol but isn't a known directive name


karthink:
Merged, thanks!

PR #1371, opened 2026-04-21
gptel-org: Lazy-load gptel in set-properties

### LINKED ISSUE #1370: gptel-org commands fail with void-variable gptel--preset when invoked before gptel.el is loaded
### Update verification

- [x] I have updated gptel to the latest commit and tested that the issue still exists (master `4747ffa`, v0.9.9.4-93-g4747ffa).

### Bug description

`gptel-org.el` uses several variables defined in `gptel.el` (`gptel--preset`, `gptel-backend`, `gptel-model`, `gptel--system-message`, …) but forward-declares them with bare `(defvar SYM)` forms (lines 33–42) and never `(require 'gptel)`. In gptel's usual flow this is fine — `gptel.el` is always loaded first via `gptel-mode` / `gptel-send` / similar — but if a `gptel-org-*` command is autoloaded and invoked cold, `gptel-org.el` loads in isolation, the forward-declared variables remain unbound, and the command errors.

This surfaces in Doom Emacs, which binds `gptel-org-set-topic` / `gptel-org-set-properties` to leader keys via autoloads and hits `void-variable gptel--preset` on first invocation. Doom-side fix (autoloads for the commands): doomemacs/doomemacs PR forthcoming.

### Backend

Other (bug is independent of backend).

### Steps to reproduce

Using `emacs -Q`:

1. Save as `/tmp/repro.el`:
   ```elisp
   (add-to-list 'load-path "/path/to/gptel")
   (autoload 'gptel-org-set-properties "gptel-org" nil t)
   (setq debug-on-error t)
   ```
2. `emacs -Q -nw -l /tmp/repro.el`
3. `M-x org-mode RET`
4. `M-x gptel-org-set-properties RET`

### Additional context

- Emacs: GNU Emacs 30.2.50
- OS: macOS 26.4.1 (Darwin 25.4.0)
- curl: 8.18.0
- gptel: master `4747ffa` (v0.9.9.4-93-g4747ffa)

### Backtrace

```elisp
Debugger entered--Lisp error: (void-variable gptel--preset)
  (and gptel--preset (gptel-get-preset gptel--preset))
  (let ((preset-spec (and gptel--preset (gptel-get-preset gptel--preset))))
    (if preset-spec
        (org-entry-put pt "GPTEL_PRESET" (gptel--to-string gptel--preset))
      (org-entry-delete pt "GPTEL_PRESET"))
    ...)
  gptel-org-set-properties(148 t)
  funcall-interactively(gptel-org-set-properties 148 t)
  command-execute(gptel-org-set-properties record)
  execute-extended-command(nil "gptel-org-set-properties" "gptel-org-set-properties")
```

### Log information

N/A — failure happens before any API call.

### Suggested fix

Add `(require 'gptel)` near the top of `gptel-org.el` alongside the existing `cl-lib` / `org-element` / `outline` requires. The forward declarations at lines 33–42 can stay (as byte-compiler hints) or be removed.

nwjsmith:
## Summary

`gptel-org-set-properties` can be autoloaded and invoked cold (for example from Doom Emacs leader bindings). In that path, `gptel-org.el` loads without `gptel.el`, so the forward-declared gptel variables such as `gptel--preset` remain unbound and the command errors with `(void-variable gptel--preset)`.

Fixes #1370.

## Change

Lazy-load `gptel` inside `gptel-org-set-properties` before reading the active gptel configuration. This keeps `(require 'gptel-org)` lightweight while ensuring this command has the variables/functions it needs when called standalone.

## Verification

- [x] `(require 'gptel-org)` in isolation does not load `gptel`
- [x] Cold `gptel-org-set-properties` call lazily loads `gptel` and writes properties when gptel config is supplied after load
- [x] `emacs -Q --batch -L . -f batch-byte-compile gptel-org.el`

### DIFF
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -24,6 +24,7 @@
 
 ;;; Code:
 (require 'cl-lib)
+(require 'gptel)
 (require 'org-element)
 (require 'outline)
 (require 'mailcap)                    ;FIXME Avoid this somehow
--- a/gptel-org.el
+++ b/gptel-org.el
@@ -607,6 +607,7 @@ gptel model and backend names, the system message, active tools, the
 response temperature, max tokens and number of conversation turns to
 send in queries.  (See `gptel--num-messages-to-send' for the last one.)"
   (interactive (list (point) t))
+  (require 'gptel)
   (let ((preset-spec (and gptel--preset (gptel-get-preset gptel--preset))))
     (if preset-spec
         (org-entry-put pt "GPTEL_PRESET" (gptel--to-string gptel--preset))


karthink:
> But when a `gptel-org-*` command is autoloaded and
> invoked cold, `gptel-org.el` loads in isolation, the forward-declared variables
> remain unbound...

Is the problematic command `gptel-org-set-properties`?  Is there anything else?  I'm not sure that I want to load all of gptel unless it's needed.  I can work around it in this command.

nwjsmith:
@karthink as far as I can would you be more comfortable with lazy loading within `gptel-org-set-properties`? i.e.

```elisp
   (defun gptel-org-set-properties (pt &optional msg)
     ; ...
     (require 'gptel)
     ; ...
     )
```

karthink:
@nwjsmith That should be fine, let's do that for now!  I need to revisit the loading order before the 1.0 release, I've made a note of this PR for then.

nwjsmith:
@karthink all set!

karthink:
Merged, thanks!

PR #1372, opened 2026-04-22
Add post-insert hooks for tool and reasoning blocks

### LINKED ISSUE #1358: Post-render hook with bounds for inserted tool result blocks
I’m prototyping an external package that uses `gptel` as the rendering surface for a different tool-using runtime, and one small seam I keep wishing for is a public post-render hook for tool result blocks.

Right now, if I want to do anything with the exact block `gptel` inserted — for example:

- add overlays
- style diffs or subregions
- attach markers/metadata
- tweak the visible body while keeping gptel’s fenced container

— I end up looking at `gptel--display-tool-results`.

Would you be open to a small hook here? Something like `gptel-after-display-tool-result-functions`, with access to the tool/spec, args, result, `info`, buffer, and ideally the inserted block bounds.

I’m not proposing any new default UI. This would just give downstream packages a supported way to post-process the rendered tool block instead of advising internals.

This may be adjacent to #1092, but the specific ask here is: once `gptel` has already inserted the tool result block, is there a public extension point for operating on that exact region?

If this seems reasonable, I’d be happy to draft a small PR.

### LINKED ISSUE #1359: Tiny public seam around reasoning auto-folding
I’m experimenting with an external frontend package that uses `gptel` for rendering, and one small thing I keep running into is reasoning auto-folding.

Right now the display policy for inline reasoning seems fixed to:

- show while streaming
- fold when the block closes

That default makes sense, but if I want to keep reasoning expanded in some sessions, the only route I’ve found is advising `gptel--display-reasoning-stream`.

Would you be open to a tiny public seam here? I think either of these would be enough:

1. a small option like `gptel-auto-fold-reasoning` (default `t`)
2. or a hook that runs when the reasoning block closes

I’m not asking for a bigger reasoning UI — just a supported way for downstream packages to opt out of the final auto-fold behavior without patching internals.

If that seems reasonable, I’d be happy to draft a minimal PR in whichever form you’d prefer.

oneingan:
This adds two small UI hooks:
- `gptel-post-tool-insert-functions`
- `gptel-post-reasoning-insert-functions`

Each is called with the beginning and end positions of the inserted block.

The idea is to provide a supported post-render seam for packages that want to act on gptel's rendered tool and reasoning blocks, without extending `gptel-request` or the tool-call hooks.

The reasoning hook runs after the block has been closed and folded.

Closes #1358
Closes #1359

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -156,6 +156,12 @@
   results, block the tool call but continue the request, or stop the
   request entirely.
 
+- New hooks ~gptel-post-tool-insert-functions~ and
+  ~gptel-post-reasoning-insert-functions~ run after gptel inserts tool
+  and reasoning blocks into a buffer.  These are UI-layer hooks, and
+  each function is called with the beginning and end positions of the
+  inserted block.
+
 - New variable ~gptel-bedrock-aws-cli-command~ to set the path to the
   AWS CLI command for the Bedrock backend.  Defaults to "ews".
 
--- a/README.org
+++ b/README.org
@@ -1660,6 +1660,12 @@ In dedicated chat buffers: you can additionally customize =gptel-prompt-prefix-a
 
 For more custom formatting, use =gptel-pre-response-hook= and =gptel-post-response-functions=.
 
+If you want to act on rendered reasoning or tool blocks, use
+=gptel-post-reasoning-insert-functions= or
+=gptel-post-tool-insert-functions=.  These are UI hooks: they run when
+gptel inserts these blocks into a buffer, such as via =gptel-send=,
+and are not part of the =gptel-request= API.
+
 #+html: </details>
 #+html: <details><summary>
 **** How does gptel distinguish between user prompts and LLM responses?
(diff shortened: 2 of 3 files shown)


karthink:
Instead of searching for the tool block start and end, it would be simpler to just record these positions beforehand?  Insertion is controlled by gptel's callbacks anyway so this information should be available.

karthink:
@oneingan Please try to record these positions in the "forward pass" instead of searching for text in the buffer afterwards.  This is fragile and harder to update if the tool/reasoning block format changes in the future.

oneingan:
Thanks, updated. The hook bounds are now recorded during the insertion path instead of being rediscovered by searching the buffer afterwards:

- string insertion callbacks return the inserted text bounds
- reasoning/tool renderers derive block bounds from those forward insertion bounds
- removed the regex-based rendered block bound helpers

Checked with:
- `git diff --check upstream/master..HEAD`
- `emacs -Q --batch -L . -f batch-byte-compile gptel.el`
- batch smoke checks for non-streaming reasoning, streaming reasoning, and tool block hook bounds

PR #1374, opened 2026-04-23
gptel-transient: Update the transient menu after applying prefixes

### LINKED ISSUE #1373: The value shown in gptel-menu doesn't update when a preset uses :eval or :function

### Bug Description

The value shown in gptel-menu doesn't update when a preset uses :eval or :function. However, the updated value is saved to the variable. I used the following to test.

    (gptel-make-preset 'test-preset
      :max-tokens '(:eval 1000))
    
    (gptel-make-preset 'test-preset2
      :max-tokens '(:function (lambda (val) (max 10000))))

### Backend

None

### Steps to Reproduce

See above.

### Additional Context

GNU Emacs 30.2 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.52, cairo version 1.18.4)
Linux 6.19.12-arch1-1 #1 SMP PREEMPT_DYNAMIC Sat, 11 Apr 2026 22:36:29 +0000 x86_64 GNU/Linux

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

djr7C4:
This fixes #1373.

gptel-save-preset already updates the transient when the prefix is actually
saved but currently this isn't done when it is merely applied.

### DIFF
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -1055,7 +1055,9 @@ existing preset, as well."
                                         gptel--known-presets nil t)))))))
   (gptel--apply-preset preset
                        (or setter (lambda (sym val) (gptel--set-with-scope
-                                                sym val gptel--set-buffer-locally)))))
+                                                     sym val gptel--set-buffer-locally))))
+  (when transient--prefix
+    (transient-setup 'gptel-menu)))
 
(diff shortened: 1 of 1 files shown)


karthink:
Thanks, merged after "prefix" -> "preset" corrections.

Issue #1351, opened 2026-04-16
Ediff changes in gptel-rewrite are discarded


djr7C4:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

After running `gptel-rewrite` and submitting a prompt to rewrite the current region, the new text replaces the region. Pressing `RET` to bring up the `gptel-rewrite` menu and then `e` (to run ediff) brings up the `ediff` buffer as expected. However, moving through the diff regions using `n` from `ediff` and then pressing `a` (to select the original version of the difference region) has no effect after exiting `ediff` with `q`. Instead, the text in the buffer reflects what was returned by the LLM regardless of any changes made in `ediff`.

### Backend

None

### Steps to Reproduce

See above.

### Additional Context

GNU Emacs 30.2 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.50, cairo version 1.18.4)
Linux 6.17.4-arch2-1 #1 SMP PREEMPT_DYNAMIC Sun, 19 Oct 2025 19:21:18 +0000 x86_64 GNU/Linux

### Backtrace

```emacs-lisp
N/A
```

### Log Information

```json

```

karthink:
Thanks, I'll try to reproduce this.

I'm a little surprised to learn this because I've used gptel-rewrite + ediff + selective merging several times in the past without issues, and this code has not been touched in years.  Have I been merging text in the wrong direction all this time, I wonder 🤔

djr7C4:
Just to be sure, I tested it in a separate init directory (using emacs --init-directory) containing only the following init.el:

    (package-initialize)
    
    (defun djr-openai-get-key ()
      (auth-source-pass-get 'secret "<...>"))
    
    (use-package gptel
      :vc (:url "https://github.com/karthink/gptel.git"
           :branch "master"
           :rev :newest))
      
    (require 'gptel)
    (setq gptel-api-key #'djr-openai-get-key)

The behavior was the same as before.

djr7C4:
[...]

PR #1375, opened 2026-04-24
Enhance gptel--rewrite-ediff as per #1351

djr7C4:
gptel--rewrite-ediff has been changed so that the LLM response overlays are set
to the contents of buffer B when ediff quits. The contents of buffer A are still
the actual contents of the original buffer before the LLM edit (not the text
shown by the overlays). Additionally, `C-c C-a` and `C-c C-b` bindings have been
added to ediff to immediately quit and accept a buffer.

~The keymap should really be configurable but its inconvenient with the current
code since there isn't a convenient way to bind the local functions in
`defvar-keymap` form. It could be refactored of course but using several global
functions seems ugly compared to the current letrec implementation.~

Edit: a configuration option has now been added for the bindings.

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -422,6 +422,11 @@ SWITCHES are diff arguments."
         (setq-local diff-jump-to-old-file t))
       (display-buffer diff-buf))))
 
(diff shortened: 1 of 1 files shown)


djr7C4:
Fixes #1351

PR #1376, opened 2026-04-25
Add gpt-5.5 and gpt-5.5-pro

djr7C4:


karthink:
@djr7C4 Thanks, merged!  I moved the models further down the list, because we want the first model in the list to be a -mini variant because that's the fallback when no backend is defined or when there is a backend-model mismatch.

PR #1377, opened 2026-04-25
Add new DeepSeek backend configurations

beacoder:
Add deepseek v4 models: deepseek-v4-flash， deepseek-v4-pro。

### DIFF
--- a/gptel-openai-extras.el
+++ b/gptel-openai-extras.el
@@ -321,7 +321,17 @@ The Deepseek API requires strictly alternating roles (user/assistant) in message
                      :capabilities (tool)
                      :context-window 128
                      :input-cost 0.28
-                     :output-cost 0.42))))
+                     :output-cost 0.42)
+		    (deepseek-v4-flash
+                     :capabilities (tool reasoning)
+                     :context-window 1000
+                     :input-cost 0.14
+                     :output-cost 0.28)
+                    (deepseek-v4-pro
+                     :capabilities (tool reasoning)
+                     :context-window 1000
+                     :input-cost 1.74
+                     :output-cost 3.48))))
   "Register a DeepSeek backend for gptel with NAME.
 
 For the meanings of the keyword arguments, see `gptel-make-openai'."


djr7C4:
The context window is 1M (so it should be 1000 in the code). Also, the prices are wrong for deepseek-v4-pro.

deepseek-reasoner and deepseek-chat are now aliases for deepseek-v4-flash and deepseek-v4-pro so the current values for those are also wrong. The DeepSeek docs state that deepseek-reasoner and deepseek-chat will be deprecated in the future. 

IMO deepseek-reasoner and deepseek-chat should be removed since they are now redundant.

beacoder:
you could check https://api-docs.deepseek.com/zh-cn/.
they mentioned: * deepseek-chat 与 deepseek-reasoner 两个模型名将于 2026/07/24 弃用。出于兼容考虑，二者分别对应 deepseek-v4-flash 的非思考与思考模式。
which means: deepseek-chat and deepseek-reasoner are all aliases for deepseek-v4-flash, not deepseek-v4-pro.

beacoder:
image

I don't see why the price is wrong?

djr7C4:
https://api-docs.deepseek.com/quick_start/pricing

djr7C4:
That version gives different prices. Not sure which one should be considered more authoritative.

karthink:
@beacoder it does look like the context-window should be 1000.  I'm not sure about the prices either.  Let me know when it's ready to merge, and please also add a NEWS entry since we're adding new models to the default configuration.

beacoder:
@karthink this pull request is ready. as for the NEWS entry, I have created another PR : https://github.com/karthink/gptel/pull/1379
reason for another PR is because I don't know how to modify the other file in this pull request.

karthink:
Merged, thanks!

yantar92:
This does not seem to be right. With the patch applied and model set of `deepseek-v4-pro`, I am getting "DeepSeek error: ((HTTP/2 400) invalid_request_error) The `reasoning_content` in the thinking mode must be passed back to the API.", which does not happen with `deepseek-reasoner`.

AFAIU, they also changed the API and the "aliases" are not merely aliases, but also compatibility layers to make old API work. According to https://api-docs.deepseek.com/guides/thinking_mode, you need to pass additional parameters to the model - thinking enabled/disabled, and also reasoning effort.

djr7C4:
I'm able to run `deepseek-v4-pro` without errors. The interface shows quite a bit of reasoning happening and it works for multiple turns too.

PR #1392, opened 2026-05-03
fix: Ensure utf-8 in tool call args

FrauH0lle:
* gptel-openai-responses.el (gptel-curl--parse-stream): Fix `json-value-p` error when a tool receives for example unicode characters as tool function arguments.


karthink:
Thanks!  The fact that `json-serialize` returns a unibyte string trips me up repeatedly.

PR #1400, opened 2026-05-05
gptel-rewrite: remove default face from overlay display

aagit:
A previous commit added the 'default' face to the overlay's display property. This causes the pre-rewrite text to look identical to the post-rewrite text, which prevents to see where the cursor is positioned during inference streaming.

The keyboard can't scroll and that can't be fixed in gptel, but at least the mouse wheel can scroll with scroll-preserve-screen-position so it'd be nice to retain the visibility of the inference progress while the text doesn't change.

Fixes: 992161fd06eb ("gptel-rewrite: Confirm tool calls from minibuffer")

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -556,7 +556,7 @@ INFO is the async communication channel for the rewrite request."
           (insert response)
           (unless (eobp) (ignore-errors (delete-char (length response))))
           (font-lock-ensure)
-          (overlay-put ov 'display (propertize (buffer-string) 'face 'default))))
+          (overlay-put ov 'display (propertize (buffer-string)))))
       (unless (plist-get info :stream) (gptel--rewrite-callback t info)))
 
      ((eq response 'abort)              ;request aborted


karthink:
The idea was that there is now a banner above the section being rewritten so additional indicators are not necessary (and may look too noisy).

I tried it out and it does indeed look very noisy to me -- and I still can't see the rewrite cursor.

To create a subtler but still clear distinction, I think it should be possible to add the `shadow` face to the text being rewritten, and the `default`  face to the new text so far.  These faces will be applied to the text in the temporary buffer used for the rewrite, and there will be no overlay face at all.  What do you think?

aagit:
Hi @karthink , when you say "I still can't see the rewrite cursor", actually I can see the rewrite cursor before it starts changing text, but only with this change applied.

You can see what I mean in this video at second 15:

https://gitlab.com/aarcange/ripgrep-edit-assets/-/raw/main/demo-follow_symlinks.webm

When the LLM starts writing all old text goes grayed out in non bright white, and the overwritten text is bright white.

It's a minor issue but I felt I missed the old behavior, it's just losing some information to use the same color for old and new text. However how you prefer to achieve a color or brightness difference it's up to you, as long as I can see some difference to infer where is the inference cursor, it's fine with me.

karthink:
I have another idea, not sure how to implement it yet -- we can add a fake cursor at interface between the rewritten and original text.  It's tricky to do since we are already displaying the entire region as an overlay's display property.  The fake cursor will have to be generated and moved using text-properties.

aagit:
I wouldn't mind either ways, but I'm wondering what's the pros. Isn't that more complicated than just using a different color for the overlay and the old text. I'm only concerned it would add more code.

I believe the main lacking UI feature relative to the overlay that warrants adding a lot more code is fixing the scrolling so it can be scrolled with the keyboards or/and it can do a "screen follow the cursor" during inference (line breaks permitting).

The overlay scrolling like mentioned before should be fixed in emacs, however if the fake cursor can workaround it in a maintainable way, it would be worth it...

PR #1401, opened 2026-05-05
don't truncate tool arguments during confirm

### LINKED ISSUE #1399: confirm is truncating the tool command

### Bug Description

I have some tools that use `:confirm t` but when I am given the tool to view, it is truncated and who knows what it's trying to run.

e.g.

```
(shell_command "grep -n \"calc\" ~/.emacs.d/init.el ~/.confi…")
```

the `...` should be showing me the full thing.

### Backend

None

### Steps to Reproduce

Run any tool with `:confirm t` enabled

### Additional Context

debian trixe (not relevant here)

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

fommil:
close #1399

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -2244,7 +2244,7 @@ NAME and ARG-VALUES are the name and arguments for the call."
                                (prin1-to-string
                                 (replace-regexp-in-string
                                  "\n" "⮐" (truncate-string-to-width
-                                           arg (floor (window-width) 2)
+                                           arg 256
                                            nil nil t))))
                               (t (prin1-to-string arg))))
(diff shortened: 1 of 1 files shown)


karthink:
@fommil I would prefer to just use a longer default value of the width instead of introducing a defvar here.  Tools can have custom previewers that bypass this issue entirely, so this will lead to some confusion eventually.

I suggest a default of 256 characters.  If it's still truncated you can always inspect the tool call in full.

fommil:
done

karthink:
Merged, thanks!

PR #1402, opened 2026-05-05
gptel-openai-extras: Fix the context windows for alias models

djr7C4:


karthink:
@djr7C4 Wasn't this changed just recently?  Sorry, I'm not using deepseek and I'm confused about the behaviors of the new v4 models.  Even if the models are aliases, the context window provided by the backend might be different for each alias depending on the API endpoint, right?  Is there official documentation that the context window of deepseek-reasoner/chat has changed?

djr7C4:
I suppose its possible that it is different. My impression from the docs was that the context windows are the same since it says

"(1) The model names deepseek-chat and deepseek-reasoner will be deprecated in the future. For compatibility, they correspond to the non-thinking mode and thinking mode of deepseek-v4-flash, respectively."

https://api-docs.deepseek.com/quick_start/pricing

karthink:
> "(1) The model names deepseek-chat and deepseek-reasoner will be deprecated in the future. For compatibility, they correspond to the non-thinking mode and thinking mode of deepseek-v4-flash, respectively."
> 
> https://api-docs.deepseek.com/quick_start/pricing

I'm not sure this means the context windows are the same.  I'd like to hold off until we can find some more concrete information.

karthink:
@djr7C4 Any idea if we have more accurate numbers now?

djr7C4:
It looks like it explicitly states 1 million for both now. The pricing has changed since this PR was opened though. I'll update it.

https://api-docs.deepseek.com/quick_start/pricing

djr7C4:
NVM, I think for chat and reasoner there is no new info.

djr7C4:
Since the aliases were removed in master, this PR is no longer relevant.

PR #1403, opened 2026-05-05
gptel-anthropic: Remove retired Anthropic models

djr7C4:
The models removed are either listed at https://platform.claude.com/docs/en/about-claude/model-deprecations as retired or verified as unavailable via calling the API.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,13 @@
 * 0.9.9.6-pre
 
 ** Breaking changes
+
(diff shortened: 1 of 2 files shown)


karthink:
@djr7C4 Please add an entry to NEWS as well.  You can use one of the past deprecation news entries for reference.

djr7C4:
Updated.

PR #1407, opened 2026-05-06
gptel-rewrite: Only use string-pixel-width on graphic display

### LINKED ISSUE #1406: Model name in rewrite header overlay overflows by 1 in terminal

### Bug Description

The right aligned model name in the rewrite header overlay consistently overflows by 1 character for me.

### Backend

None

### Steps to Reproduce

1. (this is in terminal over SSH)
2. Use gptel rewrite feature

### Additional Context

gptel version 0.9.9.5.0.20260505.145654

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

darkfeline:
This returns the string-width on non-graphic displays which necessitates the +1 so it doesn't overflow

Fixes https://github.com/karthink/gptel/issues/1406

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -216,7 +216,8 @@ which see."
            (propertize " Waiting..." 'face '(warning default)) ;status element 1
            (propertize                                         ;status element 2
             " " 'display
-            (if (fboundp 'string-pixel-width)
+            (if (and (fboundp 'string-pixel-width)
+                     (display-graphic-p))
                 `(space :align-to (- right (,(string-pixel-width hint-str))))
               `(space :align-to (- right ,(+ 1 (string-width hint-str))))))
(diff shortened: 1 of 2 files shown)


karthink:
I'm a little confused by the logic -- I don't remember why the +1 is required in the first place.  Any ideas?

darkfeline:
The history doesn't help, it was +1 for quite a while, and before that it was +2 oddly enough.

It looks like it might be for the line continuation character.  That is, if your terminal is 80 chars long, then a 80 char line will be wrapped (the last char gets pushed over for the line continuation char).

I tested that this also applies with `toggle-truncate-line`; the final char gets replaced with the "truncate indicator".

See https://www.gnu.org/software/emacs/manual/html_node/emacs/Continuation-Lines.html

karthink:
Merged with an additional fix, thanks!

PR #1408, opened 2026-05-07
gptel-request: Improve error messages on failed curl requests

thatismatt:
I'm using AWS Bedrock which sometimes returns a json body of `{"message":"..."}` this handles that case. Ultimately this means that the message in the minibuffer is now clearer, e.g.:

```
;; New message:
Bedrock error: (HTTP/2 403) The security token included in the request is expired
;; Old message:
Bedrock error: ((HTTP/2 403) Could not parse HTTP response.) Could not parse HTTP response.
```

There are two other locations where the same thing is done, so this might be needed. As yet I've not hit those code paths. But for reference they are in `gptel--url-parse-response` and `gptel-curl--steam-cleanup`.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -2678,7 +2678,9 @@ See `gptel-curl--get-response' for its contents.")
                 (string-trim resp))
               http-status http-msg))
        ((and-let* ((error-data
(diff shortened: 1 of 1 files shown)


karthink:
Updated and merged, thank you!

Issue #1388, opened 2026-05-01
bad request format when calling bedrock


fommil:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

(confirming I updated to latest git head `71b9f9414536c4cd35d8c02d3c8ad4fc4aea8718`)

I'm using the bedrock backend and i get the HTTP/2 error popping up a lot killing my session. This is the response I got

```
"HTTP/2 400 \r\ndate: Fri, 01 May 2026 12:27:19 GMT\r\ncontent-type: application/json\r\ncontent-length: 64\r\nx-amzn-requestid: REDACTED\r\nx-amzn-errortype: SerializationException:http://internal.amazon.com/coral/com.amazon.coral.service/\r\n\r"
{
  "gptel": "response body",
  "timestamp": "2026-05-01 13:27:19"
}
{
  "Message": "Start of structure or map found where not expected"
}
```

I've attached one such query that caused this (there's nothing confidential in here). The error sounds to me like it's some kind of JSON to BSON schema mapping issue?

### Backend

None

### Steps to Reproduce

send the attached request to aws. I haven't been able to minimise

### Additional Context

This is the bulk of my setup https://github.com/fommil/dotfiles/blob/master/.emacs.d/lisp/fommil-llm.el

with

```
(let ((models '(claude-opus-4-6 claude-sonnet-4-6)))
    (setq
     gptel-max-tokens 32768
     gptel-model (car models)
     gptel-backend (gptel-make-bedrock "claude"
                     :region "us-east-1"
                     :aws-profile "REDACTED"
                     :stream nil
                     :models models
                     :model-region 'us)))
```

[request.json](https://github.com/user-attachments/files/27275798/request.json)

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

fommil:
I would note that for some reason when using the tools I get this dumped out in the buffer, and I didn't expect to see this as it looks internal (I have debugging enabled, maybe that triggers this?)

````
### try again using the tools

Let me search for current tools and methods.

``` tool (web_search_exa :query "GitHub tool to bulk export Confluence pages to PDF using REST API" :numRes ...)
(:name "web_search_exa" :args (:query "GitHub tool to bulk export Confluence pages to PDF using REST API" :numResults 10))


... [redacted many many lines of tool output] ...


````

[...]

PR #1410, opened 2026-05-07
gptel-bedrock: Fix multi-turn tool use in gptel--parse-buffer

jhoodsmith:
### Problem

TLDR: The new pattern of including tool results in buffers doesn't work with Bedrock.

After a tool call, subsequent requests to the Bedrock backend are malformed.  `gptel--parse-buffer` reconstructs the conversation history by walking backwards through the buffer, but it only handles two text property values on buffer regions: `response` (→ `"assistant"`) and `nil` (→ `"user"`).

Tool call regions are propertized with `(tool . <id>)`, which has no matching arm in the `pcase`.  This causes two bugs:

1. *Wrong role*: `(tool . id)` falls through to `nil`, so the role  becomes Lisp `nil`, which JSON-encodes as `{}` instead of a string.

2. *Spurious empty user messages*: `gptel-response-separator` regions  between turns have a `nil` `gptel` property, so they were also  captured as user messages.  Their whitespace-only content trims to `nil`, which JSON-encodes as `{}`, This produces bogus `{"role": "user", "content": [{"text": {}}]}` message.

Both problems cause the Bedrock API to reject the request on the second and subsequent turns whenever a tool had been called with the message "Start of structure or map found where not expected".

### Fix

- Add a `(tool . ,id)` arm to the `pcase` in `gptel--parse-buffer` that reads the serialized `(:name ... :args ...)` plist and result  text from the buffer and reconstructs the proper `toolUse` (assistant) / `toolResult` (user) message pair required by the Bedrock Converse API.

- Wrap the `pcase` in a blank-region guard (`skip-syntax-forward " "`) so that whitespace-only regions are
  skipped entirely, matching the approach already used by the Anthropic backend.

- Add an additional nil-and-blank check on the `nil` (user) arm as an extra guard against any remaining edge cases.

### DIFF
--- a/gptel-bedrock.el
+++ b/gptel-bedrock.el
@@ -409,10 +409,46 @@ MAX-ENTRIES is the maximum number of prompts to include."
                       (/= prev-pt (point-min))
                       (goto-char (previous-single-property-change
                                   (point) 'gptel nil (point-min))))
-            (capture-prompt (pcase (get-char-property (point) 'gptel)
-                              ('response "assistant")
-                              ('nil "user"))
-                            (point) prev-pt)
+            ;; Skip blank regions (e.g. response separators) to avoid
+            ;; capturing empty content that JSON-encodes as {}.
+            (unless (save-excursion (skip-syntax-forward " ") (>= (point) prev-pt))
+              (pcase (get-char-property (point) 'gptel)
+                ('response
+                 (capture-prompt "assistant" (point) prev-pt))
+                (`(tool . ,id)
+                 ;; Reconstruct the toolUse (assistant) and toolResult (user)
+                 ;; message pair from the serialized tool call text in the buffer.
+                 (save-excursion
+                   (condition-case nil
+                       (let* ((tool-call (read (current-buffer)))
+                              (name (plist-get tool-call :name))
+                              (args (plist-get tool-call :args))
+                              (result (string-trim
+                                       (buffer-substring-no-properties
+                                        (point) prev-pt))))
+                         ;; user message: toolResult (pushed first, ends up second)
+                         (push (list :role "user"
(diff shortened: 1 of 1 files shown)


karthink:
@jhoodsmith Thanks for the PR!

> TLDR: The new pattern of including tool results in buffers doesn't work with Bedrock.

Was it changed recently?  In the blame I see only the original implementation.

The changes make sense to me, but I don't have access to Bedrock so I can't test these.  Could you raise a PR with a couple of parsing tests at https://github.com/karthink/gptel-test?  That should give us some confidence re: regressions in the future.

@fommil This looks related to #1388 -- could you test this PR?  Ideally we would look at some dry run output to make sure the payload looks like it should.

fommil:
Initial results look good, I've gotten further than I've ever got with tools enabled!

I'm happy to close #1388 and say this fixes it.

````
> use the calculator

Sure! What expression would you like me to calculate?

> anything


``` tool (calc :expr "2 + 2")
(:name "calc" :args (:expr "2 + 2"))

4
```


Oh boy, oh boy! 2 + 2 = 4! Ha-ha! Just a simple one to show it works. Got anything trickier you'd like me to crunch, pal?

> and another please


``` tool (calc :expr "sqrt(144) + 3^3")
(:name "calc" :args (:expr "sqrt(144) + 3^3"))

39
```


Hot dog! √144 + 3³ = 39! Gosh, math sure is swell! Want me to calculate something else for ya?

>
````

jhoodsmith:
@karthink Thanks for feedback.

Will this do? 

https://github.com/karthink/gptel-test/pull/19

karthink:
Yup, looks good for now.  Merged, thanks for the PR!

PR #1413, opened 2026-05-12
fix for bedrock formatting regression

### LINKED ISSUE #1388: bad request format when calling bedrock

fommil:
maybe closes #1388

not gonna lie, I asked Opus 4.7 to analze this and propose a fix, and I feel dirty for doing that.

### DIFF
--- a/gptel-bedrock.el
+++ b/gptel-bedrock.el
@@ -388,7 +388,7 @@ received."
           (_ (error "Unexpected event-type %S" event-type)))))
     (list :role role :content (vconcat (nreverse contents)))))
 
-(cl-defmethod gptel--parse-buffer ((_backend gptel-bedrock) &optional max-entries)
+(cl-defmethod gptel--parse-buffer ((backend gptel-bedrock) &optional max-entries)
   "Parse current buffer and return a list of prompt objects for Bedrock.
 
 MAX-ENTRIES is the maximum number of prompts to include."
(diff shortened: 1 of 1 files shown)


fommil:
I've been using this for the last few days and I've not encountered any problems, bedrock is working again.

@akssri / @felipeochoa can you please take a look at this and see if it makes sense?

gptel-bedrock.el
```diff
@@ -399,20 +399,41 @@ MAX-ENTRIES is the maximum number of prompts to include."
                 (let* ((content (if include-media
                                  (gptel-bedrock--parse-multipart
                                   (gptel--parse-media-links major-mode beg end))
-                                 `[(:text ,(gptel--trim-prefixes
```
fommil, line 402: this part makes a lot of sense, just defensive coding

gptel-bedrock.el
```diff
@@ -399,20 +399,41 @@ MAX-ENTRIES is the maximum number of prompts to include."
...
           (while (and (> max-entries 0)
                       (/= prev-pt (point-min))
                       (goto-char (previous-single-property-change
                                   (point) 'gptel nil (point-min))))
-            (capture-prompt (pcase (get-char-property (point) 'gptel)
-                              ('response "assistant")
-                              ('nil "user"))
-                            (point) prev-pt)
+            (pcase (get-char-property (point) 'gptel)
+              ('response (capture-prompt "assistant" (point) prev-pt))
```
fommil, line 414: I'm not entirely sure what this is doing, it seems to be adding quite a bit more logic to handle tool responses... it's as if the entire tool response handling was just missing before?

fommil:
Oh, interesting, I still get

```
Start of structure or map found where not expected
```

when I enable `gptel-include-tool-results`.

The problem seems to be with parsing the `tool` sections, because if I manually remove them then everything works fine - even without this patch. So I'm going to back out this PR (and never ever send a PR ever again that I don't understand, that has been generated by an LLM, no matter how tempting it may be).

PR #1414, opened 2026-05-12
gptel-gemini: Add gemini-3.1-flash-lite and deprecate the preview

tspiteri:
The Gemini 3.1 Flash Lite Preview model will be discontinued on May 25, 2026. The General Availability version should be the same but called `gemini-3.1-flash-lite`.


djr7C4:
3.1 flash is available now too though 3.1 pro still seems to be in preview.

karthink:
Merged, thank you!

PR #1422, opened 2026-05-17
gptel: Automatically abort gptel in buffers before they are killed.

djr7C4:
This prevents errors that otherwise occur as the streaming response comes in.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -2865,7 +2865,7 @@ PROCESS and _STATUS are process parameters."
            (info (gptel-fsm-info fsm))
            (http-status (plist-get info :http-status)))
       (when gptel-log-level (gptel-curl--log-response proc-buf info)) ;logging
-      (when (buffer-live-p (marker-buffer (plist-get info :position)))
+      (when (buffer-live-p (plist-get info :buffer))
         (cond
          ;; Curl exited with a non-zero status: connection-level failure
          ((not (zerop exit-status))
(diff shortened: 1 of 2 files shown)


karthink:
@djr7C4 Thanks.  This needs to be fixed more generally, though, by checking in the gptel-request process sentinels if the buffer is still alive.  Doing it this way will fail under many conditions, like if you're not using gptel-mode, if you have more than one ongoing request in the buffer, if the prompt and response buffers are different, etc.

Can you add a `buffer-live-p` check to the process sentinels instead?  Ideally we should also check in the process filters, but I think that can wait.

djr7C4:
I've changed it to handle killed buffers in process sentinels and filters. I
initially only changed it in sentinels like you suggested but there were still
errors from the filters so it was necessary to protect those as well.

After those two changes, I noticed that there was another issue caused by
`(marker-buffer start-marker)` not being consistent with buffer in
`gptel-request`. Based on the docstring which states that

> BUFFER and POSITION are the buffer and position (integer or marker) at which
> the response is inserted.

I changed gptel-request to bind buffer to (marker-buffer start-position). This
resolved the remaining errors.

I tested the following and no errors occur with the new code.

1. Run `M-x gptel` and start two queries at the same time. Once both start
   streaming into the buffer, kill the buffer.
2. Run `M-x gptel` and set the output to insert into `*scratch*`. Once it starts
   streaming, kill `*scratch*`. In this case, the status does get stuck on
   waiting. I haven't looked into that issue but it doesn't seem to cause any
   serious problems.
3. Run `M-x gptel` and set the output to insert into `*scratch*`. Once it starts
   streaming, kill `*gptel*`.

djr7C4:
Also, in test 2 gptel--request-alist still gets cleaned up so the problem
appears to be cosmetic.

djr7C4:
The issue with test 2 is because `(gptel--fsm-transition fsm)` in
`gptel-curl--stream-cleanup` is because that form is wrapped in the
`when`. However, when it is not wrapped (as in an earlier version I tested)
`(gptel--fsm-transition fsm)` causes other errors when the buffer is killed so
addressing this would require further fixes.

PR #1424, opened 2026-05-17
gptel-request: Show more detail on codex errors

djr7C4:
Currently, errors with Codex subscriptions are rather cryptic because :detail is
not handled.


gptel-request.el
```diff
@@ -2678,7 +2678,8 @@ See `gptel-curl--get-response' for its contents.")
                 (string-trim resp))
               http-status http-msg))
        ((and-let* ((error-data
-                    (cond ((plistp response) (or (plist-get response :error)
+                    (cond ((plistp response) (or (plist-get response :detail)
```
karthink: Could you add a one word comment denoting which backend each key is for to the end of each line here?  Something like:

``` emacs-lisp
((and-let* ((error-data
             (cond ((plistp response) (or (plist-get response :detail)    ;openai-oauth
                                          (plist-get response :error)     ;generic
                                          (plist-get response :message)   ;bedrock
                                          (plist-get response :Message))) ;bedrock
                   ((arrayp response)
                    (cl-some (lambda (el) (plist-get el :error)) response)))))
   (list nil http-status http-msg error-data)))
```

This list is growing, so in the future error handling should probably become a `cl-defmethod` defined by each backend for itself.

karthink (changes requested):
Please see suggestion.

djr7C4:
Updated.

karthink:
Merged, thanks for the PR!

PR #1425, opened 2026-05-18
fix: gptel--realize-query prompt can get deleted

FrauH0lle:
This fixes an edge case in `gptel--realize-query` which surfaced mainly due to my personal emacs config using `persp-mode` and custom popup window system: 

In `gptel--realize-query` it is assumed that the prompt buffer stays selected while the function runs and  `(kill-buffer (current-buffer))` points to it. 
Under certain (rare, crazy...) circumstances, `current-buffer` changes and points towards an already killed one and thus, this function errors.

This PR makes the selection of prompt buffer explicit.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -2235,8 +2235,9 @@ be used to rerun or continue the request at a later time."
   "Realize the query payload for FSM from its prompt buffer.
 
 Initiate the request when done."
-  (let ((info (gptel-fsm-info fsm)))
-    (with-current-buffer (plist-get info :data)
+  (let* ((info (gptel-fsm-info fsm))
+         (prompt-buffer (plist-get info :data)))
+    (with-current-buffer prompt-buffer
       (let* ((directive (gptel--parse-directive gptel--system-message 'raw))
              ;; DIRECTIVE contains both the system message and the template prompts
              (gptel--system-message
@@ -2277,8 +2278,9 @@ Initiate the request when done."
         (when (and gptel-use-tools gptel-tools)
           (plist-put info :tools gptel-tools))
         (plist-put info :data
-                   (gptel--request-data gptel-backend full-prompt)))
-      (kill-buffer (current-buffer)))
+                   (gptel--request-data gptel-backend full-prompt))))
+    (when (buffer-live-p prompt-buffer)
+      (kill-buffer prompt-buffer))
     ;; INIT -> WAIT
     (unless (plist-get info :dry-run) (gptel--fsm-transition fsm))
     fsm))


karthink:
Thanks for the PR!  Could you explain the circumstances under which `(current-buffer)` can change out from under you in `gptel--realize-query`?  It's not asynchronous and it doesn't run any hooks, so I don't understand how this can happen.

FrauH0lle:
It was a bit freaky and specific to my config and caused by implementing [this PR](https://github.com/Bad-ptr/persp-mode.el/pull/149) into `persp-mode`:

```elisp
  (set-buffer (window-buffer (selected-window))))
```

The other requirement is the warning from [this issue](https://github.com/karthink/gptel/issues/1426).

What probably happened:
- `gptel--realize-query` does its job and creates the temporary prompt in a dedicated buffer, made current via `with-current-buffer`
- The `:temperature` keyword is detected, removed and `display-warning` is called.
- The `*Warnings*` buffer is handled by my config and reaches the above `persp-mode` code
- Now current buffer is not the the query buffer anymore but the warning buffer
- Close/kill the warning buffer and `(kill-buffer (current-buffer))` now points to an already killed buffer

I guess no one else ran into this bug but it means that there can be setups where this behavior can happen and this PR hardens against these edge cases.

karthink:
>  (set-buffer (window-buffer (selected-window))))

Wow, this code is a giant footgun -- running `set-buffer` or `switch-to-buffer` noninteractively is a sign of trouble in general.

But I understand the issue now, and I would rather the request fail than implement this fix.  Here's why.  Consider this block of code, which is supposed to run in the prompt buffer:

``` emacs-lisp
        (when (and gptel-context gptel-use-context (gptel--model-capable-p 'media))
          (gptel--inject-media gptel-backend full-prompt))
        (unless stream (cl-remf info :stream))
        (plist-put info :backend gptel-backend)
        (plist-put info :model gptel-model)
        (when gptel-include-reasoning   ;Required for next-request-only scope
          (plist-put info :include-reasoning gptel-include-reasoning))
        (when (and gptel-use-tools gptel-tools)
          (plist-put info :tools gptel-tools))
        (plist-put info :data
                   (gptel--request-data gptel-backend full-prompt)))
```

Suppose some clause, say `gptel--inject-media`, changes the current buffer out from under us.  Then the following forms will use the wrong values of `gptel-backend`, `gptel-model` and `gptel-tools`, as well as the wrong values of other gptel parameters used by `gptel--request-data`.  The request will not fail, but may be sent somewhere completely different from where you intended it to, which is much worse.

There is no way to guard against this in the current design, which uses buffer-local variables to store request information.  Even adding a `(with-current-buffer prompt-buffer ...)` around each of these forms will not be enough.

I need to think some more about how to guard against this, but I consider the proposed fix worse than the current state of the code.  I am keeping this PR open for discussion -- let me know if you have any ideas!

PR #1433, opened 2026-05-21
Add keyring support to gptel-oauth and integrate it for OpenAI

djr7C4:


djr7C4:
I've been thinking about this a bit more. It should probably set the `:host` attribute to `"chatgpt.com"` instead of setting `:service` like Codex does. Also, this should probably be done using `auth-sources` instead. Using `:host` will work more nicely with that.

djr7C4:
By default, `auth-source-save-behavior` is kind of annoying which is why people often set it to `nil` instead of `ask`. Maybe we could add a variable `gptel--auth-source-save-behavior` to control that just for gptel.

karthink:
@djr7C4 What is the status of this PR?

djr7C4:
@karthink The keyring version works well and I've been using it in my personal fork since I wrote it. I can add support for `auth-source` if you'd like. I had been holding off until I heard from you.

PR #1434, opened 2026-05-21
gptel-openai-oauth: Ignore max_output_tokens as Codex does not support it

djr7C4:


gptel-openai-oauth.el
```diff
@@ -211,7 +216,7 @@ before constructing the headers."
 
 ;;;###autoload
 (cl-defun gptel-make-openai-oauth
-    (name &key curl-args (stream t) request-params
```
karthink, line 214: I think it's better to still allow a stream argument to `gptel-make-openai-oauth` for uniformity with other gptel-make-* functions, please revert this change.

It's also not clear why this change is part of this PR?

karthink (changes requested):

gptel-openai-oauth.el
djr7C4: Arguably it should be a separate PR. Non-streaming requests result in an error with the codex endpoint though (somewhat like other parameters that are not supported).

djr7C4:
Reverted.

PR #1436, opened 2026-05-27
gptel-bedrock: Support system and tool prompt caching

jhoodsmith:
**The problem:** While `gptel--bedrock-update-tokens` already tracked `cacheReadInputTokens` and `cacheWriteInputTokens` from Bedrock responses, prompt caching was never actually being activated — no `:cachePoint` block was included in requests. This meant the cache fields in responses were always zero.

**This fix:** Two caching scenarios are now supported, controlled by `gptel-cache`:

- **System prompt caching** — when `gptel-cache` is `t` or includes the symbol `system`, a `:cachePoint` block is appended to the system message in the request body.
- **Tool caching** — when `gptel-cache` is `t` or includes the symbol `tool`, a `:cachePoint` block is appended as the final entry in the `:tools` array inside `:toolConfig`.

In both cases the cache point type is `"default"`, as required by the Bedrock Converse API.

See: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html

### DIFF
--- a/gptel-bedrock.el
+++ b/gptel-bedrock.el
@@ -84,11 +84,17 @@ back to the LLM)."
   (let ((base-request-data
          (nconc
           `(:messages [,@prompts] :inferenceConfig (:maxTokens ,(or gptel-max-tokens 500)))
-          (when gptel--system-message `(:system [(:text ,gptel--system-message)]))
+          (when gptel--system-message
+            `(:system [(:text ,gptel--system-message)
+                       ,@(when (or (eq gptel-cache t) (memq 'system gptel-cache))
+                           '((:cachePoint (:type "default"))))]))
           (when gptel-temperature `(:temperature ,gptel-temperature))
           (when (and gptel-use-tools gptel-tools)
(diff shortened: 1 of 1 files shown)


jhoodsmith:
@karthink - I've gone for system prompt caching only for now, but can include tool caching in in this PR if you would like that? On message caching: the bedrock api seems to handle this very differently to anthropic's, and I've had no luck implementing with my limited experiments. Message caching would have to be included in another PR.

Also, unlike `gptel-anthropic.el`, I haven't used the `gptel--model-capable-p` guard, since bedrock models aren't registered with `cache` capability. Happy to look into this, but that would make the PR considerably bigger.

karthink:
@jhoodsmith Sure, feel free to add tool caching to this PR as well.  You can specify system prompt and tool caching separately in `gptel-cache`, that can be supported for Bedrock.

As for `(gptel--model-capable-p 'cache)`, you can ignore this.  I think I should remove this field since caching is provided/controlled by the backend more than the model and this hasn't actually been useful.

PR #1445, opened 2026-06-07
gptel-request: Unify error handling and cover all keywords in arrays

djr7C4:
There is currently a bug with Codex (using gptel-openai-oauth.el) that causes
the reasons for some errors to be omitted. For example, when attempting to use
gpt-5.4-nano with Codex an error occurs because the model is not supported but
the reason is not shown. This is because when the response is an array only the
:error keyword is checked but Codex uses :detail.

Rather then insert :detail in yet another spot and add more duplicate code, I've
added a separate function for extracting errors and replaced the copies of

    (cond ((plistp response)
           (or (plist-get response :error)     ; generic
               (plist-get response :detail)    ; openai-oauth
               (plist-get response :message)   ; bedrock
               (plist-get response :Message))) ; bedrock
          ((arrayp response)
           (cl-some (lambda (el) (plist-get el :error)) response)))

with it.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -1256,6 +1256,23 @@ returned as a list of strings."
                            (gptel--parse-list
                             gptel-backend (cdr directive))))))))
 
+;;;; Error handling
+(defun gptel--parse-response-plist-error (response)
+  "Extract errors from RESPONSE when it is a plist.
+
+Return nil if no error occurred."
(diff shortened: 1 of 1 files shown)


karthink:
Thanks @djr7C4!  I changed some names to be more descriptive and merged it.

Issue #220, opened 2024-02-15
Support Google Vertex


cashpw:
Google Vertex provides easy access to [many models](https://console.cloud.google.com/vertex-ai/model-garden). I think it'd be okay to reference [`llm-vertex.el`](https://github.com/ahyatt/llm/blob/4add6102f83176beb9655083dd3844ab91672238/llm-vertex.el) as both projects have the same GPL 3.0 license.

karthink:
@cashpw Thank you.  This (and your Google Gemini/Vertex related PRs and issues) are on the docket, I'll get to them soon.

algal:
I may have time to implement this, but first I'm trying to understand what would be involved.

If I understand right, llm-vertex.el does not actually have a working emacs lisp implementation which can authenticate and interact with vertex. Instead, they effectively require you to install the gcloud binary and configure it. That does not seem idle.

Do we know if anyone has a pure emacs lisp client that can authenticate with vertex ai?

karthink:
@algal I haven't had time to look into this.  It looks like Vertex AI [supports a REST interface](https://cloud.google.com/vertex-ai/docs/reference/rest) though.  The `generateContent` endpoint is the one that gptel would support.

algal:
Here's a quick update on this:

I just had a look at llm-vertex.el, and was able to use it to query VertexAI with authentication successfully. This required setting up the gcloud binary and authenticating with it. The lisp code for making a request looks like this, fyi:

```lisp
(customize-set-value 'llm-vertex-gcloud-binary "/Users/alexis/miniforge3/bin/gcloud")
(customize-set-value 'llm-vertex-gcloud-region "us-east5")
(llm-chat (make-llm-vertex
           :project "PROJECT_ID_HERE"
           :chat-model "gemini-pro"
           :default-chat-temperature 0.5
           :default-chat-max-tokens 1000
           :default-chat-non-standard-params (list))
          (llm-make-chat-prompt
           "Hello, are you there?"))
```

I also had a look at the extensive (and imho unnecessarily confusing) list of Vertex AI endpoints. 

As far as I can tell:
- the `generateContent` endpoint is the one which corresponds to the "metric" described in the Gcloud API as "aiplatform.googleapis.com/generate_content_requests_per_minute_per_project_per_base_model". 
- that endpoint or metric offers various Gemini models and some Llama models, but it does *not* offer any anthropic models.
[...]

PR #1449, opened 2026-06-10
Add the Vertex AI API support for Anthropic backend

dodjiseketeli:
Google Cloud Platform's Vertex AI provides access to Anthropic's Claude models through a wrapper API that differs slightly from Anthropic's direct API.  The key differences are:

1. The model identifier is embedded in the endpoint URL rather than in the request body
2. Vertex AI rejects requests that include a "model" field in the JSON payload with an "Extra inputs are not permitted" error
3. Authentication uses Google Cloud OAuth tokens instead of Anthropic API keys

This patch adds a `uses-vertex-ai` flag to the `gptel-anthropic` backend structure that, when enabled, prevents the "model" field from being added to API requests. This allows gptel to work seamlessly with Claude models accessed through Vertex AI.

An example configuration for the Vertex AI API in a user's Emacs configuration would be:

  (setq gptel-backend
        (gptel-make-anthropic "VertexAI"
          :uses-vertex-ai t
          :header (lambda ()
                    (let ((token (string-trim
                                  (shell-command-to-string
                                   "gcloud auth print-access-token"))))
                      `(("Authorization" . ,(concat "Bearer " token)))))
          :stream t
          :host "<insert-region-here>-aiplatform.googleapis.com"
          :request-params '(:anthropic_version "vertex-2023-10-16")
          :endpoint (format "/v1/projects/%s/locations/%s/publishers/anthropic/models/%s:streamRawPredict"
                            "YOUR-PROJECT-ID"
                            "<insert-region-here>"
                            gptel-model)))

The `uses-vertex-ai` parameter defaults to nil, maintaining backward compatibility with existing Anthropic API configurations.

        * gptel-anthropic.el (gptel-anthropic): Add `uses-vertex-ai` slot
        to the structure.
        (gptel--request-data): Conditionally omit the `:model` field from the
        request plist when `uses-vertex-ai` is non-nil.
        (gptel-make-anthropic): Add `uses-vertex-ai` keyword parameter and
        pass it to the backend constructor.

### DIFF
--- a/gptel-anthropic.el
+++ b/gptel-anthropic.el
@@ -38,7 +38,8 @@
 ;;; Anthropic (Messages API)
 (cl-defstruct (gptel-anthropic (:constructor gptel--make-anthropic)
                                (:copier nil)
-                               (:include gptel-backend)))
+                               (:include gptel-backend))
+  (uses-vertex-ai nil))
 
 (defun gptel--anthropic-update-tokens (usage info)
   "Update token usage information from USAGE.
@@ -223,12 +224,16 @@ Mutate state INFO with response metadata."
 (cl-defmethod gptel--request-data ((backend gptel-anthropic) prompts)
   "JSON encode PROMPTS for sending to ChatGPT."
   (let ((prompts-plist
-         `( :model ,(gptel--model-name gptel-model)
-            :stream ,(or gptel-stream :json-false)
+         `( :stream ,(or gptel-stream :json-false)
             :max_tokens ,(or gptel-max-tokens 4096)
             :messages [,@prompts]))
         (cachep (and (or (eq gptel-cache t) (memq 'system gptel-cache))
                      (gptel--model-capable-p 'cache))))
+    ;; The Vertex AI API specifies the model in the endpoint URL, not
+    ;; in the request body.  Thus, let's not specify the model in the
+    ;; requet body if we are using the Vertex AI API.
+    (when (not (gptel-anthropic-uses-vertex-ai backend))
+      (plist-put prompts-plist :model (gptel--model-name gptel-model)))
     (when gptel-system-prompt
       ;; gptel-system-prompt is a string or a list of strings
       (plist-put
@@ -691,7 +696,8 @@ URL `https://docs.anthropic.com/en/docs/about-claude/models#model-comparison-tab
           (models gptel--anthropic-models)
           (host "api.anthropic.com")
           (protocol "https")
-          (endpoint "/v1/messages"))
+          (endpoint "/v1/messages")
+          (uses-vertex-ai nil))
   "Register an Anthropic API-compatible backend for gptel with NAME.
 
 Keyword arguments:
@@ -728,6 +734,11 @@ PROTOCOL (optional) specifies the protocol, https by default.
 ENDPOINT (optional) is the API endpoint for completions, defaults to
 \"/v1/messages\".
 
+USES-VERTEX-AI (optional) is a boolean to toggle support for using
+anthropic models through the Google Cloud Platform.  That platform uses
+the Vertex AI API which specifies the model in the endpoint URL rather
+than in the body of the request.  This is set to nil by default.
+
 HEADER (optional) is for additional headers to send with each
 request.  It should be an alist or a function that retuns an
 alist, like:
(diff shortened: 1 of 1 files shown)


dodjiseketeli:
hello,  I just wanted to add that I have been testing this patch and it does work both when using the GCP/Vertex AI API and when using the direct anthropic API.  Thank you for such a great project!
Cheers,

karthink:
@dodjiseketeli Thanks for the PR.  I'm quite confused about what's happening here, though:

You are accessing Anthropic's models via Vertex and want to add support to gptel for doing that.  Is that correct?

If it is, I think it will be better to add a gptel-vertex feature to gptel adding explicit support for Vertex via gptel.  You can then use `gptel-make-vertex` to access the Vertex API.

Since you're familiar with the Vertex API, do you know which API format it uses?  Is it OpenAI-Completions, Anthropic or Gemini, or is it more than one of these, or is the API used model-dependent?

dodjiseketeli:
> @dodjiseketeli Thanks for the PR.

You are welcome.

[...]

> You are accessing Anthropic's models via Vertex and want to add support to gptel for doing that. Is that correct?

That is correct.

> If it is, I think it will be better to add a gptel-vertex feature to gptel adding explicit support for Vertex via gptel.

You mean, something like what this issue suggests? https://github.com/karthink/gptel/issues/220.

I have seen that issue after I solved my issue by trial and error.

>You can then use `gptel-make-vertex` to access the Vertex API.

Right. 

In the grand scheme of things, it's arguable that having a vertex backend for gptel would be the way to go.  However, we need to keep in mind that we would then need specific support for each kind of model vendor supported by Vertex AI; that is anthropic support, openai support, gemini support, etc.

Alternatively, we could also think that each existing gptel backend could have a "parameter" to be configured to run either natively (as it is now) or, on the Vertex AI platform.

The later is the approach I have taken for the gptel-anthropic backend, just because it was more direct to do that my own personal need, without having to write an entire vertex AI backend with anthropic support.

I believe the former approach is the one taken by https://github.com/ashlineldridge/gptel-vertex, where it has support for anthropic and gemini.
 
> Since you're familiar with the Vertex API, do you know which API format it uses? Is it OpenAI-Completions, Anthropic or > Gemini, or is it more than one of these, or is the API used model-dependent?

From my limited exposure of this thing, I would say that Vertex API generic format is specific to Vertex, especially for things like authentication and the selection of the model to run.  But then, it uses a model-dependent format that re-uses the format of the model that is selected.

You can see an example of the format used for anthropic models at https://github.com/GoogleCloudPlatform/vertex-ai-samples/blob/main/notebooks/official/generative_ai/anthropic_claude_3_intro.ipynb.

Maybe I should play with https://github.com/ashlineldridge/gptel-vertex and report back.  This looks like the approach you would prefer, right?

karthink:
> Maybe I should play with https://github.com/ashlineldridge/gptel-vertex and report back.  This looks like the approach you would prefer, right?

Yes, this would be better.  However, if this is true:

> But then, it uses a model-dependent format that re-uses the format of the model that is selected.

then it might take a bit more work than expected.  Does https://github.com/ashlineldridge/gptel-vertex handle the model-dependent API formats?

dodjiseketeli:
karthink ***@***.***> a écrit:

>> Maybe I should play with https://github.com/ashlineldridge/gptel-vertex and report back.  This looks like the approach you would prefer, right?
>
> Yes, this would be better.  However, if this is true:

OK, I have started to use gptel-vertex with the Claude models and it's
been working for me so far.  I haven't tried the Gemini models yet, so I
cannot report on that.

>> But then, it uses a model-dependent format that re-uses the format of the model that is selected.
>
> then it might take a bit more work than expected.

Indeed.

> Does https://github.com/ashlineldridge/gptel-vertex handle the
> model-dependent API formats?

Yes, it does so for two model families (aka publishers) hosted on the
Google platform so far: the Anthropic models and the Gemini ones.

For instance, you can see in the gptel-vertex.el file that the
implementation of the gptel--request-data method invokes either
gptel-vertex--request-data-anthropic or
gptel-vertex--request-data-gemini depending on if the user wants to use
models from the Anthropic publisher or from the Gemini publisher.

That's the model-dependent API format handling I was alluding to.  I
should have said more precisely, a publisher-dependent API format.

We would need to add specific support for the other publishers whose
models are hosted on the Google platform (and thus accessible via the
Vertex API).

-- 
		Dodji

PR #1451, opened 2026-06-11
Add support to copilot for claude-opus-4.8 and claude-fable-5

skittishdev:


karthink:
Rebased and merged, thanks @skittishdev! I'm not sure about the costs of the
models when used via GitHub Copilot, since the billing structure has been
changing.  Model costs across the whole `gptel--gh-models` list may need to be
adjusted at some point.

PR #1454, opened 2026-06-16
Grok 4.3

endgame:
xAI has narrowed its range of text-output models to `grok-4.3` and `grok-build-0.1`. All other model names are just aliases to these.


karthink:
@endgame thanks.  Could you undo all the changes except for the ones to the xai backend?

endgame:
Done.

PR #1455, opened 2026-06-16
Sanitize raw bytes in tool results for JSON

alx242:
A tool result may contain non-UTF-8 bytes, for example from reading a binary file via a shell command.  Emacs represents these as raw-byte codepoints in the #x3FFF80..#x3FFFFF range, and passing such a string to `json-serialize' signals (wrong-type-argument json-value-p ...), which tears down the request and leaves it unrecoverable.

Add `gptel--sanitize-raw-bytes' to replace raw bytes with "?" and apply it to the tool result string.  Legitimate Unicode is preserved.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -980,6 +980,25 @@ and \"apikey\" as USER."
     (string s)
     (otherwise (prin1-to-string s))))
 
+(defun gptel--sanitize-raw-bytes (s)
+  "Replace raw bytes in string S with \"?\" so it is JSON-serializable.
+
+Tool results may contain non-UTF-8 bytes (e.g. from reading a binary
+file), which `json-serialize' rejects.  Returns S unchanged if clean;
+otherwise mutates a single fresh copy in place (O(N), one allocation)."
+  (if (not (memq 'eight-bit (find-charset-string s)))
+      s
+    ;; Allocate exactly once, then mutate in place with `aset'.
+    (let* ((s (if (multibyte-string-p s)
+                  (copy-sequence s)
+                (string-to-multibyte s)))
+           (end (length s))
+           (pos (unencodable-char-position 0 end 'utf-8 nil s)))
+      (while pos
+        (aset s pos ??)
+        (setq pos (unencodable-char-position (1+ pos) end 'utf-8 nil s)))
+      s)))
+
 (defsubst gptel--intern (s)
   "Intern S, if possible."
   (cl-etypecase s
@@ -1925,7 +1944,7 @@ injects the results into the prompt data and transitions the FSM."
          ;; MAYBE(tool-hooks): Use plist-member for valid nil :result?
          (remaining (cl-loop for call in (plist-get info :tool-use)
                              count (not (plist-get call :result)))))
-    (let ((result (gptel--to-string result)))
+    (let ((result (gptel--sanitize-raw-bytes (gptel--to-string result))))
       ;; FIXME(tool-hooks): If a hook has changed the tool that was called
       ;; tool-spec needs to be updated.
       (push (list tool-spec (plist-get tool-call :args) result)


karthink:
I like the idea, and have wanted to implement this feature for a while.  But the implementation isn't great, since `replace-regexp-in-string` returns a new string.  If a tool call unexpectedly produced a 5 MB byte stream (perhaps it read a PDF as literal bytes), then replace-regexp-in-string is going to be (i) pretty slow, blocking Emacs, and (ii) allocate another 5 MB string.

This is an issue even if there are no raw-bytes in the string.  `gptel--sanitize-raw-bytes` is always called and  `replace-regexp-in-string` _always_ returns a new string.  There should be a better way to do this, perhaps using the `*-unencodable-*` functions, like `unencodable-char-position` called on the tool result string.

It also gets a lot easier if we insert the tool result into a buffer first, but for performance that's not a great idea either, at least until we can use the `with-work-buffer` macro from Emacs 31 freely in gptel.  (It's provided by compat, a gptel dependency, but I'm not confident about using it yet -- most users won't have compat-31 available.)

alx242:
@karthink See what you mean. I reworked it a bit.

The new solution uses find-charset-string which should be a cheap C-level scan. If it is missing the eight-bit then string should be "safe" to read as is. If a replace is necessary then a single allocation is done with in-place aset in the loop. 

I did not implement a non-blocking handling since I think it might require unwanted complexity until it is solvable at lower levels.

yantar92:
This implementation is better than #1495, although I am not sure if `aset' is a good idea - AFAIR, aset on strings is to be deprecated, if not already. In my patch, I intercepted the json encoding directly though. I think it is more reliable.

djr7C4:
This is also an issue in ordinary gptel buffers (not just as tool calls) if octal escape sequences find their way into the text.

alx242:
> This implementation is better than #1495, although I am not sure if `aset' is a good idea - AFAIR, aset on strings is to be deprecated, if not already. In my patch, I intercepted the json encoding directly though. I think it is more reliable.

Looking in the manual the `aset` is mentioned as the primary method of modifying strings:

https://www.gnu.org/software/emacs/manual/html_node/elisp/Modifying-Strings.html
https://www.gnu.org/software/emacs/manual/html_node/elisp/Mutability.html

I think it should be safe until something actually replaces it.

renatofdds:
Hello! 

I'm also facing this problem in my own AI utilities (not gptel), and stumbled upon this issue. In my case it was for serialization of emacs buffer text to be sent to LLMs (which is not solved by the solution here). 

In my case it was the char `?\x3FFF7F` that was giving me problems, which has a different charset:

```elisp
(find-charset-string (char-to-string ?\x3FFF7F))
;; => (emacs)
```

I ended up implementing the following solution and thought of sharing as a suggestion. It deals with multibyte correctly, avoids `find-charset-string` for ordinary multibyte, and uses `U+FFFD` instead of `?` which i think is more AI-friendly for models to attribute meaning.

```elisp
(defun l/ai/to-json-serializable-string (s)
	"Replace raw bytes in S with the Unicode replacement character.
Return S unchanged if it is already UTF-8-encodable; otherwise return a
sanitized copy."
	(let ((pos (if (multibyte-string-p s)
								 (unencodable-char-position 0 (length s) 'utf-8 nil s)
							 (and (memq 'eight-bit (find-charset-string s)) 0))))
		(if (not pos)
				s
			(let* ((s (if (multibyte-string-p s)
										(copy-sequence s)
									(string-to-multibyte s)))
						 (end (length s))
						 (pos (unencodable-char-position pos end 'utf-8 nil s)))
				(while pos
					(aset s pos #xfffd)
					(setq pos (unencodable-char-position (1+ pos) end 'utf-8 nil s)))
				s))))
```

alx242:
>

> Hello!
> 
> I'm also facing this problem in my own AI utilities (not gptel), and stumbled upon this issue. In my case it was for serialization of emacs buffer text to be sent to LLMs (which is not solved by the solution here).
> 
> In my case it was the char `?\x3FFF7F` that was giving me problems, which has a different charset:
> 
> ```emacs-lisp
> (find-charset-string (char-to-string ?\x3FFF7F))
> ;; => (emacs)
> ```

Good point! I recently did a refresh of the code into something that is a bit more stable by using a temp-buffer. I will evaluate to incorporate your point as well :)

renatofdds:
> Good point! I recently did a refresh of the code into something that is a bit more stable by using a temp-buffer. I will evaluate to incorporate your point as well :)

Great! I ended up refactoring it afterwards to decode utf-8 content as much as i can:

```elisp
(defun l/ai/to-json-serializable-string (s)
	"Return S unchanged if it is already UTF-8-encodable; otherwise return a
sanitized copy, decoding UTF-8 byte sequences and replacing raw bytes with
Unicode replacement characters."
	(when (or (memq 'eight-bit (find-charset-string s))
						(unencodable-char-position 0 nil 'utf-8 nil s))
		(setq s (decode-coding-string s 'utf-8))
		(let ((start 0) (end (length s)))
			(while-let ((pos (unencodable-char-position start end 'utf-8 nil s)))
				(aset s pos #xfffd)
				(setq start (1+ pos)))))
	s)
```

PR #1459, opened 2026-06-24
gptel-request: Allow including only tool call without results

yantar92:
* gptel-request.el (gptel-include-tool-results): (gptel-tool): Add new value as include parameter: `call' to include the call, but not the results.
(gptel-make-tool):
* gptel-transient.el (gptel--infix-include-tool-results): Document all possible values of :include.
* gptel-transient.el (gptel--infix-include-tool-results): Allow selecting the new value from transient menu.
* gptel.el (gptel--display-tool-results): Support the new option, leaving a hint to the LLM that the result is omitted to avoid confusion after the next prompt.
* NEWS (New features and UI changes): Document the new feature.

Some kinds of results, like grep output or web search, are pretty wordy and may not exactly be meaningful in the logs (read - not human-readable).  Yet, it is important to have a trace that such calls have been made, to understand the reasoning process and catch potential issues, like model *not* calling web search and instead relying on internal memory.  The new option allows logging the tool call without inserting the call results.

Assisted-by: DeepSeek + macher.el + custom tools

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -64,6 +64,13 @@
 
 ** New features and UI changes
 
+- The gptel tool's =:include= slot and the global option
+  ~gptel-include-tool-results~ now also accept the symbol =call=: when
+  set, only the call parameters are inserted into the buffer; the result
+  is omitted.  This is useful when the tool results are large or not
+  very human-readable, yet it is important for you or the LLM to track
+  what calls have been made.
+
 - You can now set gptel's system prompt directly via
   ~gptel-system-prompt~.  Previously this required setting an internal
   gptel variable (~gptel--system-message~, now obsolete), or defining a
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -1468,15 +1468,21 @@ confirmation only when the corresponding tool spec has a non-nil
 (defcustom gptel-include-tool-results 'auto
   "Whether tool call results should be included in the buffer.
 
-If set to t or nil, results of tool calls are always or never
-included in the LLM response, respectively.
+If set to t, tool calls and results are always included in the LLM
+response.
 
-If set to the symbol auto (the default), a tool call result is
-included only when the corresponding tool spec has a non-nil
-:include slot.  See `gptel-make-tool'."
+If set to nil, tool calls are never shown in the buffer.
+
+If set to the symbol `call', only the tool call parameters are included
+in the buffer; results are omitted.
+
+If set to the symbol auto (the default), this is left up to the tool: a
+tool call is included only when the corresponding tool spec has a
+non-nil :include slot.  See `gptel-make-tool'."
   :type '(choice
           (const :tag "Tool decides" auto)
           (const :tag "Always" t)
+          (const :tag "Call only, no results" call)
           (const :tag "Never" nil)))
 
 (defcustom gptel-tools nil
@@ -1507,7 +1513,10 @@ feed the LLM the results.  You can add tools via
   (async nil :type boolean :documentation "Whether the function runs asynchronously")
   (category nil :type string :documentation "Use to group tools by purpose")
   (confirm nil :type boolean :documentation "Seek confirmation before running tool?")
-  (include t :type boolean :documentation "Include tool results in buffer?"))
+  (include t :type (choice (const :tag "Include call and result" t)
+                           (const :tag "Include call only, no result" call)
+                           (const :tag "Exclude" nil))
+           :documentation "Include tool call in buffer?"))
 
 (defun gptel--preprocess-tool-args (spec)
   "Convert symbol :type values in tool SPEC to strings destructively."
@@ -1636,7 +1645,11 @@ user should be prompted.
 INCLUDE: Whether the tool results should be included as part of
 the LLM output.  This is useful for logging and as context for
 subsequent requests in the same buffer.  This is primarily useful
-in chat buffers.
+in chat buffers.  Possible values:
+
+- t (default): Include both the call parameters and the result.
+- symbol `call': Include only the call parameters, not the result.
+- nil: Exclude the tool call from the buffer entirely.
 
 Here is an example definition:
 
--- a/gptel-transient.el
+++ b/gptel-transient.el
@@ -1589,14 +1589,14 @@ This sets the variable `gptel-confirm-tool-calls', which see."
 (transient-define-infix gptel--infix-include-tool-results ()
   "Whether tool call results should be included in the response.
 
-This is a three-way toggle between these behaviors:
+This sets the variable `gptel-include-tool-results', which see.
+The possible values are:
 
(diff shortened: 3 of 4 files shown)


karthink:
> Some kinds of results, like grep output or web search, are pretty wordy and may not exactly be meaningful in the logs (read - not human-readable). Yet, it is important to have a trace that such calls have been made, to understand the reasoning process and catch potential issues, like model not calling web search and instead relying on internal memory.

Is it important for debugging/introspecting the state of the conversation, or to inform the model that it called certain tools in past turns, or both?

If it's important for debugging/introspection, I think it's better to set include-tool-results to nil and implement some other debug-level mechanism for logging that a tool call was made.

If it's to inform the model, have you found it useful to have a tool call in the conversation history with dummy results?

yantar92:
karthink ***@***.***> writes:

> Is it important for debugging/introspecting the state of the conversation, or to inform the model that it called certain tools in past turns, or both?

It is important to my own introspection when I read the model reasoning
chain and want to understand if the response is correctly grounded on
data or just uses internal model knowledge. Not just at the time of the
conversation but also much later.

It is also important to avoid confusing the model when I want to
restart the conversation other time, in different Emacs session. I
learned in practice that simply removing tool calls from the reasoning
chain can confuse the model as it can draw wrong conclusions from the
reasoning and completely break the logic of the conversation.

> If it's important for debugging/introspection, I think it's better to set include-tool-results to nil and implement some other debug-level mechanism for logging that a tool call was made.

The purpose is not debugging. But following the conversation.

> If it's to inform the model, have you found it useful to have a tool call in the conversation history with dummy results?

Yes. For example, I want to know the fact that model, for example,
searched certain local files, but have 0 interest to have their contents being
duplicated stored in the conversation history and my notes.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

karthink:
There is some ambiguity in the behavior of `call`:

What should be included when:
|                             | `gptel-include-tool-results` | t    | auto      | call      | nil     |
|-----------------------------|------------------------------|------|-----------|-----------|---------|
| `(gptel-tool-include tool)` |                              |      |           |           |         |
|-----------------------------|------------------------------|------|-----------|-----------|---------|
| t                           |                              | both | both      | call-only | neither |
| call                        |                              | ?    | call-only | call-only | neither |
| nil                         |                              | both | neither   | ?         | neither |

where
- both: include calls and results
- call-only: include only calls, no results
- neither: don't include anything

What should go in the "?" columns, and do I understand the other possibilities correctly?

yantar92:
karthink ***@***.***> writes:

> What should be included when:
> |                             | `gptel-include-tool-results` | t    | auto      | call      | nil     |
> | `(gptel-tool-include tool)` |                              |      |           |           |         |
> |-----------------------------|------------------------------|------|-----------|-----------|---------|
> | t                           |                              | both | both      | call-only | neither |
> | call                        |                              | ?    | call-only | call-only | neither |
> | nil                         |                              | both | neither   | ?         | neither |
>
> What should go in the "?" columns, and do I understand the other possibilities correctly?

AFAIU, gptel-include-tool-results takes priority over per-tool settings,
unless it is set to 'auto. Following this logic, if
gptel-include-tool-results is t, we always include "both". If 'call,
always include "call-only".

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

karthink:
> AFAIU, gptel-include-tool-results takes priority over per-tool settings, unless it is set to 'auto.

Yes, I guess I never formally specified this priority until now, as there was only one value where it mattered.

> Following this logic, if gptel-include-tool-results is t, we always include "both". If 'call, always include "call-only".

Yeah, this makes sense.  I'll verify if the PR does this correctly and merge it.

karthink:
Modified slightly to make the `gptel-include-tool-results > (gptel-tool-include ...)` precedence explicit and merged.  Thanks for the PR Ihor.

It seems like this is worse for staying within the KV cache in extended conversations, depending on how many tool calls are made.

I'm interested in seeing a concrete example of how you use the `call` option.  It seems to me that it's almost always better to have prior call results in subsequent conversation turns, even if they are wordy.

Another option could be to truncate the tool results at 1000 chars or something, with a notice that the rest of the tool result is not available.

yantar92:
karthink ***@***.***> writes:

> It seems like this is worse for staying within the KV cache in extended conversations, depending on how many tool calls are made.

Yes, KV cache may suffer. Although not too much in my (anecdotal) experience.

> I'm interested in seeing a concrete example of how you use the `call` option.  It seems to me that it's almost always better to have prior call results in subsequent conversation turns, even if they are wordy.

See https://gist.github.com/yantar92/cc7141d7895eec51b6131f869d3c524a
and then try the "raw" file.

> Another option could be to truncate the tool results at 1000 chars or something, with a notice that the rest of the tool result is not available.

I am not sure about the logic here. KV cache will still suffer, and the
1000 chars may or may not contain the useful information for subsequent
conversation turns. IMHO, it should be either including everything or
nothing, so that it is clear later that a tool call need to be repeated
if more details are needed.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

karthink:
>> It seems like this is worse for staying within the KV cache in
>> extended conversations, depending on how many tool calls are made.
>
> Yes, KV cache may suffer. Although not too much in my (anecdotal)
> experience.

It shouldn't be too bad -- it's one cache miss per gptel-send
invocation, not a cache miss after each conversation turn (like a tool
call).  If the API has many KV-cache checkpoints the cost should be even
lower.

>> I'm interested in seeing a concrete example of how you use the `call`
>> option.  It seems to me that it's almost always better to have prior
>> call results in subsequent conversation turns, even if they are
>> wordy.
>
> See https://gist.github.com/yantar92/cc7141d7895eec51b6131f869d3c524a
> and then try the "raw" file.

I hadn't considered that your tool calls might contain serialized hash
tables.  That's pretty wild.

>> Another option could be to truncate the tool results at 1000 chars or
>> something, with a notice that the rest of the tool result is not
>> available.
>
> I am not sure about the logic here. KV cache will still suffer, and the
> 1000 chars may or may not contain the useful information for subsequent
> conversation turns. IMHO, it should be either including everything or
> nothing, so that it is clear later that a tool call need to be repeated
> if more details are needed.

Yeah, including no result is probably a cleaner signal.

PR #1460, opened 2026-06-27
Fix FSM hang when model calls unknown tool name

randazzo-ignacio:
## DISCLAIMER

I am working on a self modifying AI tool based on gptel, I found it sometimes hallucinated non-existant tool calls and this hangs the FSM of gptel-request, it found and fixed the issue, but the code and following explanation is AI generated. I thought it appropiate to explain this so nobody wastes any time if the fix isn't up to standard.

## Problem

When the model calls a tool name that doesn't match any registered
`gptel-tool`, the FSM hangs indefinitely in the TOOL state. The status
line shows "Calling tool (whatever)" and never recovers.

This is particularly common with local models (Ollama) that hallucinate
slightly-wrong tool names (e.g., `read_directory` instead of
`list_directory`).

## Root cause

In `gptel--handle-tool-use` (gptel-request.el), when `tool-spec` is nil
and the tool name is not the ersatz JSON tool, the code only calls
`(message "Unknown tool called by model: %s" name)` and does nothing
else. `process-tool-result` is never called, so the pending-tool counter
never reaches zero and the FSM never transitions out of the TOOL state.

## Fix

Call `process-tool-result` with an error message listing the available
tools. This unblocks the FSM and feeds the error back to the model so
it can retry with the correct tool name.

Additionally, `gptel-ollama.el` had a related issue: both the streaming
and non-streaming parsers used `cl-loop ... finally` to set
`info :tool-use`, which overwrites instead of appends. If Ollama sends
tool calls across multiple JSON chunks, earlier calls are silently
lost. Changed to accumulate with `append`.

## Testing

Tested with Ollama backend using models that frequently hallucinate
tool names. Before the fix, the hang was reproducible within a few
turns. After the fix, the model receives the error and self-corrects.

### DIFF
--- a/gptel-ollama.el
+++ b/gptel-ollama.el
@@ -78,7 +78,9 @@ back to the LLM)."
                              (plist-get call-spec :arguments))
                (plist-put call-spec :arguments nil)
                collect call-spec into tool-use
-               finally (plist-put info :tool-use tool-use)))
+               finally
+               (plist-put info :tool-use
+                          (append (plist-get info :tool-use) tool-use))))))))
             (if (and reasoning (not (eq reasoning :null)))
                 (plist-put info :reasoning
                            (concat (plist-get info :reasoning) reasoning))
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -1948,7 +1948,11 @@ injects the results into the prompt data and transitions the FSM."
                      (funcall (plist-get info :callback)
                               (gptel--json-encode (plist-get tool-call :args))
                               info)
-                   (message "Unknown tool called by model: %s" name))
+		   (message "Unknown tool called by model: %s" name)
+                   (funcall process-tool-result
+                            (format "Error: Tool '%s' is not available. Available tools: %s"
+                                    name (mapconcat #'gptel-tool-name
+                                                    (plist-get info :tools) ", "))))
                (let ((confirm))         ;Check if tool requires confirmation
                  (cond      ;:confirm in tool-call (from hooks) takes precedence
                   ((and-let* ((call-confirm (plist-member tool-call :confirm)))


karthink:
@randazzo-ignacio the fix is mostly fine, but please split this into two commits, one with the FSM hang issue and one with the gptel-ollama fix.

randazzo-ignacio:
> @randazzo-ignacio the fix is mostly fine, but please split this into two commits, one with the FSM hang issue and one with the gptel-ollama fix.

Done.

gptel-ollama.el
```diff
@@ -67,18 +67,20 @@ back to the LLM)."
...
                `(:role "assistant" :content :null :tool_calls ,(vconcat tool-calls)))
-              (cl-loop
-               for tool-call across tool-calls  ;replace ":arguments" with ":args"
-               for call-spec = (copy-sequence (plist-get tool-call :function))
-               do (plist-put call-spec :args
-                             (plist-get call-spec :arguments))
-               (plist-put call-spec :arguments nil)
-               collect call-spec into tool-use
-               finally (plist-put info :tool-use tool-use)))
+              (let ((new-calls
```
karthink: Replace the let block with

```suggestion
(cl-loop
 for tool-call across tool-calls        ;replace ":arguments" with ":args"
 for call-spec = (copy-sequence (plist-get tool-call :function))
 do (plist-put call-spec :args
               (plist-get call-spec :arguments))
 (plist-put call-spec :arguments nil)
 collect call-spec into tool-use
 finally
 (plist-put info :tool-use
            (append (plist-get info :tool-use) tool-use))))
```

gptel-ollama.el
```diff
@@ -108,14 +110,16 @@ Store response metadata in state INFO."
...
       ;; Then capture the tool call data for running the tool
-      (cl-loop
-       for tool-call across tool-calls  ;replace ":arguments" with ":args"
-       for call-spec = (copy-sequence (plist-get tool-call :function))
-       do (plist-put call-spec :args
-                     (plist-get call-spec :arguments))
-       (plist-put call-spec :arguments nil)
-       collect call-spec into tool-use
-       finally (plist-put info :tool-use tool-use)))
+      (let ((new-calls
```
karthink: Why is this required for non-streaming calls?

gptel-ollama.el
randazzo-ignacio: The new-calls var is needed for the last append, as the issue is not the same as the FSM hanging, but the ollama issue is when multiple tool calls are sent together. Although I will concede that this might be more to do with the way I am using the library, than an issue with the library itself, as I am using gptel's tool hooks to display some data between tool calls and this might be introducing the bug. If you don't consider this necessary the ollama commit could be dropped.

gptel-ollama.el
randazzo-ignacio: Same reasoning as on line 74 above

gptel-ollama.el
karthink: My proposed replacement is equivalent to using new-calls in the let block.

gptel-ollama.el
karthink: I don't follow, in the non-streaming branch this is only called once per response.

randazzo-ignacio: My bad, I am only using the streaming calls and assumed it was needed on both, we can remove this change

gptel-ollama.el
randazzo-ignacio: I missed the append on line 83, you are right!

karthink:
Merged with some changes, thanks for the PR!

PR #1462, opened 2026-06-29
gptel-request: Send request body via stdin to avoid temp files

bohonghuang:
Previously, when using curl, gptel handled large payloads by first writing them to a temporary `.json` file and then passing that filename to curl on the command line, which, IMO, has several issues:
1. Resource cleanup is not robust enough; I have noticed leftover temporary files when requests fail.
2. It can cause unnecessary disk writes, especially on Windows where the temporary directory is not memory-backed, which may risk sensitive data being exposed or read by other applications.
3. There also seem to be encoding issues on Windows.

Therefore, this PR changes the implementation to always send request data via stdin, regardless of payload size, to eradicate these problems.

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -140,6 +140,13 @@ It is set to (#x8000 - #x1000 - 2) to account for other (non-data) Curl
 command line arguments."
   :type 'natnum)
 
+(make-obsolete-variable
+ 'gptel-curl-file-size-threshold
+ "gptel now sends request contents via stdin, no longer subject to command-line
+argument limits. `gptel-curl-file-size-threshold' and the related temporary-file
+mechanism are no longer needed."
+ "0.9.9.6")
+
 (define-obsolete-variable-alias 'gptel-prompt-filter-hook
   'gptel-prompt-transform-functions "0.9.9")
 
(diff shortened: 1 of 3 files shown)


karthink:
Thanks for the PR.  This is a very good idea.

But by moving the request body to stdin, you've moved the headers into the command line.  The headers were moved to stdin for security reasons, see 85126d3 for the explanation.  So this is not acceptable.  If you can find a way to send both the headers and data via stdin, it will be welcome.

bohonghuang:
@karthink Done. Now all request parameters (except customizable command args) should be passed to `curl` via stdin together with the request body.

bohonghuang:
`curl --config -` seems flexible enough for this kind of use. If the data comes right after `data-binary = @-`, curl allows to pass request parameters before it, and the request body after it.

karthink:
> `curl --config -` seems flexible enough for this kind of use. If the data comes right after `data-binary = @-`, curl allows to pass request parameters before it, and the request body after it.

Thanks, I like the new design.

I'm surprised that this works though.  To be safe, we should verify (at least) with the default Curl binary versions that ship with Windows, MacOS and Linux.

gptel-request.el
```diff
@@ -114,32 +114,6 @@ Can be set to t, nil, or a string path to the curl executable."
 This only affects requests originating from Org mode buffers."
   :type 'boolean)
 
-(defcustom gptel-curl-file-size-threshold
```
karthink, line 117: - Please add an obsoletion for this user option (with `make-obsolete-variable`), otherwise we will be breaking the configs of users who have set this variable.
- Please add a NEWS entry (breaking changes section) about the removal of this defcustom, and briefly explain why it's not required any more.

gptel-request.el
```diff
@@ -2748,52 +2722,87 @@ See `gptel-curl--get-response' for its contents.")
 
 ;;; Curl request response handling
 
-(defun gptel-curl--get-args (info uuid include-headers)
```
karthink, line 2751: `gptel-curl--get-args` has only one consumer now, the dry-run system in gptel.el.  Please move this function to gptel.el around where the dry-run system calls it.

gptel.el
```diff
@@ -1751,8 +1751,7 @@ kill ring instead."
           (cl-assert (cl-typep gptel--fsm-last 'gptel-fsm))
           (plist-put (gptel-fsm-info gptel--fsm-last) :data data)
           (if copy                 ;Copy Curl command instead of sending request
-              (let ((args (gptel-curl--get-args (gptel-fsm-info gptel--fsm-last)
-                                                (md5 (format "%s" (random))) t)))
+              (let ((args (gptel-curl--get-args (gptel-fsm-info gptel--fsm-last) (md5 (format "%s" (random))))))
```
karthink: Formatting (see main review comment)

gptel-request.el
```diff
@@ -2810,46 +2819,31 @@ plist with the following keys, among others:
...
 the response is inserted into the current buffer after point."
-  (let* ((uuid (md5 (format "%s%s%s%s"
-                            (random) (emacs-pid) (user-full-name)
-                            (recent-keys))))
+  (let* ((uuid (md5 (format "%s%s%s%s" (random) (emacs-pid) (user-full-name) (recent-keys))))
          (info (gptel-fsm-info fsm))
          (backend (plist-get info :backend))
-         (args (gptel-curl--get-args info uuid nil))
          (stream (plist-get info :stream))
+         (command (nconc (list (gptel--curl-path)) (gptel-curl--get-config-args info) (list "--config" "-")))
```
karthink: Formatting (see main review comment)

karthink (changes requested):
Some minor changes.

Also, gptel has a soft line-width limit of 80 chars and hard limit of 100, please format long lines.

gptel-request.el
bohonghuang: @karthink Hmm... I'm a bit skeptical about the need to move `gptel-curl--get-args` to `gptel.el`. `gptel-curl--get-args` and `gptel-curl--get-config` are actually alike in structure, and if the former is moved to `gptel.el`, I'm worried they could easily drift out of sync if future contributors don't notice the relationship.

karthink: Yeah, that's a good point.  Let's leave it here.

karthink:
Tested on Linux (Curl 8.19.0) and merged optimistically.  If some version of Curl doesn't like the stdin double feeding, we'll hear soon from users, and I might revert this commit.

Thanks for the PR @bohonghuang!

karthink:
Is there a reason you used `utf-8-unix` encoding instead of `utf-8` in `gptel-curl--get-config`?

In `gptel-curl--get-args`:
```elisp
(data-json (decode-coding-string (gptel--json-encode data) 'utf-8 t)))
```

In `gptel-curl--get-config`:
```elisp
(data-json (encode-coding-string (gptel--json-encode data) 'utf-8-unix t))
```

bohonghuang:
> Is there a reason you used `utf-8-unix` encoding instead of `utf-8` in `gptel-curl--get-config`?
> 
> In `gptel-curl--get-args`:
> 
> ```emacs-lisp
> (data-json (decode-coding-string (gptel--json-encode data) 'utf-8 t)))
> ```
> 
> In `gptel-curl--get-config`:
> 
> ```emacs-lisp
> (data-json (encode-coding-string (gptel--json-encode data) 'utf-8-unix t))
> ```

@karthink I noticed that `gptel-curl-get-response` uses `utf-8-unix` for some reason, so I decided to keep it consistent with that. Tested to work fine on both Linux (Ubuntu 24) and Windows (built-in `curl` and the MSYS2 version). I tried to keep changes to `gptel-curl--get-args` to a minimum; the `utf-8` there should be something left over from before.

gptel-request.el
```diff
@@ -2748,52 +2755,87 @@ See `gptel-curl--get-response' for its contents.")
...
+         (items
+          (nconc
+           (cl-loop for (key . val) in headers
+                    collect (cons "header" (format "%s: %s" key val)))
+           (when (not (string-empty-p gptel-proxy))
+             (list (cons "proxy" gptel-proxy) "proxy-negotiate" (cons "proxy-user" ":")))
+           (list
+            (cons "url" url)
+            (cons "write-out" (format "(%s . %%{size_header})" uuid))
+            (cons "data-binary" '@-))))
```
karthink, line 2827: Why is `@-` a symbol instead of a string `"@-"`?

bohonghuang: > Why is `@-` a symbol instead of a string `"@-"`?

Because we want the `data-binary` entry in the curl config text to be:

```text
data-binary = @-
```

rather than:

```text
data-binary = "@-"
```

Using a symbol here conveniently makes `format` leave the value unquoted.

karthink: Oh, I didn't notice the format spec was "%S".  That's a nice trick.  It would be better not to pollute the obarray with meaningless symbols, but I guess json-parse does much more damage there already...

bohonghuang: If you really care, `'#:@-` is equivalent here, without interning the symbol. 🤣

Issue #1458, opened 2026-06-22
gptel-openai-oauth-login (OpenAI ChatGPT Plus/Pro [Codex Subscription])  : "Authorization Code Flow with PKCE"


u797595:
gptel-openai-oauth-login  doesn't seem to support[ Open Id Connect](https://openid.net/specs/openid-connect-core-1_0.html)  fully.

Specifically, the OAuth2 "Authorization Code Flow with PKCE"

Example:
https://auth.openai.com/oauth/authorize?response_type=code&client_id=app_EMoamEEZ73f0CkXaXp7hrann&redirect_uri=http%3A%2F%2Flocalhost%3A1455%2Fauth%2Fcallback&scope=openid+profile+email+offline_access&code_challenge=JQQqMVOWqZjQkWKg5SdJYhUf9PuTpIz8Rcqbc092rpY&code_challenge_method=S256&id_token_add_organizations=true&codex_cli_simplified_flow=true&state=aCpFfHv7WrrFN0lSC3s_goJyVG15UXDGkNiTtdzJ_gw&originator=opencode

gptel-openai-oauth-login seems to only support Oauth2 "Device Authorization Grant" 
Example:
https://auth.openai.com/codex/device

My organization hasn't enabled "device code authorization".

I must use OAuth2 "Authorization Code Flow with PKCE" flow for my "ChatGPT Plus/Pro (Codex Subscription)".

gptel-openai-oauth-login (OpenAI ChatGPT Plus/Pro [Codex Subscription])  : "Authorization Code Flow with PKCE"

ChetanKoneru:
@u797595 , Can you please test #1463.
Everything looks good on my end, I want someone else to do a smoke test and sign off on this.

u797595:
> [@u797595](https://github.com/u797595) , Can you please test [#1463](https://github.com/karthink/gptel/pull/1463). Everything looks good on my end, I want someone else to do a smoke test and sign off on this.

Unfortunately, I am not able to obtain access.  

`error_code: missing_required_parameter`

**Authentication Error**
```
A required parameter is missing. Please check the request and try again.
```



It looks like one small change is needed to the `scope` URL query parameter: 
```
scope=openid%20profile%20email%20offline_access
```
to
```
scope=openid+profile+email+offline_access
```

The `redirect_uri` URL query parameter value should probably be escaped too, but that seems to be tolerated by Chromium Web Browser.
```
redirect_uri=http://localhost:1455/auth/callback
```
[...]

PR #1463, opened 2026-07-01
Implement #1458 gptel-openai-oauth-login  - Authorization Code Flow with PKCE

ChetanKoneru:
Hello @karthink ,

Implementing OAuth2 `Authorization Code Flow with PKCE` support.

Should fix - https://github.com/karthink/gptel/issues/1458

this should also fix an existing bug

The `gptel-openai-oauth-login` interactive form (lines 109–116) uses `cl-find-if` to search `gptel--known-backends` for any registered OAuth backend:

```emacs-lisp name=gptel-openai-oauth.el url=https://github.com/karthink/gptel/blob/33ed0f6f69857157973bca72fb4d454e29965bdd/gptel-openai-oauth.el#L109-L116
(interactive (list
              (cond
               ((gptel-openai-oauth-p gptel-backend)
                gptel-backend)
               ((cl-find-if #'gptel-openai-oauth-p gptel--known-backends
                            :key #'cdr))   ;; <-- returns the CONS cell (name . backend)
               (t (user-error "No OpenAI OAuth backend found...")))))
```

**The problem:** `cl-find-if` returns the **full cons cell** `("codex" . #s(gptel-openai-oauth ...))` from the alist — not just the backend struct. But the function signature `(defun gptel-openai-oauth-login (&optional backend)` expects a `gptel-backend` struct, and the type check on line 118:

```emacs-lisp name=gptel-openai-oauth.el url=https://github.com/karthink/gptel/blob/33ed0f6f69857157973bca72fb4d454e29965bdd/gptel-openai-oauth.el#L118-L119
(unless (gptel-openai-oauth-p backend)
  (user-error "%s is not an OpenAI OAuth backend" (gptel-backend-name backend)))
```

…fails because `gptel-openai-oauth-p` receives a cons cell, not the struct. This triggers the `wrong-type-argument` error you see.

## The Fix

The `cl-find-if` call needs to extract the **`cdr`** (the actual backend struct) from the found cons cell. Change line 113:

```emacs-lisp
;; Before (buggy):
((cl-find-if #'gptel-openai-oauth-p gptel--known-backends
             :key #'cdr))

;; After (fixed):
((when-let* ((found (cl-find-if #'gptel-openai-oauth-p gptel--known-backends
                                :key #'cdr)))
   (cdr found)))
```

Or more concisely:

```emacs-lisp
((cdr (cl-find-if #'gptel-openai-oauth-p gptel--known-backends :key #'cdr)))
```

## Workaround I used

You can call `gptel-openai-oauth-login` **non-interactively**, passing the backend directly:

```emacs-lisp
(let ((codex-backend (gptel-get-backend "codex")))
  (gptel-openai-oauth-login codex-backend))
```

This bypasses the broken `(interactive ...)` form entirely.

Did a smoke test and it did work well. 

Are you OK with having an additional custom variable `gptel-openai-oauth-login-method`?

### DIFF
--- a/gptel-openai-oauth.el
+++ b/gptel-openai-oauth.el
@@ -22,6 +22,7 @@
 ;;
 
 ;;; Code:
+(eval-when-compile (require 'cl-lib))
 (require 'gptel-openai-responses)
 (require 'gptel-oauth)
 
@@ -32,6 +33,21 @@
 (defvar gptel--openai-oauth-token-file
   (expand-file-name ".cache/gptel-openai/openai-oauth-token"
                     user-emacs-directory))
+
+(defcustom gptel-openai-oauth-login-method 'authorization-code
+  "OAuth login method used by `gptel-openai-oauth-login'.
+
+The `authorization-code' method uses OAuth 2.0 Authorization Code
+Flow with PKCE and a localhost callback.  It is not supported when
+Emacs is running over SSH.  The `device' method uses OAuth 2.0
+Device Authorization Grant."
+  :type '(choice (const :tag "Authorization Code Flow with PKCE" authorization-code)
+                 (const :tag "Device Authorization Grant" device))
+  :group 'gptel)
+
+(defconst gptel--openai-oauth-redirect-port 1455)
+(defconst gptel--openai-oauth-redirect-timeout 300)
+(defconst gptel--openai-oauth-redirect-path "/auth/callback")
 
 ;;;; OpenAI OAuth backend
 (cl-defstruct (gptel-openai-oauth (:constructor gptel--make-openai-oauth)
@@ -56,7 +72,8 @@ Removes unsupported temperature settings from the payload."
       (cl-remf prompts-plist :max_output_tokens))
     prompts-plist))
 
-;;;; OpenAI OAuth login
+;;;; OpenAI OAuth methods
+;;;;; OpenAI device-based Oauth
 
 (defconst gptel--openai-oauth-poll-interval 2)
 (defconst gptel--openai-oauth-poll-timeout 30)
@@ -99,24 +116,8 @@ request times out."
     (or response
         (user-error "Timed out waiting for OpenAI OAuth device authorization"))))
 
-(defun gptel-openai-oauth-login (&optional backend)
-  "Authenticate BACKEND using the OpenAI device flow.
-
-If BACKEND is nil, use `gptel-backend'.  Interactively, prompt
-for or infer an OpenAI OAuth backend."
-  ;; TODO: Handle the case when there are multiple openai-oauth backends
-  ;; defined.  We want to pick the one to log into.
-  (interactive (list
-                (cond
-                 ((gptel-openai-oauth-p gptel-backend)
-                  gptel-backend)
-                 ((cl-find-if #'gptel-openai-oauth-p gptel--known-backends
-                              :key #'cdr))
-                 (t (user-error "No OpenAI OAuth backend found.  \
-Please set one up with `gptel-make-openai-oauth' first")))))
-  (unless backend (setq backend gptel-backend))
-  (unless (gptel-openai-oauth-p backend)
-    (user-error "%s is not an OpenAI OAuth backend" (gptel-backend-name backend)))
+(defun gptel--openai-oauth-login-with-device-code (backend)
+  "Authenticate BACKEND using OpenAI Device Authorization Grant."
   (pcase-let (((map :device_auth_id :user_code)
                (gptel--url-retrieve
                    (concat gptel--openai-oauth-url "/api/accounts/deviceauth/usercode")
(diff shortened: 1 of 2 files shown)


karthink:
@ChetanKoneru thanks for the PR.  Is this LLM generated?

ChetanKoneru:
Hi @karthink ,

I initially ported the functionality from here 

https://github.com/editor-code-assistant/eca/blob/master/src/eca/llm_providers/openai.clj

and then asked gpt-5.5 to "Review changes to  gptel-openai-oauth.el, verify that the changes did not introduce regression and also make sure that the changes match coding practices and patterns of the maintainer's."

So yes LLM has been used to enhance my code and make sure `device` login is not broken.

ChetanKoneru:
Tested here, see comments in #1458

karthink:
> So yes LLM has been used to enhance my code and make sure `device` login is not broken.

"Enhance" is unfortunately a subjective description, since I see the usual problems here -- LLM-induced collateral damage, removing comments and modifying lines unrelated to the PR.

Anyway, I've reviewed the PR, please see the comments.  I don't understand all the changes.  It looks like there are some changes unrelated to PKCE-based auth, like the changes involving `gptel--openai-oauth-account-id` and `gptel--openai-oauth-persist`.

> Are you OK with having an additional custom variable `gptel-openai-oauth-login-method`?

This is fine, but there are a bunch of other defcustoms introduced that I don't think should be defcustoms, or perhaps even exist.

gptel-openai-oauth.el
```diff
@@ -184,7 +395,7 @@ reauthenticate as needed."
                             gptel--openai-oauth-token-file)))
         (setf (gptel-openai-oauth-token backend) token-plist)
       (gptel-openai-oauth-login backend)))
-
```
karthink, line 187: Please undo this (LLM-produced) whitespace change.

gptel-openai-oauth.el
```diff
@@ -99,32 +124,169 @@ request times out."
...
+(defun gptel--openai-oauth-login-with-device-code (backend)
+  "Authenticate BACKEND using OpenAI Device Authorization Grant."
   (pcase-let (((map :device_auth_id :user_code)
                (gptel--url-retrieve
                    (concat gptel--openai-oauth-url "/api/accounts/deviceauth/usercode")
                  :method 'post
                  :data (list :client_id gptel--openai-oauth-client-id)
                  :headers `(("User-Agent" . ,(format "Emacs %s" emacs-version)))))
               (verification-uri (concat gptel--openai-oauth-url "/codex/device")))
-    ;; User authentication for user_code
```
karthink, line 127: Please undo this (LLM-produced?) comment removal.

gptel-openai-oauth.el
```diff
@@ -99,32 +124,169 @@ request times out."
...
-                 ((cl-find-if #'gptel-openai-oauth-p gptel--known-backends
-                              :key #'cdr))
-                 (t (user-error "No OpenAI OAuth backend found.  \
-Please set one up with `gptel-make-openai-oauth' first")))))
-  (unless backend (setq backend gptel-backend))
-  (unless (gptel-openai-oauth-p backend)
-    (user-error "%s is not an OpenAI OAuth backend" (gptel-backend-name backend)))
+(defun gptel--openai-oauth-backends ()
+  "Return registered OpenAI OAuth backend entries."
+  (cl-remove-if-not #'gptel-openai-oauth-p gptel--known-backends
```
karthink: `gptel--openai-oauth-backends` is only used in one place, please inline it.

gptel-openai-oauth.el
```diff
@@ -138,23 +300,72 @@ Please set one up with `gptel-make-openai-oauth' first")))))
                              ("client_id"     ,gptel--openai-oauth-client-id)
                              ("code_verifier" ,code_verifier)))
                     :content-type "application/x-www-form-urlencoded")))
-      ;; TODO Handle case where access_token was not granted
```
karthink, line 141: Is this case handled now?  It's not clear.  If it's not please add the TODO back.

gptel-openai-oauth.el
```diff
@@ -33,6 +36,28 @@
...
+The `authorization-code' method uses OAuth 2.0 Authorization Code
+Flow with PKCE and a localhost callback.  The `device' method uses
+OAuth 2.0 Device Authorization Grant."
+  :type '(choice (const :tag "Authorization Code Flow with PKCE" authorization-code)
+                 (const :tag "Device Authorization Grant" device))
+  :group 'gptel)
+
+(defcustom gptel-openai-oauth-redirect-port 1455
+  "Localhost port used for OpenAI OAuth authorization-code callbacks."
+  :type 'integer
```
karthink: Is the port expected to vary?  Why is this a defcsutom instead of being hardcoded?

gptel-openai-oauth.el
```diff
@@ -33,6 +36,28 @@
...
+  :group 'gptel)
+
+(defcustom gptel-openai-oauth-redirect-timeout 300
+  "Seconds to wait for an OpenAI OAuth authorization-code callback."
+  :type 'integer
```
karthink: Please turn this into a defconst, see `gptel--openai-oauth-poll-timeout`.

gptel-openai-oauth.el
```diff
@@ -33,6 +36,28 @@
...
+  :group 'gptel)
+
+(defconst gptel--openai-oauth-redirect-path "/auth/callback")
```
karthink, line 50: Please hardcode and remove this unless it's expected to change depending on your organization.

```suggestion
(defconst gptel--openai-oauth-redirect-path "/auth/callback")
```

gptel-openai-oauth.el
```diff
@@ -99,32 +124,169 @@ request times out."
...
+(defun gptel--openai-oauth-default-backend ()
+  "Return the current or first registered OpenAI OAuth backend."
+  (cond
+   ((gptel-openai-oauth-p gptel-backend)
+    gptel-backend)
+   ((cdr (car (gptel--openai-oauth-backends))))
+   (t (user-error "No OpenAI OAuth backend found.  \
+Please set one up with `gptel-make-openai-oauth' first"))))
+
+(defun gptel--openai-oauth-redirect-uri ()
```
karthink: Depending on the other comments this function may not be required.

gptel-openai-oauth.el
```diff
@@ -99,32 +124,169 @@ request times out."
...
+            (setq server
+                  (make-network-process
+                   :name "gptel-openai-oauth-callback"
+                   :server t
+                   :host "localhost"
+                   :service gptel-openai-oauth-redirect-port
+                   :filter #'filter
+                   :noquery t))
+            (message "OpenAI OAuth authorization URL: %s" authorization-url)
+            (browse-url authorization-url)
```
karthink, line 260: Calling browse-url directly has many edge cases, such as when you're connecting to Emacs over ssh.  See `gptel-oauth--device-auth-prompt`.

gptel-openai-oauth.el
```diff
@@ -99,32 +124,169 @@ request times out."
...
+  "Send PROCESS an HTTP response with STATUS, TITLE and BODY."
+  (let ((payload (format "<!doctype html><meta charset=\"utf-8\"><title>%s</title><p>%s</p>"
+                         title body)))
+    (process-send-string
+     process
+     (format "HTTP/1.1 %s %s\r\nContent-Type: text/html; charset=utf-8\r\nContent-Length: %d\r\nConnection: close\r\n\r\n%s"
+             status title (string-bytes payload) payload))))
+
+(defun gptel--openai-oauth-read-code (authorization-url state)
+  "Open AUTHORIZATION-URL and wait for a localhost callback matching STATE."
```
karthink, line 194: Indent this function.

gptel-openai-oauth.el
```diff
@@ -99,32 +124,169 @@ request times out."
...
+                 (delete-process process))
+         (filter (process string)
+                 (let ((request (concat (or (process-get process :gptel-request) "")
+                                        string)))
+                   (process-put process :gptel-request request)
+                   (when (string-match-p "\r\n\r\n" request)
+                     (pcase-let* ((`(:path ,path :query ,query)
+                                   (gptel--openai-oauth-callback-request request))
+                                  (callback-state
+                                   (gptel--openai-oauth-query-value "state" query))
```
karthink: Inline `gptel--openai-oauth-query-value`, as mentioned in above comment.

gptel-openai-oauth.el
```diff
@@ -99,32 +124,169 @@ request times out."
...
+             (error (user-error "%s" error))
+             (t (user-error "Timed out waiting for OpenAI OAuth callback"))))
+        (when (process-live-p server)
+          (delete-process server))))))
+
+(defun gptel--openai-oauth-login-with-authorization-code (backend)
+  "Authenticate BACKEND using OpenAI Authorization Code Flow with PKCE."
+  (let* ((redirect-uri (gptel--openai-oauth-redirect-uri))
+         (verifier (gptel-oauth--generate-code-verifier))
+         (state (gptel-oauth--generate-code-verifier))
```
karthink: I don't understand what `state` is for, or why it needs to be a PKCE-acceptable string.

gptel-openai-oauth.el
```diff
@@ -221,7 +433,7 @@ before constructing the headers."
           (host "chatgpt.com")
           (protocol "https")
           (endpoint "/backend-api/codex/responses")
-          (models
```
karthink, line 224: Undo LLM change

gptel-openai-oauth.el
```diff
@@ -138,23 +300,72 @@ Please set one up with `gptel-make-openai-oauth' first")))))
...
 it in BACKEND."
-  (pcase-let (((map :access_token :expires_in :id_token :refresh_token) token-plist))
-    ;; TODO Handle case where access_token was not granted
-    (unless (and access_token expires_in refresh_token)
-      (user-error "OpenAI OAuth Authentication failed"))
+  (pcase-let* (((map :access_token :expires_in :id_token :refresh_token)
+                token-plist)
+               (existing-token (gptel-openai-oauth-token backend))
+               (refresh-token (or refresh_token
+                                  (plist-get existing-token :refresh_token)))
```
karthink: Why is this needed?

gptel-openai-oauth.el
```diff
@@ -138,23 +300,72 @@ Please set one up with `gptel-make-openai-oauth' first")))))
...
+               (id-payload (gptel-oauth--jwt-payload id_token))
+               (access-payload (gptel-oauth--jwt-payload access_token))
+               (account-id
+                (or (gptel--openai-oauth-account-id id-payload access-payload)
+                    (plist-get existing-token :account_id))))
```
karthink: Why are you storing the account-id here?

gptel-openai-oauth.el
```diff
@@ -138,23 +300,72 @@ Please set one up with `gptel-make-openai-oauth' first")))))
...
+           ('device
+            (gptel--openai-oauth-login-with-device-code backend))
+           (login-method
+            (user-error "Unknown OpenAI OAuth login method: %S" login-method)))))
+    (when (and (called-interactively-p 'interactive)
+               (plist-get token-plist :access_token))
+      (message "Successfully logged in to OpenAI OAuth."))
+    token-plist))
+
+(defun gptel--openai-oauth-account-id (id-payload access-payload)
```
karthink: Why is this function needed?

gptel-openai-oauth.el
```diff
@@ -203,7 +414,8 @@ before constructing the headers."
   (let* ((token (gptel-openai-oauth-token gptel-backend))
          (key (plist-get token :access_token))
          (account-id
-          (or (map-nested-elt
+          (or (plist-get token :account_id)
+              (map-nested-elt
                token '( :id_token :https://api.openai.com/auth
                         :chatgpt_account_id))
```
karthink: Conversely, if you have `gptel--openai-oauth-account-id` shouldn't you be using it here?  (But I'd prefer to remove `gptel--openai-oauth-account-id` unless there's a reason to keep it.)

karthink (changes requested):

gptel-openai-oauth.el
ChetanKoneru: Not needed

gptel-openai-oauth.el
ChetanKoneru: Not needed

ChetanKoneru:
Thank you for the review, some of the space and formatting were inflicted by my formatter, while LLM introduced AI slop that I missed while glancing. Will pay more attention in the future.

gptel-openai-oauth.el
```diff
@@ -99,24 +116,161 @@ request times out."
...
+(defun gptel--openai-oauth-send-callback-response (process status title body)
+  "Send PROCESS an HTTP response with STATUS, TITLE and BODY."
+  (let ((payload (format "<!doctype html><meta charset=\"utf-8\"><title>%s</title><p>%s</p>"
+                         title body)))
+    (process-send-string
+     process
+     (format "HTTP/1.1 %s %s\r\nContent-Type: text/html; charset=utf-8\r\nContent-Length: %d\r\nConnection: close\r\n\r\n%s"
+             status title (string-bytes payload) payload))))
+
+(defun gptel--openai-oauth-auth-prompt (authorization-url)
```
karthink: This function is very similar to `gptel-oauth--device-auth-prompt`.  Consider reusing that function instead if possible, modifying it if necessary.

gptel-openai-oauth.el
```diff
@@ -99,24 +116,161 @@ request times out."
...
+                           "OpenAI OAuth authorization succeeded.  You may close this tab."
+                           callback-code))
+                  (t
+                   (finish process "400" "OpenAI OAuth Error"
+                           "OpenAI OAuth callback did not include a code.  You may close this tab."
+                           nil "OpenAI OAuth callback did not include a code"))))))))
+      (unwind-protect
+          (progn
+            (setq server
+                  (make-network-process
```
karthink, line 246: There's still an SSH issue here.  Suppose the user is connecting to Emacs over SSH.  This callback server starts on the remote.  `gptel--openai-oauth-auth-prompt` then directs them to open the authorization URL in the local browser.  The redirect from the OpenAI server will try to run on localhost in the local browser, i.e. the SSH client machine, and the authentication will fail.

I don't see an easy way to fix this -- the only way I can think of is if the SSH connection reverse-tunnels `gptel--openai-oauth-redirect-port`, which we don't control.

Any user connecting over SSH will be utterly confused.  We can at least avoid that by 

- disallowing authorization-code based authentication when connecting over SSH.  (Then there will be no need for `gptel--openai-oauth-auth-prompt` either.)
- Or asking the user to connect via `ssh -R 1455:localhost:1455 ...` first (replacing 1455 with `gptel--openai-oauth-auth-prompt`.)

karthink (changes requested):

gptel-openai-oauth.el
ChetanKoneru: Handled by disabling auth-code flow over SSH.

The callback server runs where Emacs runs. Over SSH that’s the remote machine, but the browser redirect goes to `localhost` on the local machine, so the callback never reaches Emacs.

So now `gptel--openai-oauth-read-code` checks `SSH_CLIENT`, `SSH_CONNECTION`, or `SSH_TTY` before starting the server. If any are set, it raises:

```elisp
OpenAI authorization-code login requires a local browser callback and is not supported over SSH. Set `gptel-openai-oauth-login-method' to `device' and retry
```

Because SSH is rejected before prompting, `gptel--openai-oauth-auth-prompt` is no longer needed and was removed.

I chose this over suggesting `ssh -R` because gptel can’t know or control whether the tunnel exists.

Correct me if I am wrong

gptel-openai-oauth.el
ChetanKoneru: make sense, done

gptel-openai-oauth.el
karthink: I think disallowing auth-code-based login by default when using SSH is fine for now.  In a future PR this can be improved by providing the SSH tunneling instructions and throwing up a confirmation prompt

"This requires a SSH tunnel (ssh -R 1405:localhost:1405 ...). Do you still want to try code-based authentication? (y/n)"

so users who absolutely need it still have the option to reconnect with the tunnel and authenticate.

karthink:
Merged with some code reorganization.  Thanks for the PR @ChetanKoneru.

There's an easy way to allow auth-code based OAuth over SSH.  Instead of running a local server like in `gptel--openai-oauth-read-code`, we can let the callback "fail" in the user's local browser.  We just have to ask the user to copy the (long) callback URL from the browser's URL bar into Emacs.  The rest of the handling is unchanged.

Essentially we make the user do one extra manual step that the local server performs automatically right now.  I merged this PR since its scope is large enough already, but this can be an additional PR for the future.

karthink:
Another task here is to extract most of the server logic from `gptel--openai-oauth-read-code` into a function in `gptel-oauth.el`.  It's pretty generic and can be reused for OAuth login for other backends, like Claude.

ChetanKoneru:
Thanks @karthink ,
I will look in to it, when I get a chance.

PR #1473, opened 2026-07-07
gptel-openai-oauth: Don't wait for RET to open authorization URLs

djr7C4:
The current OAuth flow asks the user to press enter before opening the browser which is unnecessary.

### DIFF
--- a/gptel-oauth.el
+++ b/gptel-oauth.el
@@ -112,10 +112,9 @@ when appropriate for the current session."
            (format "(One-time code %s copied) Visit %s in your local browser, \
 enter the code and authorize.  Press ENTER after authorizing: "
                    user-code verification-uri)))
-      (read-from-minibuffer
-       (format "(One-time code %s copied) Press ENTER to open the authorization page. \
+      (message "One-time code %s copied. \
 If your browser does not open automatically, browse to %s: "
-               user-code verification-uri))
+               user-code verification-uri)
       (browse-url verification-uri)
       (read-from-minibuffer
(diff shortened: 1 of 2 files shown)


karthink:
Is it unnecessary?  What happens when `(browse-url verification-uri)` runs is not well-defined.  Depending on the user's window manager settings, the browser may take focus or emacs may stay focused.   If you see the message "(One-time code %s copied) Press Enter after Authorizing" but the browser did not take focus, it might be confusing, don't you think?  On the other hand, if the browser did take focus immediately, then it's not clear that the one-time code has been copied to the clipboard and may be pasted in the right browser prompt.

djr7C4:
> If you see the message "(One-time code %s copied) Press Enter after Authorizing" but the browser did not take focus, it might be confusing, don't you think?

I suppose so but it could be addressed by improving the message and would only be confusing the first time.

> On the other hand, if the browser did take focus immediately, then it's not clear that the one-time code has been copied to the clipboard and may be pasted in the right browser prompt.

This is true. In my window manager, my browser opens to the side and I can see both windows. I have to refresh tokens once a week anyway (as they expire) so I'm used to how it opens even if my window manager didn't behave that way. 

One option is to add a configuration option and make the current behavior the default.

djr7C4:
Also, even in most window managers where the browser could open in an iconified frame, there is some sort of visual  indication that a new window has been opened.

karthink:
I think it's better to be explicit about this, especially since this is an action that typically only runs once every few days.

PR #1483, opened 2026-07-15
Remove deprecated or no longer available Copilot models

igmacs:
Some Copilot models are no longer available. I realized when my default model, gpt-5.2, stopped working, but after checking I realized a lot more don't work. I checked and some depend on the specific subscription plan ([source](https://docs.github.com/en/copilot/reference/ai-models/supported-models#supported-ai-models-per-copilot-plan)), but some like gpt-5.2 don't. The model retirement history can be found [here](https://docs.github.com/en/copilot/reference/ai-models/supported-models#model-retirement-history), although it's not completely reliable (for example gpt-4.1 should have been retired according to that, but it still works for me).

I checked the commit history and I saw the the list of available models is regularly kept up to date for other providers, so I'm submitting this PR, using as reference https://github.com/karthink/gptel/commit/5c82ff85be0beed57a923935e18b9c4d1a8d0858.

There are also some models listed in the documentation as available that are not supported yet by gptel. I will try those available in my subscription and add them in a different PR

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -8,6 +8,12 @@
   removed from the default list of Deepseek models.  These models are
   either deprecated or no longer available.
 
+- The models =gptel-5-1=, =gpt-5.1-codex=, =gpt-5.1-codex-max=,
+  =gpt-5.1-codex-mini=, =gpt-5.2=, =gpt-5.2-codex=, =claude-sonnet-4=,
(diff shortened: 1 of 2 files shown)


karthink:
@igmacs Thanks for checking, and thanks for the PR!

PR #1485, opened 2026-07-16
Update News for remove soon to be deprecated deepseek-chat and deepseek-reasoner model.

beacoder:
image

According to deepseek's api doc, these two models will be deprecated soon.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,10 @@
 
 ** Breaking changes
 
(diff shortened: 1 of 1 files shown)


beacoder:
Since I'm not familiar with merging two patches into one PR, so I created two PRs for this removing model work.
sorry for the inconvenience.

Issue #1491, opened 2026-07-23
`gptel-rewrite` causes Emacs v31.0.90 to segfault


oreodave:
### Please update gptel first -- errors are often fixed by the time they're reported.

- [x] I have updated gptel to the latest commit and tested that the issue still exists

### Bug Description

Disclaimer: I do believe I've already found the root cause for this
bug.  Therefore, this issue report mostly serves as justification for
it.

When performing `gptel-rewrite` on Emacs v31.0.90 I got a segfault.
I've tested this against the latest version of `gptel`.  This doesn't
occur when using Emacs v30.2, nor does it occur in v31.0.90 when using
a TTY frame.  This suggests that there's something going on in the GUI
display code within Emacs specifically that's causing the segfault.

### Backend

None

### Steps to Reproduce

1) Mark some text in any buffer
2) Call `gptel-rewrite`, or use the transient if so desired.
3) SEGFAULT!

### Additional Context

Emacs version: 31.0.90
Operating System: Arch Linux

### Backtrace

```emacs-lisp
N/A
```

### Log Information

```json
N/A
```

oreodave:
Turns out I was completely wrong regarding the fix - `:align-to` has a
perfectly fine specification.  I'll need some advice here to debug
this further.
The issue is, because of the multiple handlers dealing with the
various states of the FSM in `gptel-rewrite` - as well as the fact
that the bug happens during a `redisplay` - it's kinda difficult to
get a handle on which specific overlay adjustment is causing the bug
(if it's even that in the first place).

karthink:
> I'll need some advice here to debug this further.

Emacs segfaulting is not really debuggable from the Lisp runtime inside Emacs, and is unlikely to be a gptel issue.  Since you're on the pre-release or master branch, I recommend running Emacs from gdb, producing the crash and posting the backtrace to the Emacs devel mailing list.  You can format the issue with M-x report-emacs-bug.

oreodave:
Definitely agree that I can't see how this is a `gptel` specific issue since Emacs is segfaulting.
I'll close this issue report it to Emacs devel itself.
Thank you :)

PR #1492, opened 2026-07-23
Fix for #1491

oreodave:
As described in #1491, `gptel-rewrite` causes segfaults in Emacs
v31.0.90.

As it turns out this is due to an illegal `:align-to` spec when
propertising the `rewrite`.  As per `info:elisp#Specified Space`,
we're allowed to have pixel specifications for `:align-to`.
`info:elisp#Pixel Specification` indicates that a pixel spec with an
OP should have the form `(OP EXPR...)`.

The fix here is changing `(OP EXPR (EXPR))` (incorrect) to `(OP EXPR EXPR)`.  
There were two instances I found in the `gptel` source code,
in `gptel-rewrite` (the impetus for my exploration) as well as in the
header-line code for `gptel-mode` so I fixed them.

### DIFF
--- a/gptel-rewrite.el
+++ b/gptel-rewrite.el
@@ -218,7 +218,7 @@ which see."
             " " 'display
             (if (and (fboundp 'string-pixel-width)
                      (display-graphic-p))
-                `(space :align-to (- right (,(string-pixel-width hint-str))))
+                `(space :align-to (- right ,(string-pixel-width hint-str)))
               `(space :align-to (- right ,(+ 1 (string-width hint-str))))))
            (propertize hint-str 'face '(warning default)))) ;status element 3
     (overlay-put ov 'before-string (apply #'concat (overlay-get ov 'status)))))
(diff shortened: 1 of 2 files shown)


oreodave:
BTW - I double checked this fix against Emacs v30.2 as well and it's completely fine - this doesn't cause a regression with the current stable version of Emacs.  Yipee!

oreodave:
Wait actually... looking at the spec:
```
       EXPR ::= NUM | (NUM) | UNIT | ELEM | POS | IMAGE | XWIDGET | FORM
```
... and the fact that:
```
   The form NUM specifies a fraction of the default frame font height or
width.  The form ‘(NUM)’ specifies an absolute number of pixels. 
```
...means that the original code was correct for the usecase (pixel specific display).
So this really is a bug with Emacs v31.0.90.

oreodave:
Closing this PR as this doesn't really fix the core issue - which is Emacs segfaulting during a completely valid :align-to spec.

PR #1495, opened 2026-07-27
gptel-request: Prevent failed Emacs encoding from leaking into json

yantar92:
* gptel-request.el (gptel--json-encode): Make that that raw bytes originating from failed Emacs decoding are never passed to `json-serialize' which simply errs on such strings. (gptel--sanitize-string):
(gptel--sanitize-for-json): New helper functions to replace raw bytes with \xhex strings (as they are usually displayed in Emacs) in strings and json lists/vectors recursively.

The unibyte characters may be coming from ordinary Emacs buffers if they are included into context, from custom tools, etc.  Intercepting them at the very last moment, before serializing is the most robust way to fix this problem.

Assisted-by: DeepSeek + macher.el + custom tools

### DIFF
--- a/gptel-request.el
+++ b/gptel-request.el
@@ -859,19 +859,60 @@ when including context from these major modes.")
     `(let ((json-object-type 'plist))
       (json-read-from-string ,str))))
 
+(defun gptel--sanitize-string (str)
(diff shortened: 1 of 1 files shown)


yantar92:
Hmm. Duplicate of #1455.

Issue #15, opened 2023-03-10
Query can be restarted multiple times


minad:
When pressing C-c RET on a headline multiple times, it seems possible to restart the query. This does not seem desired. It would be better if gptel prints an error message if a query is already running and hasn't completed yet.

karthink:
Yes, this needs to be fixed.  I think multiple queries from one buffer is fine -- for example if you are in a non-dedicated (i.e. regular) buffer and want to ask multiple unrelated queries:

``` text
How many satellites are in orbit around the earth?

What is speedrunning in videogames and when did this practice begin?
```

You can select each line and call `gptel-send`.  However, we want to avoid sending the same request multiple times. I'll create a list of hashes (buffer-local) for currently active queries and check against that.  If you have a better idea please let me know.

minad:
I think it is okay to keep it simple and just allow a single requetat a time, since the API is not fast anyway.

karthink:
My reason for enabling multiple parallel requests per buffer is _because_ the API is not fast.  That said one request per buffer is a simpler solution, I can implement it and see if it feels limiting in usage.

minad:
Sure, but I usually start a query and then refine the prompt. Starting multiple independent queries seems only useful if you run multiple "conversations" in parallel. There is also the danger of hitting some rate limit. Maybe each gptel buffer could run independently.

guibor:
@karthink perhaps it would be possible to configure a `gptel-send-hook` variable, and for anyone interested in avoiding multiple parallel queries it would be possible to add a `gptel-abort` to the hook?
It would be imperfect as it would not be buffer-specific but users could configure that, or even make the hook local to certain types of buffers (e.g if using the gptel conversational buffer it is more likely that this behavior would be desired)

karthink:
@guibor here is a recipe for disallowing multiple requests from one buffer -- only works with Curl though:

``` emacs-lisp
(defun gptel-throttle-request (orig-fn &rest args)
  (if (seq-some (lambda (request)
                  (eq (current-buffer)
                      (plist-get request :buffer)))
                (map-values gptel-curl--process-alist))
      (message "gptel request cancelled: there is an active request in buffer %s"
               (buffer-name (current-buffer)))
    (apply orig-fn args)))

(advice-add 'gptel-request :around #'gptel-throttle-request)
```

I can add it as an option to gptel but it needs more work to support `url-retrieve`.

guibor:
[...]

PR #1496, opened 2026-07-30
Allow tool calls to be rejected, yet continuing the conversation

yantar92:
* gptel.el (gptel--reject-tool-calls-and-continue): New function that feeds "Tool call rejected by user." (or custom response from prompt) as all pending tool results and continues the REACT chain. (gptel--inspect-reject-tool-calls-and-continue): A counterpart of `gptel--inspect-reject-tool-calls' inside tool inspection buffer. (gptel-tool-call-actions-map):
(gptel--display-tool-calls):
(gptel-tool-call-inspection-map):
(gptel--inspect-tool-calls): Bind the rejection to C-c C-r and update help hints accordingly.
* NEWS (New features and UI changes): Document the new feature.

This feature is something I often miss when I see that the reasoning chain is getting in completely wrong direction and start doing crazy staff like editing files via python instead of using dedicated tools or simply trying to call heavy tool that will waste time, tokens, and lead to nothing.  Without this feature, the only option is canceling the whole reasoning chain, sometimes long and involving complicated finished tool calls like searching web or querying databases.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -154,6 +154,16 @@
   starting MCP servers.  Previously all tools provided by an MCP server
   would be added automatically.  See ~gptel-mcp-connect~ for details.
 
+- When tool calls require confirmation, you can now choose to reject
+  them and continue the conversation instead of stopping it entirely.
+  Choosing "Reject (continue)" feeds a rejection message back to the
+  LLM for each tool call, allowing the model to adapt and proceed
+  without running the tools.
+
+  This option is available in the tool call dispatch overlay (~C-c C-r~),
+  the minibuffer confirmation prompt (=r=), and the tool call inspection
+  buffer (~C-c C-r~ or the "Reject" button in the header line).
+
 ** Notable bug fixes
 
 - Improved error messages: Curl request failures now include more
--- a/NEWS
+++ b/NEWS
@@ -118,6 +118,16 @@
   outputs simultaenously in a request.  Previously the two were
   incompatible.
 
+- When tool calls require confirmation, you can now choose to reject
+  them and continue the conversation instead of stopping it entirely.
+  Choosing "Reject (continue)" feeds a rejection message back to the
+  LLM for each tool call, allowing the model to adapt and proceed
+  without running the tools.
+
+  This option is available in the tool call dispatch overlay (~C-c C-r~),
+  the minibuffer confirmation prompt (=r=), and the tool call inspection
+  buffer (~C-c C-r~ or the "Reject" button in the header line).
+
 ** Notable bug fixes
 
 - Improved error messages: Curl request failures now include more
--- a/gptel.el
+++ b/gptel.el
@@ -2034,6 +2034,7 @@ Note: This tool call preview API is currently experimental.")
   "<mouse-1>" #'gptel--dispatch-tool-calls
   "C-c C-c" #'gptel--accept-tool-calls
   "C-c C-k" #'gptel--reject-tool-calls
+  "C-c C-r" #'gptel--reject-tool-calls-and-continue
   "C-c C-i" #'gptel--inspect-tool-calls)
 
 (defun gptel--display-tool-calls (tool-calls info &optional use-minibuffer)
@@ -2068,20 +2069,23 @@ USE-MINIBUFFER is non-nil)."
                                  backend-name len (if (> len 1) "calls" "call")
                                  tool-call-names))
                  (choices '((?y "Run tools") (?n "Cancel (resumable)")
-                            (?i "Inspect or edit")))
+                            (?r "Reject (continue)") (?i "Inspect or edit")))
                  (choice (read-multiple-choice prompt choices)))
             (pcase (car choice)
               (?y (gptel--accept-tool-calls tool-calls))
               (?n (gptel--reject-tool-calls))
+              (?r (gptel--reject-tool-calls-and-continue tool-calls))
               (?i (gptel--inspect-tool-calls tool-calls info))))
         ;; Prompt for confirmation from the response buffer
         (let* ((backend-name (gptel-backend-name (plist-get info :backend)))
                (actions-string
                 (concat (propertize "Run tools: " 'face 'font-lock-string-face)
                         (propertize "C-c C-c" 'face 'help-key-binding)
-                        (propertize ", Cancel request: " 'face 'font-lock-string-face)
+                        (propertize ", Cancel: " 'face 'font-lock-string-face)
                         (propertize "C-c C-k" 'face 'help-key-binding)
-                        (propertize ", Inspect or Edit: " 'face 'font-lock-string-face)
+                        (propertize ", Reject: " 'face 'font-lock-string-face)
+                        (propertize "C-c C-r" 'face 'help-key-binding)
+                        (propertize ", Inspect: " 'face 'font-lock-string-face)
                         (propertize "C-c C-i" 'face 'help-key-binding)))
                (confirm-strings)
                ;; FIXME(tool) use a wrapper instead of a manual text-property search,
@@ -2335,16 +2339,46 @@ OV is the tool call dispatch overlay."
                          (overlay-end prompt-ov)))))
     (delete-overlay ov)))
 
+(defun gptel--reject-tool-calls-and-continue (&optional tool-calls ov)
+  "Reject pending TOOL-CALLS and continue the request chain.
+
+Feed a rejection message to the LLM for each tool call so the
+conversation can proceed.  OV is the tool call dispatch overlay."
+  (interactive (pcase-let ((`(,resp . ,o) (get-char-property-and-overlay
+                                           (point) 'gptel-tool)))
+                 (list resp o)))
+  (gptel--update-status " Tools rejected" 'error)
+  (when tool-calls
+    (let* ((default "Tool call rejected by user.")
+           (msg (read-string
+                 (format-prompt "Rejection message: " default)
+                 nil nil default)))
+      (cl-loop for (_tool-spec _arg-plist process-tool-result) in tool-calls
+               do (funcall process-tool-result msg))))
+  (when (and (overlayp ov) (overlay-buffer ov))
+    (with-current-buffer (overlay-buffer ov)
+      (when-let* ((preview-handles (overlay-get ov 'previews)))
+        (dolist (func-to-handle preview-handles)
+          (when (car func-to-handle) (apply func-to-handle))))
+      (dolist (prompt-ov (overlay-get ov 'prompt))
+        (when-let* (((overlay-buffer prompt-ov))
+                    (inhibit-read-only t))
+          (delete-region (overlay-start prompt-ov)
+                         (overlay-end prompt-ov)))))
+    (delete-overlay ov)))
+
 (defun gptel--dispatch-tool-calls (choice)
   "Dispatch on tool-calls with CHOICE."
   (interactive
    (list
     (let ((choices '((?y "yes") (?n "do nothing")
-                     (?k "cancel request") (?i "inspect call(s)"))))
+                     (?k "cancel request") (?r "reject (continue)")
+                     (?i "inspect call(s)"))))
       (read-multiple-choice "Run tool calls? " choices))))
   (pcase (car choice)
     (?y (call-interactively #'gptel--accept-tool-calls))
     (?k (call-interactively #'gptel--reject-tool-calls))
+    (?r (call-interactively #'gptel--reject-tool-calls-and-continue))
     (?i (call-interactively #'gptel--inspect-tool-calls))))
 
 ;;;; Tool call inspection UI
@@ -2352,6 +2386,7 @@ OV is the tool call dispatch overlay."
   :doc "Actions in the gptel tool inspection buffer."
   "C-c C-c" #'gptel--inspect-accept-tool-calls
   "C-c C-k" #'gptel--inspect-reject-tool-calls
+  "C-c C-r" #'gptel--inspect-reject-tool-calls-and-continue
   "C-c C-i" #'gptel--inspect-quit-tool-calls)
 
 (defun gptel--inspect-accept-tool-calls (&optional _)
@@ -2414,8 +2449,18 @@ This is a bug, please report it!"))))
   "Cancel tool-calls and return to query buffer."
   (interactive)
   (apply #'gptel--reject-tool-calls
-   (thread-first (gptel-fsm-info gptel--fsm-last)
-                 (plist-get :tool-display)))
+         (thread-first (gptel-fsm-info gptel--fsm-last)
+                       (plist-get :tool-display)))
+  (quit-window t))
+
+(defun gptel--inspect-reject-tool-calls-and-continue (&optional _)
+  "Reject tool-calls from the inspection buffer and continue the chain.
+
+Feed a rejection message to the LLM for each tool call and clean up."
+  (interactive)
+  (apply #'gptel--reject-tool-calls-and-continue
+         (thread-first (gptel-fsm-info gptel--fsm-last)
+                       (plist-get :tool-display)))
   (quit-window t))
 
 (defun gptel--inspect-quit-tool-calls (&optional _)
@@ -2499,6 +2544,8 @@ query buffer."
               " \\[gptel--inspect-accept-tool-calls], "
               (buttonize "Cancel" #'gptel--inspect-reject-tool-calls)
               " \\[gptel--inspect-reject-tool-calls], "
+              (buttonize "Reject" #'gptel--inspect-reject-tool-calls-and-continue)
+              " \\[gptel--inspect-reject-tool-calls-and-continue], "
               (buttonize "Return" #'gptel--inspect-quit-tool-calls)
               " \\[gptel--inspect-quit-tool-calls], "
               (buttonize "Edit" (lambda (_) (read-only-mode 'toggle)))
--- a/gptel.el
+++ b/gptel.el
@@ -2031,6 +2031,7 @@ Note: This tool call preview API is currently experimental.")
   "<mouse-1>" #'gptel--dispatch-tool-calls
   "C-c C-c" #'gptel--accept-tool-calls
   "C-c C-k" #'gptel--reject-tool-calls
+  "C-c C-r" #'gptel--reject-tool-calls-and-continue
   "C-c C-i" #'gptel--inspect-tool-calls)
 
 (defun gptel--display-tool-calls (tool-calls info &optional use-minibuffer)
@@ -2065,20 +2066,23 @@ USE-MINIBUFFER is non-nil)."
                                  backend-name len (if (> len 1) "calls" "call")
                                  tool-call-names))
                  (choices '((?y "Run tools") (?n "Cancel (resumable)")
-                            (?i "Inspect or edit")))
+                            (?r "Reject (continue)") (?i "Inspect or edit")))
                  (choice (read-multiple-choice prompt choices)))
             (pcase (car choice)
               (?y (gptel--accept-tool-calls tool-calls))
               (?n (gptel--reject-tool-calls))
+              (?r (gptel--reject-tool-calls-and-continue tool-calls))
               (?i (gptel--inspect-tool-calls tool-calls info))))
         ;; Prompt for confirmation from the response buffer
         (let* ((backend-name (gptel-backend-name (plist-get info :backend)))
                (actions-string
                 (concat (propertize "Run tools: " 'face 'font-lock-string-face)
                         (propertize "C-c C-c" 'face 'help-key-binding)
-                        (propertize ", Cancel request: " 'face 'font-lock-string-face)
+                        (propertize ", Cancel: " 'face 'font-lock-string-face)
                         (propertize "C-c C-k" 'face 'help-key-binding)
-                        (propertize ", Inspect or Edit: " 'face 'font-lock-string-face)
+                        (propertize ", Reject: " 'face 'font-lock-string-face)
+                        (propertize "C-c C-r" 'face 'help-key-binding)
+                        (propertize ", Inspect: " 'face 'font-lock-string-face)
                         (propertize "C-c C-i" 'face 'help-key-binding)))
                (confirm-strings)
                ;; FIXME(tool) use a wrapper instead of a manual text-property search,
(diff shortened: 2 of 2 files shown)


karthink:
Thanks for the PR Ihor.  I agree with the general idea of guiding the model mid chain-of-thought, but I'm not sure yet if this is the best way to do it.

- In the past when I tried sending a generic "tool call rejected by user" command (using `gptel-pre-tool-call-functions`), most frontier models stubbornly continued to try similar things -- using the same or other tools, like trying to write a file with shell commands when I reject a `write_file` tool call.  This was around Dec 2025.
- Sending a custom rejection message like "You're trying to write the file in the wrong directory, write it in ./foo/" as the result of a `write_file` tool call seems like a hack -- it's including a user instruction as the output of a tool call.

I've been playing around some other ideas for steering.  The most promising one is to send a generic rejection message as the tool call message but append to it a regular user prompt with the steering instructions read from the minibuffer prompt.  I haven't added this to gptel because it's both trickier to implement and invalidates the cache on the next call to gptel-send.

I'll try using this PR for a while and see how it does.

yantar92:
karthink ***@***.***> writes:

> Thanks for the PR Ihor.  I agree with the general idea of guiding the model mid chain-of-thought, but I'm not sure yet if this is the best way to do it.
>
> - In the past when I tried sending a generic "tool call rejected by user" command (using `gptel-pre-tool-call-functions`), most frontier models stubbornly continued to try similar things -- using the same or other tools, like trying to write a file with shell commands when I reject a `write_file` tool call.  This was around Dec 2025.

This may still be similar if you use generic rejection message.
AFAIU, most modern LLMs are trained to handle errors during tool calls,
including access denied and similar.
They try workarounds or different tools in such cases, which is often
useful, but sometimes undesired. And sometimes leads to models hacking
huggingface :)

> - Sending a custom rejection message like "You're trying to write the file in the wrong directory, write it in ./foo/" as the result of a `write_file` tool call seems like a hack -- it's including an instruction as the output of a tool call.

I had more success with this option, but sometimes even a generic
message is enough. If the model keeps being stubborn, I explain why its
current line of reasoning is not right.

Is it a hack? Yes. Moreover, models may steer away from following such
instructions in the future, fighting context poisoning attacks.
We may need to amend the tool description to accept rejection as a
viable return value, but I do not think that such complication is
necessary just yet.

> I've been playing around some other ideas for steering.  The most promising one is to send a generic rejection message as the tool call message but append to it a regular user prompt with the steering instructions read from the minibuffer prompt.  I haven't added this to gptel because it's both trickier to implement and invalidates the cache on the next call to gptel-send.

This sounds cleaner - adding a proper user message.
I guess we would need to make gptel insert some kind special
message-in-the-middle block into chain of thought to keep the cache
valid. Indeed more tricky.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

karthink:
> > I've been playing around some other ideas for steering.  The most promising one is to send a generic rejection message as the tool call message but append to it a regular user prompt with the steering instructions read from the minibuffer prompt.  I haven't added this to gptel because it's both trickier to implement and invalidates the cache on the next call to gptel-send.
> 
> This sounds cleaner - adding a proper user message.  I guess we would need to make gptel insert some kind special message-in-the-middle block into chain of thought to keep the cache valid. Indeed more tricky.

I've implemented this on top of your branch in the `feature-steering` branch.  Could you try it out and let me know what you think?  From looking at the dry-run outputs, it handles reasoning blocks correctly.

I'm calling this feature "steering" instead of "reject-and-continue" in gptel.  I think some other harnesses also call this feature "steering messages".

My plan is to allow adding a steering message to the buffer at any time, not just when rejecting a tool call.  So you should be able to send a query, and type in more text into the buffer below an ongoing response. The message will be picked up and sent along with the next tool call result.

Exactly how to set up a steering message is not clear.  There are several ways: 
- A special prefix argument to gptel-send (1?) to send the prompt as a steering message and enqueue it instead of sending
- A new command `gptel-send-steer`  or `gptel-steer` to send the prompt as a steering message
- A command to mark a region as a steering message for a response above the region.

In all cases there will be some edge cases to work out, but I think it's useful.  I've also found myself wanting to correct the model when it's off on a wild goose chase.

karthink:
> My plan is to allow adding a steering message to the buffer at any time, not just when rejecting a tool call.

@yantar92 I've implemented this now in the `feature-steering` branch.  You can add a steering message to the in-flight response at any time, not just when accepting/rejecting tool calls.  The steering message will be sent on the next user turn, which is usually with the next tool results.

I know that you like to use a single gptel-send invocation to send LLMs on a long-duration task, so you might appreciate the "steer anytime" feature -- you don't need to require tool confirmation for the opportunity to steer the LLM. 

Currently this is only accessible from `gptel-menu`, the command should show up when there is an active LLM query in the buffer.  (It's bound to `M-RET`.)

I'm undecided on how to implement this interface more seamlessly.  I don't want to add more top-level commands like `gptel-steer`.  I like the idea of limiting the surface area of gptel to `gptel-send`, and shoving more advanced commands into the transient menu instead.

------

One idea I have is to provide two presets, `@gptel-steer` and `@gptel-queue` with gptel.
- Using `@gptel-steer` in the prompt in the buffer will cause the prompt to be added as a steering message to an ongoing request (or throw a warning if there isn't one).  It will be removed from the buffer.
- Using `@gptel-queue` in the prompt in the buffer will cause the prompt to be queued as the next `gptel-send`, sending it automatically after the current response fully ends.  If there is no ongoing request it does nothing, and the prompt is sent as a regular `gptel-send`.

It will be possible to add multiple steering and queuing requests to the ongoing request.  The steering messages will be sent all at once, the queuing messages will be sent one-by-one.

(I can change the presets to just `@steer` and `@queue`, but I want to play it safe and use the `gptel-` prefix.)

What are your thoughts on this interface?  Do you have any other ideas for implementing steering/queuing?

karthink:
Another idea (suggested by John Wiegley) is to add more numeric prefix-args to `gptel-send`.
- A numeric prefix-arg of 0 will cause the prompt to be used to steer an ongoing query
- A prefix arg of 1 will cause the prompt to be queued for after the current query is done
- A `C-u` will bring up `gptel-menu` like right now.

yantar92:
karthink ***@***.***> writes:

> I know that you like to use a single gptel-send invocation to send LLMs on a long-duration task, so you might appreciate the "steer anytime" feature.
>
> Currently this is only accessible from `gptel-menu`, the command should show up when there is an active LLM query in the buffer.  (It's bound to `M-RET`.)

Interesting. I will try it out.

> I'm undecided on how to implement this interface more seamlessly.  I don't want to add more top-level commands like `gptel-steer`.  I like the idea of limiting the surface area of gptel to `gptel-send`, and shoving more advanced commands into the transient menu instead.
>
> ------
>
> One idea I have is to provide two presets, ***@***.***` and ***@***.***` with gptel.
> - Using ***@***.***` in the prompt in the buffer will cause the prompt to be added as a steering message to an ongoing request (or throw a warning if there isn't one).  It will be removed from the buffer.
> - Using ***@***.***` in the prompt in the buffer will cause the prompt to be queued as the next `gptel-send`, sending it automatically after the current response fully ends.

Another idea is doing it in more dwim fashion.

If gptel request is in progress, there is usually little point sending
yet another request. In fact, I did a mistake multiple times by calling
gptel-send accidentally, with point in some weird place in buffer,
resulting in two different gptel requests typing together into the same
buffer.

So, I consider several sensible scenarios:

1.
MY PROMPT

<LLM reasoning in progress...>

MY ADDITIONAL PROMPT

2.
MY PROMPT
MY ADDITION PROMPT (after possibly editing previous prompt)

<LLM reasoning in progress...>

In (1), the user may either inject a message into a reasoning chain or
extend the original prompt. The user may also queue next prompt, but I
would personally never consider doing such thing without seeing the
previous response first.
In (2) The user is basically editing the request before the answer is
generated, with the intention to amend the reasoning. The most likely
intention is doing gptel-abort -> delete unfinished response ->
re-submit. Or one may want to insert updated edited prompt mid-way into
the reasoning chain.

What gptel-send can do then is ask the user what to do - restart the
query or inject the text next turn.

Your idea with preset-like keywords can also work (I like less UI
interaction and more text that self-documents the interactions)

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

karthink:
> If gptel request is in progress, there is usually little point sending yet another request.

There are many ways to use gptel (`gptel-send` specifically), so this is not the case.  For example, I often fire multiple simultaneous requests in a setup like this:

``` org
<context and some question>

* Opinion 1
:PROPERTIES:
:GPTEL_MODEL: gpt-5.6-sol
:GPTEL_BACKEND: OpenAI
:END:

<Call gptel-send here>

* Opinion 2
:PROPERTIES:
:GPTEL_MODEL: glm-5.3
:GPTEL_BACKEND: zai
:END:

<Call gptel-send here>
```

> In fact, I did a mistake multiple times by calling gptel-send accidentally, with point in some weird place in buffer, resulting in two different gptel requests typing together into the same buffer.

Maybe we can add an option to guard against this, separately from this PR.  I would prefer the default behavior of gptel-send to be unchanged, because the mental model is very consistent -- "send a request from point" (no matter where you are in Emacs or what else is going on in the buffer.)

Previous discussion about this in #15, with some (now out of date) guard code: https://github.com/karthink/gptel/issues/15#issuecomment-2445733755

> So, I consider several sensible scenarios:
> 
> 1.  MY PROMPT
> 
> <LLM reasoning in progress...>
> 
> MY ADDITIONAL PROMPT
> 
> In (1), the user may either inject a message into a reasoning chain or extend
> the original prompt.

What do you mean by "extend the original prompt"?  That's not an action the API allows for?

> The user may also queue next prompt, but I would personally
> never consider doing such thing without seeing the previous response first.

Queuing is quite useful when you want to give it a set of independent tasks but don't want to send parallel requests.  Another use is to set up a "looping" prompt, where you want it to keep trying something until it succeeds.

> 2.  MY PROMPT MY ADDITION PROMPT (after possibly editing previous prompt)
> 
> <LLM reasoning in progress...>
> 
> In (2) The user is basically editing the request before the answer is
> generated, with the intention to amend the reasoning.  The most likely
> intention is doing gptel-abort -> delete unfinished response -> re-submit.  Or
> one may want to insert updated edited prompt mid-way into the reasoning chain.

If you've had a chance to try the steer-anytime command (`gptel--suffix-steer`), you'll have noticed that the steering instructions are read from the minibuffer.  I would like to read the steering instructions from the buffer and tried both (1) and (2), but both had problems:

1. In (1), it was difficult to type `MY ADDITIONAL PROMPT` below the `<LLM reasoning in progress...>` block when it is streaming in.  It's just very visually distracting.  (It works fine for non-streaming responses.)
2. In (2), I had trouble determining the right buffer extents to use for the steering message.  There are many edge cases here, especially if the original prompt has been edited instead of just being appended to.

> What gptel-send can do then is ask the user what to do - restart the query or inject the text next turn.

I want to avoid making `M-x gptel-send` pop up a multiple-choice question.

> Your idea with preset-like keywords can also work (I like less UI interaction and more text that self-documents the interactions)

Yes, writing the `@preset` behaviors for `@steer` and `@queue` is easy, but I have to solve the above two issues first.

-----

Let me know if tool call rejection + steering and standalone steering are working as expected for now.  I'm continuing to experiment with the steering flow and UI in the meantime.

elij:
I've implemented steering on top of gptel and I had no luck with user prompts before and after tool calls before DONE state. LLMs acknowledged (dismissed as tool feedback artefact) but didn't respond. I had to risk cache for single or persistent system directive appending until DONE. Has this implementation worked in evals? (So I can switch to upstream function)

karthink:
>  I had to risk [kv-] cache [invalidation] for single or persistent system directive appending until DONE. 

I don't follow this, sorry.  Could you elaborate?

elij:
Usually we don't touch the system directive for implicit cache but I could only get steering working with a modified system directive if an agent was in a long series of thoughts/tool calling

karthink:
I still don't follow, sorry.  Why does the system directive need to change at all for steering?

elij:
When I was testing this the agent didn't respond to injected user prompts between toolcalls -- that's why I'm asking if it worked for you

karthink:
It works as expected.

karthink:
> The agent thoughts would acknowledge message but not respond.

How a model responds to user messages is not something gptel can enforce.  I've only verified that the message is injected correctly and that the model acts on it in some way.

karthink:
@yantar92 I've now added the ability to steer the conversation from buffer text.  See 2b81e015903e4e37d24497321aa41f9d6037faf0.  You can include the steering message under the original prompt, or after the ongoing response, and call `gptel-send` with a prefix-arg of `0`.  It should be reasonably clever about determining the buffer bounds of the steering message.

There are lots of edge cases involving multiple active requests in a buffer, response redirection, read-only buffer handling and state cleanup, I'll fix them after we've decided on the UI for steering.

karthink:
> It should be reasonably clever about determining the buffer bounds of the steering message.

After more testing, I'd say this logic is completely broken.  For now, you can add a steering message from the buffer reliably by selecting a region.

karthink:
> After more testing, I'd say this logic is completely broken.

Fixed.  Typing in steering messages at "intuitive" locations relative to the ongoing response should work now.

karthink:
> You can include the steering message under the original prompt,

After more testing, I've concluded that it's not possible to reliably specify a steering message by modifying/appending to the original prompt, there are too many edge cases.  So that method is now removed/unsupported.

There are only two ways to specify a steering message in the buffer.  Both involve calling `gptel-send` with a prefix-arg of 0.

1. Type in your message anywhere and select the region you want to specify,
2. OR type below an ongoing response.

yantar92:
karthink ***@***.***> writes:

> karthink left a comment (karthink/gptel#1496)
>
>> If gptel request is in progress, there is usually little point sending yet another request.
>
> There are many ways to use gptel (`gptel-send` specifically), so this is not the case.  For example, I often fire multiple simultaneous requests in a setup like this:
> ...

Fair. I usually use several different buffers, because doing multiple
conversations in the same buffer feels risky given that gptel operates
with buffer-local variables to store state.

>> In fact, I did a mistake multiple times by calling gptel-send accidentally, with point in some weird place in buffer, resulting in two different gptel requests typing together into the same buffer.
>
> Maybe we can add an option to guard against this, separately from this PR.  I would prefer the default behavior of gptel-send to be unchanged, because the mental model is very consistent -- "send a request from point" (no matter where you are in Emacs or what else is going on in the buffer.)
>
> Previous discussion about this in #15, with some (now out of date) guard code: https://github.com/karthink/gptel/issues/15#issuecomment-2445733755

Thanks!

>> 1.  MY PROMPT
>> 
>> <LLM reasoning in progress...>
>> 
>> MY ADDITIONAL PROMPT
>> 
>> In (1), the user may either inject a message into a reasoning chain or extend
>> the original prompt.
>
> What do you mean by "extend the original prompt"?  That's not an action the API allows for?

What I meant is a feature in ChatGPT - "edit" original prompt will
basically stop current reasoning and restart afresh with the edited
version of the prompt.

>> The user may also queue next prompt, but I would personally
>> never consider doing such thing without seeing the previous response first.
>
> Queuing is quite useful when you want to give it a set of independent tasks but don't want to send parallel requests.  Another use is to set up a "looping" prompt, where you want it to keep trying something until it succeeds.

(I do not want to get too much out of topic, but I think that for
looping, there should be some way to set loop finish condition, which
calls for something a bit more involved than queuing prompts; maybe some
Elisp/text form to define rudimentary agent)

>> 2.  MY PROMPT MY ADDITION PROMPT (after possibly editing previous prompt)
>> 
>> <LLM reasoning in progress...>
>> 
>> In (2) The user is basically editing the request before the answer is
>> generated, with the intention to amend the reasoning.  The most likely
>> intention is doing gptel-abort -> delete unfinished response -> re-submit.  Or
>> one may want to insert updated edited prompt mid-way into the reasoning chain.
>
> If you've had a chance to try the steer-anytime command (`gptel--suffix-steer`), you'll have noticed that the steering instructions are read from the minibuffer.  I would like to read the steering instructions from the buffer and tried both (1) and (2), but both had problems:
>
> 1. In (1), it was difficult to type `MY ADDITIONAL PROMPT` below the `<LLM reasoning in progress...>` block when it is streaming in.  It's just very visually distracting.  (It works fine for non-streaming responses.)

Yeah, distracting. Also, If I query at the end of buffer, I just can't
put the point at the end of buffer so that point stays there.

> 2. In (2), I had trouble determining the right buffer extents to use for the steering message.  There are many edge cases here, especially if the original prompt has been edited instead of just being appended to.

Upon testing, I found another edge case - if the current reasoning never
calls any tools, steering silently fails with the steering prompt I
enter disappearing into the abyss.

> Let me know if tool call rejection + steering and standalone steering are working as expected for now.  I'm continuing to experiment with the steering flow and UI in the meantime.

Things are working for me now, except the above scenario when no tools
calls are done.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

karthink:
> Fair. I usually use several different buffers, because doing multiple conversations in the same buffer feels risky given that gptel operates with buffer-local variables to store state.

gptel requests bundle their state with the request, the buffer-local state is only relevant at the time that a request is sent.  So you should not have any issues with multiple requests in a buffer -- if you do it's a bug.  

That said, 
1. There are a few intentional exceptions to this, like the tool call confirmation settings -- this is read from the buffer-local value and not bundled with the request.   But these shouldn't be an issue.
2. Steering doesn't work with multiple simultaneous requests yet (it can get confused).  This will be fixed by the time I merge the feature.

> What I meant is a feature in ChatGPT - "edit" original prompt will basically stop current reasoning and restart afresh with the edited version of the prompt.

I understand what you meant by "gptel-abort + edit + gptel-send" now.

> I think that for looping, there should be some way to set loop finish condition

Yes, this is a separate topic, I also need to look at how this is handled in other LLM clients.

> Yeah, distracting. Also, If I query at the end of buffer, I just can't put the point at the end of buffer so that point stays there.

I'm planning some small tweaks to make streaming responses easier to work around.  One of them will be to add a newline after the trailing response marker if you are at `(eobp)` so you can jump to the end and type.

> Upon testing, I found another edge case - if the current reasoning never calls any tools, steering silently fails with the steering prompt I enter disappearing into the abyss.

Yes, I'm aware.  gptel will message the user that the steering prompt could not be sent, for sure.  But I'm not sure what else should happen.  If the steering was from the buffer text, should the text be deleted?

yantar92:
karthink ***@***.***> writes:

>> Upon testing, I found another edge case - if the current reasoning never calls any tools, steering silently fails with the steering prompt I enter disappearing into the abyss.
>
> Yes, I'm aware.  gptel will message the user that the steering prompt could not be sent, for sure.  But I'm not sure what else should happen.  If the steering was from the buffer text, should the text be deleted?

IMHO, it should _not_ be deleted. I'd expect an opposite - the text
should be highlighted with a message or even warning.
And if the steering message was added from prompt, it should be added to
the buffer after LLM response, so that it is not lost, and the user can
decide what to do with it.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

karthink:
> IMHO, it should _not_ be deleted. I'd expect an opposite - the text should be
> highlighted...

I'm not sure about this.  If the steering text was specified below the response, it makes sense to leave it there.  If it was specified near the prompt or somewhere else in the buffer (and provided with a region selected), leaving it there might be confusing and will invalidate the KV cache unless the user remembers to move it.

Maybe an unused steering prompt can always be moved to below the response, after the prompt-prefix?  Then it's naturally part of the next prompt, where the user's attention will be.

> ...with a message or even warning.

Agreed.

> And if the steering message was added from prompt, it should be added to the buffer after LLM response, so that it is not lost, and the user can decide what to do with it.

This case folds into the above.

yantar92:
karthink ***@***.***> writes:

>> IMHO, it should _not_ be deleted. I'd expect an opposite - the text should be
>> highlighted...
>
> I'm not sure about this.  If the steering text was specified below the response, it makes sense to leave it there.  If it was specified near the prompt or somewhere else in the buffer (and provided with a region selected), leaving it there might be confusing and will invalidate the KV cache unless the user remembers to move it.

Right. I was mostly referring to messages added via prompt.

> Maybe an unused steering prompt can always be moved to below the response, after the prompt-prefix?  Then it's naturally part of the next prompt, where the user's attention will be.

I actually did not try this approach yet as I practically used M-RET +
add from prompt in the recent days when I was testing.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,
Learn more about Org mode at <https://orgmode.org/>.
Support Org development at <https://liberapay.com/org-mode>,
or support my work at <https://liberapay.com/yantar92>

karthink:
Unsent/unusable steering messages will are now moved to the next prompt.  This works irrespective of how the message is specified (from gptel-menu/buffer text).

I've made it possible to type below the response without making a mess, but there's still no easy way to actually do this if the response is being added at the end of the buffer.

I've also ironed out more edge cases.  The main remaining edge case is that steering messages are not supported when there is more than one simultaneous request in a buffer.  I need to think about how to fix this.

karthink:
@yantar92, @elij: Steering (+ tool call rejection) has been merged into master.

Quick summary: steering requires that the LLM be able to make tool calls, and there are a few equivalent ways to specify a steering message:

1. From `gptel-menu` when a request is ongoing (bound to `M-RET` for now).  The steering message is read from the minibuffer.  To queue up more messages, press `M-p` in the prompt edit the previous message.  To cancel a queued message, provide an empty prompt.

2. From the response/chat buffer by selecting any region of text, and calling `gptel-send` with a prefix-arg of 0.  You can continue to edit the region (it is highlighted) until the queued message is picked up and sent by gptel.  You can delete the region text to cancel the steering message.

3. By typing below the ongoing response and calling `gptel-send` with a prefix-arg of 0.  This mirrors how you run `gptel-send` to send everything above the cursor.

I've made it easier to type below a (streaming) response now.  Your text will no longer get mixed up with the response chunks.  If the response is being added at the end, you can now /always/ just run `end-of-buffer` and start typing, and it should work as expected.

karthink:
There are probably still edge cases, but I think we need wider adoption to find those.

PR #1498, opened 2026-08-01
gptel-gh: Remove deprecated Copilot models

gessen:
Both `gemini-2.5-pro` and `gemini-3-flash-preview` are deprecated as of today (https://docs.github.com/en/copilot/reference/ai-models/supported-models#model-retirement-history)


karthink:
Merged, thanks!

PR #1500, opened 2026-08-02
gptel-gh: Update context window and costs

gessen:
This updates the context window and costs based on https://models.dev and model picker in VSCode.


karthink:
Merged, thanks!

PR #1504, opened 2026-08-07
gptel-openai-oauth: Carry model capabilities

OSadovy:
The models registered by `gptel-make-openai-oauth` are named as bare
symbols, so `gptel--process-models` has no specifications to attach to
them and they end up with empty symbol plists. Every capability check
against a model on this backend then fails, including for models whose
capabilities `gptel--openai-models` already describes.

Three effects, on a freshly registered backend:

- `gptel-add-file` on an image answers "Ignoring unsupported binary
  file", because `gptel-context--add-binary-file` checks
  `gptel--model-capable-p`.
- `gptel-track-media` never follows an image link in a chat buffer,
  because the MIME check in `gptel--parse-media-links` fails.
- The model menu shows neither a description nor a context window,
  since `gptel--infix-provider` reads both off the model symbol.

To reproduce, on master:

```elisp
(let ((backend (gptel-make-openai-oauth "Codex")))
  (mapcar (lambda (m) (cons m (gptel--model-capabilities m)))
          (gptel-backend-models backend)))
;; => ((gpt-5.2) (gpt-5.3-codex) ... (gpt-5.6-luna))   ;; all empty
```

This adds `gptel--openai-oauth-models`, following the
`gptel--<backend>-models` convention, which looks each name up in
`gptel--openai-models`. The advertised list and its order are
unchanged, and a model with no entry in that table — `gpt-5.3-codex`
and `gpt-5.3-codex-spark` are the two today — is still included as a
bare symbol.

After the change:

```
gpt-5.2                media=yes  mimes=(image/jpeg image/png image/gif image/webp)
gpt-5.3-codex          media=no   mimes=-
gpt-5.3-codex-spark    media=no   mimes=-
gpt-5.4-mini           media=yes  mimes=(image/jpeg image/png image/gif image/webp)
gpt-5.4                media=yes  mimes=(image/jpeg image/png image/gif image/webp)
gpt-5.5                media=yes  mimes=(image/jpeg image/png image/gif image/webp)
gpt-5.6-sol            media=yes  mimes=(image/jpeg image/png image/gif image/webp)
gpt-5.6-terra          media=yes  mimes=(image/jpeg image/png image/gif image/webp)
gpt-5.6-luna           media=yes  mimes=(image/jpeg image/png image/gif image/webp)
```

I have been using the fix against a ChatGPT subscription: images now
reach `gpt-5.6-sol` through the Codex endpoint and are described
correctly, and follow-up questions in the same chat buffer resend the
image as expected.

### DIFF
--- a/gptel-openai-oauth.el
+++ b/gptel-openai-oauth.el
@@ -23,6 +23,7 @@
 
 ;;; Code:
 (require 'cl-lib)
+(require 'gptel-openai)
 (require 'gptel-openai-responses)
 (require 'gptel-oauth)
 
--- a/gptel-openai-oauth.el
+++ b/gptel-openai-oauth.el
@@ -402,6 +402,17 @@ before constructing the headers."
        ("Originator"    . "gptel"))
      (and account-id `(("ChatGPT-Account-Id" . ,account-id))))))
 
+(defconst gptel--openai-oauth-models
+  (mapcar (lambda (name) (or (assq name gptel--openai-models) name))
+          '( gpt-5.2 gpt-5.3-codex gpt-5.3-codex-spark gpt-5.4-mini
+             gpt-5.4 gpt-5.5 gpt-5.6-sol gpt-5.6-terra gpt-5.6-luna))
+  "List of ChatGPT subscription models and their capabilities.
+
+Each model is looked up in `gptel--openai-models', so that the
+models served by the Codex endpoint carry the same specifications
+as their API counterparts.  A model absent from that table is
+included as a bare symbol, and has no known capabilities.")
+
 ;;;###autoload
 (cl-defun gptel-make-openai-oauth
     (name &key curl-args (stream t) request-params
@@ -409,9 +420,7 @@ before constructing the headers."
           (host "chatgpt.com")
           (protocol "https")
           (endpoint "/backend-api/codex/responses")
-          (models
-           '( gpt-5.2 gpt-5.3-codex gpt-5.3-codex-spark gpt-5.4-mini
-              gpt-5.4 gpt-5.5 gpt-5.6-sol gpt-5.6-terra gpt-5.6-luna)))
+          (models gptel--openai-oauth-models))
   "Register a ChatGPT Plus/Pro OAuth backend for gptel with NAME.
 
 This backend uses ChatGPT OAuth tokens (not OpenAI API keys) and


djr7C4:
Sending images seems to work for me already with Codex. 

Also, the code you provided is a no-op. All interned symbols with the same name share the same property list.

OSadovy:
Thanks for taking a look.

My config registers this backend and nothing else, which is the setup the
README documents:

```elisp
(setq gptel-model 'gpt-5.4-mini
      gptel-backend (gptel-make-openai-oauth "OpenAI-sub"))
```

There `gpt-5.6-sol` has an empty plist, so `gptel--model-capable-p`
returns nil and `gptel-add-file` answers "Ignoring unsupported binary
file".

Symbols do share one property list, and I think that's what differs
between our setups rather than being the reason the change does nothing:
something still has to populate it. `gptel--process-models` calls
`(setf (symbol-plist name) props)` only for entries carrying props, and
this backend names its models as bare symbols — so the plists stay empty
unless some other backend built from `gptel--openai-models` was
registered first, which then populates them for Codex too.

Same file, `emacs -Q --batch`:

```elisp
;; capcheck.el
(require 'gptel-openai-oauth)
(when (member "other-backend-first" command-line-args-left)
  (gptel-make-openai "ChatGPT" :key "dummy"))
(gptel-make-openai-oauth "Codex")
(princ (format "%S\n" (gptel--model-capabilities 'gpt-5.6-sol)))
```

```
master, Codex only            => nil
master, other backend first   => (media tool-use json url responses-api)
this PR, Codex only           => (media tool-use json url responses-api)
```

Do you have another OpenAI backend registered? That would explain images
working for you on master.

And if the intent is that this backend be used alongside an OpenAI one,
that's worth documenting — the README example above registers it on its
own, and the `gpt-5.4-mini` it suggests has no capabilities in that
configuration. Happy to send a README patch instead if you'd prefer that
to this change.

djr7C4:
1. The only change that actually does anything is `(require gptel-openai)`. The mapcar is a no-op.
2. I'm not the maintainer.

OSadovy:
Appreciate you testing this.

I think I can account for the difference: `gptel-gh.el` defines its own
`gpt-5.6-sol` entry with the same `:capabilities`, so registering a
Copilot backend populates that symbol's plist and the Codex backend picks
it up — no other OpenAI backend needed. Same for any backend built from
`gptel--openai-models`. Once something has populated the plists, this
change is genuinely redundant in that session.

The two halves of the diff on their own, `emacs -Q --batch` on master:

```
+ (require 'gptel-openai), no mapcar   => nil
+ mapcar, no explicit require          => (media tool-use json url responses-api)
```

Loading `gptel-openai` only defines the table; `gptel--process-models`
has to be handed those entries before anything is written onto the model
symbols, and the mapcar is what hands them over. The table was already
reachable via `gptel-openai-responses`, which is why the mapcar works
without the require — I added that separately because the reference is
now direct, not to change behaviour.

karthink:
I just pushed a commit that solves this (and related) problems in a much simpler way.  Now running `(require 'gptel-openai)` or `gptel-openai-responses` should be all that's needed to apply the model properties to the model symbols.

Please test the changes, and I can close this PR after.

OSadovy:
Confirmed fixed on a990c46, closing this. Thanks!

Two leftovers I can send separately:

- `gpt-5.3-codex` and `gpt-5.3-codex-spark` still have empty plists, as
  `gptel--openai-models` has no entry for either. They're only reachable
  through the Codex endpoint, so selecting one still gives "Ignoring
  unsupported binary file" and a blank entry in the model menu.
- `gptel-make-openai` says "For a list of currently recognized plist keys,
  see `gptel--openai-models'". That's now a list of bare symbols, so the
  pointer shows no keys. Same line in the Gemini, Anthropic and responses
  constructors.

karthink:
> Two leftovers I can send separately:
>
> - `gpt-5.3-codex` and `gpt-5.3-codex-spark` still have empty plists, as
>   `gptel--openai-models` has no entry for either. They're only reachable
>   through the Codex endpoint, so selecting one still gives "Ignoring
>   unsupported binary file" and a blank entry in the model menu.

Yeah, the properties for these will have to be set in gptel-openai-oauth.  A PR for this is welcome.

> - `gptel-make-openai` says "For a list of currently recognized plist keys,
>   see `gptel--openai-models'". That's now a list of bare symbols, so the
>   pointer shows no keys. Same line in the Gemini, Anthropic and responses
>   constructors.

Hmm, thanks for catching this!  The plist keys will have to be mentioned in the docstring now.  A PR for this is welcome too -- or I can get an LLM to do this since it's a very simple fix.  The same text should appear in most of the gptel-make-* backend constructor docstrings.

PR #1506, opened 2026-08-11
gptel: Prevent new user prompts from inheriting response property

jamesaimonetti:
The gptel text property is marked front-sticky while a response is being inserted so that streamed chunks seamlessly continue the response watermark.  Once the response completes, the trailing edge kept that front-stickiness, so text typed immediately after the response (e.g. before its trailing newline) inherited the gptel response property and was sent to the backend as an assistant message continuation.  Models like Claude 4.6+ reject assistant message prefill with a 400 error.

* gptel.el (gptel--seal-response): New function.  Remove front-sticky for the gptel property from the trailing edge of a completed response.
(gptel--inherit-stickiness): Only mark edited text as part of a response when the character following the change is itself front-sticky for the gptel property.
(gptel--handle-post-insert): Seal the response when it is finalized.

### DIFF
--- a/gptel.el
+++ b/gptel.el
@@ -808,11 +808,39 @@ send in queries.  (See `gptel--num-messages-to-send' for the last one.)"
   "Mark any change to an LLM response region as a response.
 
 Intended to be added to `after-change-functions' in gptel chat buffers,
-which see for BEG, END and PRE."
+which see for BEG, END and PRE.
+
+Text is only marked as part of a response if the character following
+the change is itself front-sticky for the `gptel' property.  This
+prevents the very last character of a completed response (whose
+stickiness is cleared by `gptel--seal-response') from marking
+subsequently-typed user text as an assistant continuation."
   (and (/= beg end) (< end (point-max))
-       (and-let* ((val (get-text-property end 'gptel)))
+       (and-let* ((val (get-text-property end 'gptel))
+                  ((or (eq (get-text-property end 'front-sticky) t)
+                       (memq 'gptel (ensure-list
+                                     (get-text-property end 'front-sticky))))))
          (add-text-properties
           beg end `(gptel ,val front-sticky (gptel))))))
+
+(defun gptel--seal-response (&optional pos)
+  "Remove front-stickiness of the `gptel' property at POS.
+
+POS defaults to the end of the last response in the buffer, i.e.
+the character at `point-max' if it is part of a response.
+
+The `gptel' text property is marked front-sticky while a response
+is being inserted, so that streamed chunks seamlessly continue the
+\"this is a response\" watermark.  Once the response is complete
+the trailing edge must not remain sticky: text typed immediately
+after the response is a new user turn, not a continuation of the
+assistant message.  Leaving it sticky causes such text to inherit
+the `gptel' response property and be sent to the backend as
+assistant message prefill, which models like Claude 4.6+ reject
+with a 400 error."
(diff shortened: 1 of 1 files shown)


karthink:
> The gptel text property is marked front-sticky while a response is being inserted so that streamed chunks seamlessly continue the response watermark.

I think you have this backwards?  To seamlessly continue the watermark, the property would have to be rear-sticky, not front-sticky.

Preserving the response watermark happens because we explicitly add the text property when inserting chunks, and not because of front or rear-stickiness.

> Once the response completes, the trailing edge kept that front-stickiness, so text typed immediately after the response (e.g. before its trailing newline) inherited the gptel response property and was sent to the backend as an assistant message continuation. Models like Claude 4.6+ reject assistant message prefill with a 400 error.

This is also backwards.  text typed immediately _after_ the response inherits the text property if the response is _rear-sticky_, not front sticky.

So I don't understand how removing the front-sticky property from the response's trailing edge helps, sorry.  Let me know if I missed something, or I can close this issue.

jamesaimonetti:
Sorry, tried finding the original session with the agent that this issue occurred under.

Basically, I had a tool call come from the model malformed (something like `Write` with path/filename = `nil`) and no user prompt `###`. When I manually typed `### ` to start a prompt to ask what happened to the tool call construction and sent that up, Claude API responded about not prefilling.

running `C-u C-x =` on a character typed immediately after a completed response showed front-sticky (gptel) and gptel: response present on it, in a buffer where that text was never part of any response. I think I was just trying both front and rear sticky options and it "worked" with front-sticky (but i also was hurrying a bit).

Would  `rear-nonsticky` get set on that property anywhere and be useful? The stickyness is new-to-me Emacs stuff so apologies for getting it flipped.

jamesaimonetti:
oh forgot, i did add this to my config:

```
If a tool-call fails, there's metadata on the text that could cause the next user prompt to be tagged as an assistant message, getting the 400 error that the model won't prefill any more.

Use =C-c C-n= to start a new user prompt
#+begin_src emacs-lisp
(defun my/gptel-new-prompt ()
  "Insert a fresh prompt prefix, explicitly stripping any inherited
`gptel' response text-property so this new text can never be
misclassified as a continuation of the previous response."
  (interactive)
  (let ((start (point)))
    (insert "\n\n### ")
    (remove-text-properties start (point) '(gptel nil front-sticky nil))))

(with-eval-after-load 'gptel
  (define-key gptel-mode-map (kbd "C-c C-n") #'my/gptel-new-prompt))
```

so rather than pressing enter after the failed tool call and getting into this mess, I default to using the keybinding and that sorted it out too. Not sure if this is helpful or not?

jamesaimonetti:
Found the session where the agent derived the fix, in response to your comment above, it wrote the reproduction step:

> (streamed response ending in `\n`, point placed before it, `gptel--inherit-stickiness` re-stamping typed text via `(get-text-property end 'gptel)`)

karthink:
There appears to be some confusion here.  At https://github.com/karthink/gptel/blob/master/gptel.el#L804 you can see that the gptel property is rear-nonsticky by default.  This means typing at the end of a response should not cause the gptel text property to be inherited.

So I would first check the value of `text-property-default-nonsticky` in a session where you are experiencing this issue.  If it does not include gptel (i.e. `(gptel . t)`), then we can investigate why this might have happened.

The LLM you're using clearly has no idea of how gptel's response boundaries are handled, so I would stop relying on it.

PR #1512, opened 2026-08-14
gptel-anthropic: Handle mid-stream event: error

jamesaimonetti:
Anthropic can return HTTP 200 and begin a stream, then signal failure in-band with an event: error block (e.g. overloaded_error).  The stream parser had no case for it, so the error was silently swallowed and the request finished as if it had succeeded.

Set :error from the event payload so the request is routed through the existing error path (status, message, post-response hooks).

### DIFF
--- a/gptel-anthropic.el
+++ b/gptel-anthropic.el
@@ -167,7 +167,14 @@ information if the stream contains it.  Not my best work, I know."
               (plist-put info :stop-reason
                          (map-nested-elt response '(:delta :stop_reason)))
               ;; Capture token usage
-              (gptel--anthropic-update-tokens (plist-get response :usage) info)))))
+              (gptel--anthropic-update-tokens (plist-get response :usage) info)))
+           ;; In-band error event: HTTP 200 but the stream signals failure
+           ;; (e.g. `overloaded_error').  Treat it like an HTTP-level error.
+           ((looking-at "error")
+            (forward-line 1) (forward-char 5)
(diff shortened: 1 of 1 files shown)


karthink:
@jamesaimonetti Thanks for the PR.  I don't have an example of a log for a request that failed like this, so: is it possible to also capture the token use and the stop reason if these are available?

This is how it's done in the "message_delta" branch of the parser:
``` emacs-lisp
;; Capture stop reasion
(plist-put info :stop-reason
           (map-nested-elt response '(:delta :stop_reason)))
;; Capture token usage
(gptel--anthropic-update-tokens (plist-get response :usage) info)))
```

jamesaimonetti:
```json
data: {
  "type": "error",
  "error": {
    "details": null,
    "type": "overloaded_error",
    "message": "Overloaded"
  },
  "request_id": "req_011Ce3HwQb6vtbs9LQVmh4oe"
}
```
from a note i made, checking if i have the full example somewhere. it was an ephemeral error talking to claude though

karthink:
This means the API does not report the stop-reason or the token use?

karthink:
@jamesaimonetti gentle ping -- are the stop-reason or token usage statistics not available at this mid-stream error event?

jamesaimonetti:
@karthink here's the e2e test co-developed with an agent: https://github.com/jamesaimonetti/gptel-test/blob/main/e2e/e2e-stream.el

it was captured during the debugging session when I was getting the overloaded_error.

PR #1513, opened 2026-08-15
gptel-request: Call gptel-post-request-hook in corresponding buffer

hiecaq:
Hi, I think calling this hook in the caller buffer can be helpful for what the the hooks would want to do.  Without this PR, the current buffer would be randomly chosen, so there is no easy way AFAIK to know which buffer instantiates this gptel-request.

For example, I want to read some buffer-local variables in the chat buffer.

Also it's possible now to search for the corresponding FSM to figure out the runtime configurations, if the user wants to.

Thanks in advance!


karthink:
@hiecaq That makes sense, thanks for the PR!  Merged.

PR #1516, opened 2026-08-18
gptel-oauth: Extract generic auth-code callback handling, support SSH

ChetanKoneru:
Follow-up to #1463

1. **Extract the callback server logic into `gptel-oauth.el` .**
   `gptel--openai-oauth-read-code` now delegates to a new generic
   `gptel-oauth--read-code (authorization-url redirect-path state port timeout)`.
   It handles the loopback server, request parsing, path/state/error/code
   validation and the HTTP response, so other backends (e.g. Claude) can reuse it.
   OpenAI-specific pieces (authorization URL, constants, token exchange) are unchanged.

2. **Allow auth-code OAuth over SSH.**
   Over SSH no server is started. The user opens the authorization URL in their
   local browser, lets the callback fail there, and pastes the callback URL from
   the address bar into Emacs. That URL goes through the same validation as the
   server callback, so the rest of the handling is unchanged. The previous
   `user-error` blocking auth-code login over SSH and the corresponding TODO are removed.

Also updates the one `gptel-openai-oauth-login-method` docstring sentence that
said auth-code login was unsupported over SSH.

- Under `SSH_CLIENT` / `SSH_CONNECTION` / `SSH_TTY` , the `make-network-process=`server branch is skipped entirely.
-  The authorization URL is shown/copied, and `read-from-minibuffer` asks the user to paste the callback URL from the browser's address bar. That URL is parsed with `url-generic-parse-url` and validated by the same path/state/error/code logic used for the server callback.

#### Authorization Code login over SSH

Two machines are involved:

- **Remote host** — where Emacs runs, reached over SSH. No browser.
- **Local machine** — where the user's terminal and browser run. "localhost"
  in the browser refers to this machine.

Steps:

1. Emacs (on the remote host) builds the authorization URL, copies it to the
   clipboard and displays it. It does not start a callback server and does not
   open a browser, since a server on the remote host would be unreachable from
   the user's browser.

2. The user opens that URL in the browser on the local machine.

3. The user authorizes with OpenAI. OpenAI redirects the browser to the
   registered callback: `http://localhost:1455/auth/callback?code=...&state=...`

4. The page fails to load, because nothing is listening on port 1455 on the
   local machine. The address bar still contains the full callback URL,
   including `code` and `state`.

5. The user copies the URL from the address bar and pastes it at the Emacs
   minibuffer prompt on the remote host.

6. Emacs parses the pasted URL, verifies the callback path and `state`, reports
   any `error`/`error_description`, extracts `code`, and exchanges it for a
   token — identical to the handling in the local-server flow.

Haven't tested it yet as I do not run Emacs over ssh. Need help testing.

### DIFF
--- a/gptel-oauth.el
+++ b/gptel-oauth.el
@@ -121,6 +121,103 @@ If your browser does not open automatically, browse to %s: "
        (format "(One-time code %s copied) Press ENTER after authorizing: "
                user-code)))))
 
+(defun gptel-oauth--read-code (authorization-url redirect-path state port timeout)
+  "Open AUTHORIZATION-URL and return an OAuth authorization code.
+
+REDIRECT-PATH and STATE validate the callback.  PORT specifies the
+loopback callback server and TIMEOUT is the maximum wait in seconds.
+When running over SSH, prompt for the callback URL after authorization
+instead of starting a local server."
+  (cl-labels
+      ((callback-code (target)
+         (let* ((query-start (string-search "?" target))
+                (path (if query-start (substring target 0 query-start) target))
+                (query (and query-start
+                            (url-parse-query-string
+                             (substring target (1+ query-start)))))
+                (callback-state (cadr (assoc "state" query)))
+                (callback-code (cadr (assoc "code" query)))
+                (callback-error (cadr (assoc "error" query)))
+                (callback-error-description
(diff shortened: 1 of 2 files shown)


karthink:
@ChetanKoneru thanks for the update!  I'll take a look soon.

PR #1520, opened 2026-08-22
gptel-openai-extras: Add deepseek-v4-flash-vision-exp

hiecaq:
See https://api-docs.deepseek.com/guides/vision/

It's other parameters (cost etc) are the same as deepseek-v4-flash's

BTW, the cost for deepseek models has [changed](https://api-docs.deepseek.com/quick_start/pricing) recently, but AFAIK currently there is no way in gptel to express their peak/off-peak prices, so I didn't change them right now.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -83,6 +83,8 @@
 - OpenAI backend: Add support for =gpt-5.6-sol=, =gpt-5.6-terra= and
   =gpt-5.6-luna=.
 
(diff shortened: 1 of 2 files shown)


karthink:
Merged, thanks!

> currently there is no way in gptel to express their peak/off-peak prices, so I didn't change them right now.

This is one of the reasons gptel displays token use and not dollar costs in the UI.

PR #1521, opened 2026-08-23
fix(bedrock): include model attributes by default

### LINKED ISSUE #1412: bedrock does not include model attributes by default

### Bug Description

This is a cleaner and shorter dupe of #1405 

`gptel-make-bedrock`'s default `:models` do not include attributes. To get full model support we need to provide an explicit list from the other definitions. It might make sense to do this in the default, and looking up properties from a concatenation of all the other backends. But that may also be fragile. In any case, this was an unexpected behaviour.

```
  (setq
   gptel-model 'claude-opus-4-7
   gptel-backend (gptel-make-bedrock "Bedrock"
                   :region "us-east-1"
                   ;; WORKAROUND: models are not media capable by default, so
                   ;; pass an explicit list from the other backend.
                   :models (mapcar (lambda (sym) (assq sym gptel--anthropic-models))
                                   '(claude-opus-4-7 claude-opus-4-6 claude-sonnet-4-6))
                   :model-region 'us))
```

### Backend

Other (please specify in Additional Context)

### Steps to Reproduce

try to attach media to a simple bedrock backend

### Additional Context

trixie

### Backtrace

```emacs-lisp

```

### Log Information

```json

```

mikemikimike:
## Summary

Closes #1412.

AWS Bedrock backends currently initialize their default model list with bare symbols, so model metadata such as capabilities and supported MIME types is unavailable to gptel's UI and request handling. This is especially visible when attaching media to a Bedrock backend.

## Changes

- Reuse metadata from the built-in Anthropic model definitions for matching Bedrock models.
- Keep non-Anthropic Bedrock models as bare symbols, preserving existing model IDs and custom `:models` behavior.
- Add a regression test covering default Bedrock model metadata.

## Verification

- `git diff --check` — passed.
- ERT test command was not run: Emacs is not installed in the environment, and the attempted GNU Emacs `winget` installation could not complete.

Known limitation: the full ERT suite and Emacs byte-compilation still need to run in CI or a local Emacs environment.


karthink:
I just pushed a commit that solves this (and related) problems in a much simpler way.  Now running `(require 'gptel-anthropic)` should be all that's needed to apply the model properties to the model symbols.

Please test the changes with the scope of this PR reduced to `(require 'gptel-anthropic)`, and I can merge it.

mikemikimike:
The test scope is now reduced to the requested load path.

At head `443f1fe7cddeed24c004a1b884e883b603221252`, `gptel-anthropic-test.el` does not load the backend at file scope. The ERT test clears the target model's properties, runs only:

```elisp
(require 'gptel-anthropic)
```

and then verifies that `claude-sonnet-4-6` receives the `media` capability and JPEG MIME support.

I could not execute ERT in this Windows environment because Emacs is not installed. I attempted both the available GNU Emacs package installation and a disposable Docker-based setup, but neither completed because the package download/index stalled. `git diff --check` passes, and no production code changes are included in this PR.

PR #1502, opened 2026-08-05
gptel-anthropic: Add Claude Opus 5 model

maxecharel:
* gptel-anthropic.el (gptel--anthropic-models):
- Add `claude-opus-5' model

* NEWS (New models and backends): Mention changes.

PR #1523, opened 2026-08-24
add recent anthropic models to aws bedrock

fommil:
add missing bedrock references see #1502


karthink:
Merged, thank you @fommil!

PR #1524, opened 2026-08-24
fix typo in gptel-rewrite.el

nano-o:


karthink:
Thanks for catching that!  Merged.

PR #1536, opened 2026-09-05
gptel-openai: Add gpt-6-astra

djr7C4:


karthink:
Merged, thanks @djr7C4!

PR #1542, opened 2026-09-13
gptel-openai-extras: Add deepseek-flash and deprecate old models

hiecaq:
DeepSeek has deprecated deepseek-v4-flash and deepseek-v4-flash-vision-exp in favor of deepseek-flash (which is actually v4.1-flash), see [here](https://api-docs.deepseek.com/news/news260910/#-v41-flash-is-now-live-on-the-deepseek-api-with-native-multimodal-support).

I also changed the cost to the current off-peak prices.


karthink:
@hiecaq Thanks for the update!  Could you also mention this change in the NEWS file?

hiecaq:
@karthink done ;)

karthink:
Merged, thanks!

PR #1549, opened 2026-09-16
Add Perplexity integration attribution

qirh:
## Summary

Perplexity would like to identify API requests made through gptel.

- Add `X-Pplx-Integration: gptel` to the default Perplexity backend headers.
- Preserve custom backend header configuration.

## Verification

The change only extends the existing Perplexity default-header alist. This repository does not run automated pull-request checks.


qirh:
@karthink could you take a look at this when you get a chance?

